PR #82, opened 2016-06-04
Added url for pending transactions

kLkA:


kLkA:
will be recreated from different branch to not to mix up many pull request

PR #105, opened 2016-07-13
updated by polarislee

polarislee1984:
last static gulp work result


polarislee1984:
Need to update main repository

PR #179, opened 2016-10-31
{NW} reflect.go added, TestReflectData added

KarloB:
Removed my stuff from encoder.go, moved to new file.
Added new test to encoder_test.go, named TestReflectData.
Added some stuff to function, still ain't finished of course.
Rename new file if needed, unfortunately i haven't thought of a better name.


KarloB:
Encoded reflected struct to sha256

PR #200, opened 2016-11-30
node test is fixed

zgordan-vv:


Ch3ck:
Just a suggestion @zgordan-vv I think a description of the actual changes might make more sense

PR #204, opened 2016-12-03
Reverted changes to Transport factory && Added fixes to transport package

Ch3ck:


Ch3ck:
@skycoin removed the go routine added in stub.

skycoin:
thx

PR #230, opened 2016-12-23
Fix bug about genesis block

iketheadore:
- Fix the bug in parsing genesis block
- Fix the display bug of uxout hash in genesis block


skycoin:
only affects json representation and printing of genesis transaction

PR #297, opened 2017-03-02
src/daemon/gnet add ability to use raw connections before starting internal read/write loop

logrusorgru:
The ability to use raw connections is good for handshakes. When I need to send-receive some messages with known size and known timeout, I uses messages. But, it's very hard and can be replaced with two functions that works on raw connections.

+ add Acquire/Release methods to be able to use raw incoming connections asynchronously
+ add tests to Acquire and Release methods
+ add documentation to the methods
+ change test behavior to use '-v' flag to show debug logs


logrusorgru:
Doesn't need anymore

PR #300, opened 2017-03-04
Fix all detected race condition

iketheadore:
- Rewrite the daemon/gnet connection pool to make it thread safe
- Rewrite member variables in daemon.go to make it thread safe


src/daemon/gnet/pool.go
```diff
@@ -115,523 +117,568 @@ type Connection struct {
...
+
+// NewConnection creates a new Connection around a net.Conn.  Trying to make a connection
+// to an address that is already connected will panic.
+func (pool *ConnectionPool) NewConnection(conn net.Conn) (*Connection, error) {
+	a := conn.RemoteAddr().String()
+	c := make(chan struct{})
+	var nc *Connection
+	var err error
+	pool.memChannel <- func() {
+		defer func() {
```
logrusorgru:
```go
c := make(chan struct{})
pool.memChannel <- func() {
     defer close(c)
    // stuff here
}
<-c
return
```

src/daemon/gnet/pool.go
```diff
@@ -115,523 +117,568 @@ type Connection struct {
...
+	close(pool.Quit)
+	pool.listener.Close()
+	pool.listener = nil
+	pool.pool = make(map[int]*Connection)
+	pool.addresses = make(map[string]*Connection)
+}
+
+// NewConnection creates a new Connection around a net.Conn.  Trying to make a connection
+// to an address that is already connected will panic.
+func (pool *ConnectionPool) NewConnection(conn net.Conn) (*Connection, error) {
```
logrusorgru: It's possible to use
```go
func (pool *ConnectionPool) NewConnection(conn net.Conn) (nc *Connection, err error) {
    // blah
   return // nc, err
}
```
instead of 
```go
var nc *Connection
var err error
// blah
return nc, err
```

src/daemon/gnet/pool.go
```diff
@@ -115,523 +117,568 @@ type Connection struct {
...
-	conns := make([]*Connection, len(self.Pool))
-	i := 0
-	for _, c := range self.Pool {
-		conns[i] = c
-		i++
-	}
-	return conns
-}
+	// process the received message
+	go func() {
```
logrusorgru: the same as
```go
go func() {
    <-q // read the channel and retrurn
}()
```

iketheadore: Hi @logrusorgru, this code has already been deleted in the latest commit

src/daemon/gnet/pool.go
iketheadore: Thanks, I will change it

src/daemon/gnet/pool.go
iketheadore: Thanks, this looks much cleaner, I will use it in strand function.

src/daemon/gnet/pool.go
logrusorgru: Never mind, I just want to receive notifications about the PR.

skycoin:
Great job!

PR #303, opened 2017-03-11
src/cipher/encoder failed test case for fields

logrusorgru:
src/cipher/encoder fields related failed tests case

The problem is: the encoder can decode structs from structs but we need a way to decode this structs back. But we're using 1-demensional schema (see `encoder.StructField`). It seems that we need another `StructField`.


logrusorgru:
It treats fields kind of `reflect.Array` as `cipher.SHA256` and structs too (so wired but true). And returning to structs in structs, we can't determine size of field.

logrusorgru:
+ move from `ParseFields(blah) map[string]string` to ParseFields(blah) (map[string]interface{}, error)
+ remove unnecessary coping

logrusorgru:
Remove test cases that we don't need anymore.

logrusorgru:
Don't need anymore

PR #358, opened 2017-06-12
Add local dup wallet checking

iketheadore:
Check the first address of each wallet to see if this wallet is duplicated, if found the first address is dup, then choose the wallet whose address number is bigger.

### Note:
The api for creating and reloading wallet have been updated for checking dup wallet, we need the frontend wallet to check the return value of the api, and remind user when they trying to create dup wallet, @nakulpandey.

### DIFF
--- a/src/gui/wallet.go
+++ b/src/gui/wallet.go
@@ -21,24 +21,12 @@ import (
 	wh "github.com/skycoin/skycoin/src/util/http" //http,json helpers
 )
 
-//var Wallets wallet.Wallets
-
-/*
-REFACTOR
-*/
-
-/*
-This section is redundant
-- after moving the wallets out of visor and daemon, wallet should be in the wallet module
-- there is no need for multiple wallets in the same application
-*/
-//type WalletRPC struct{}
-
 // WalletRPC wallet rpc
 type WalletRPC struct {
 	Wallets         wallet.Wallets
 	WalletDirectory string
 	Options         []wallet.Option
+	firstAddrIDMap  map[string]string // key: first address in wallet, value: wallet id
 }
 
 // NotesRPC note rpc
@@ -77,7 +65,9 @@ func NewNotesRPC(walletDir string) *NotesRPC {
 
 // NewWalletRPC new wallet rpc
 func NewWalletRPC(walletDir string, options ...wallet.Option) *WalletRPC {
-	rpc := &WalletRPC{}
+	rpc := &WalletRPC{
+		firstAddrIDMap: make(map[string]string),
+	}
 	if err := os.MkdirAll(walletDir, os.FileMode(0700)); err != nil {
 		logger.Panicf("Failed to create wallet directory %s: %v", walletDir, err)
 	}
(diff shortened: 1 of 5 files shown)


src/gui/wallet.go
```diff
@@ -113,62 +104,105 @@ func NewWalletRPC(walletDir string, options ...wallet.Option) *WalletRPC {
 	return rpc
 }
 
+func (wrpc *WalletRPC) removeDup(wlts wallet.Wallets) wallet.Wallets {
+	var dupWltIDS []string
+	// remove dup wallets
+	for wltID, wlt := range wlts {
+		addr := wlt.Entries[0].Address.String()
```
gz-c: are `Entries` sorted?

iketheadore: The entries is a slice that loaded from wallet json file, it's order is always the same.

iketheadore: And we should not sort it, cause when creating new wallet from seed, the new wallet's first address might be different from the wallet that has the same seed but sorted.

src/wallet/deterministic.go
```diff
@@ -31,23 +31,19 @@ type Option func(w *Wallet)
 
 // NewWallet generates Deterministic Wallet
 // generates a random seed if seed is ""
-func NewWallet(wltName string, opts ...Option) Wallet {
-	//old seed generation
-	//seedRaw := cipher.SumSHA256(secp256k1.RandByte(64))
-	//seed := hex.EncodeToString(seedRaw[:])
-
+func NewWallet(wltName string, opts ...Option) (*Wallet, error) {
```
gz-c: make sure that all callers of this method are checking the error, since previously it expected it to panic.  alternatively create a separate `MustNewWallet` that panics on error, if necessary

src/gui/wallet.go
```diff
@@ -423,9 +457,14 @@ func walletCreate(gateway *daemon.Gateway) http.HandlerFunc {
 		// the wallet name may dup, rename it till no conflict.
 		for {
 			wlt, err = Wg.CreateWallet(wltName, wallet.OptSeed(seed), wallet.OptLabel(label))
-			if err != nil && strings.Contains(err.Error(), "renaming") {
-				wltName = wallet.NewWalletFilename()
-				continue
+			if err != nil {
+				if strings.Contains(err.Error(), "renaming") {
```
gz-c: I know that this is not a new line, but this error check is strange

src/wallet/deterministic.go
iketheadore: We have lots of remaining code use panic directly, which will cause node can't gracefully shutdown, I will refactor that later.

src/gui/wallet.go
iketheadore: This is old legacy code, will refactor later.

src/gui/wallet.go
gz-c: I meant to ask, is it safe to compare Entries[0] to detect duplicate wallet

iketheadore: I will add `len(Entries)` checking, in case the wallet is empty. But that will only occur if user delete the entries in the wallet json file manually, new created wallet will have a default entry. It should be safe to compare the first address to detect dup wallet.

Issue #352, opened 2017-06-12
Store blockchain seqs and unspents in boltdb


iketheadore:
### Description

For now, we have block tree struct to store all blocks in db, each depth of the tree may have multiple blocks, we use consensus algorithm to determine the path and form the blockchain.  

Each time we start the node, will go through the block tree and get the blockchain head and unspent pool, which is a waste of time. So we now decide to store unspents and block head in boltdb, which will speed up our wallet opening speed.

iketheadore:
Fixed in #363

PR #363, opened 2017-06-16
Use db to store unspent outputs and block head

iketheadore:
- Fix the #352. 
- Update READEME about the formatting issue
- Update test cases in historydb package
- Service unspent outputs in RAM


iketheadore:
Have bug if the unspent pool bucket is broken, need to reset the bucket and parse again.

gz-c:
should we retain the in-memory unspent pool as an option?

src/gui/wallet.go
```diff
@@ -499,7 +483,8 @@ func walletNewAddresses(gateway *daemon.Gateway) http.HandlerFunc {
 		}
 
 		if err := Wg.SaveWallet(wltID); err != nil {
-			wh.Error500(w, "")
+			logger.Error("save wallet failed when generate new addresses: %v", err)
```
gz-c: `logger.Errorf` or `logger.Error("save wallet failed when generate new addresses:", err)`

gz-c: There are multiple places this change is needed in this file, I am referencing this documentation: https://godoc.org/github.com/op/go-logging#Logger.Error

src/visor/blockchain.go
```diff
@@ -235,20 +263,25 @@ func (bc *Blockchain) ExecuteBlock(b *coin.Block) (coin.UxArray, error) {
 	txns := b.Body.Transactions
 	for _, tx := range txns {
 		// Remove spent outputs
-		bc.unspent.DelMultiple(tx.In)
+		if err := bc.Unspent.Delete(tx.In); err != nil {
```
gz-c: does it need to rollback previous changes if it encounters error partway through?

gz-c: same question for several locations in this file

src/visor/blockchain_parser.go
```diff
@@ -41,16 +39,6 @@ func NewBlockchainParser(hisDB *historydb.HistoryDB, bc *Blockchain, ops ...Pars
 	return bp
 }
 
-// // BindParsedNotifier sets the parsedFunc
-// func (bcp *BlockchainParser) BindParsedNotifier(f func(height uint64)) {
-// 	bcp.parsedFunc = f
-// }
-
-// Start the parsing process
-func (bcp *BlockchainParser) Start() {
```
gz-c: is `startC` no longer used? can remove from struct if so

src/visor/blockchain_parser.go
```diff
@@ -93,35 +82,24 @@ func (bcp *BlockchainParser) Stop() {
...
-		// logger.Critical("historydb parse %d/%d", bcHeight, bcHeight)
 	}
-	bcp.parsedHeight = bcHeight
-	return nil
+
+	if parsedHeight < int64(bcHeight) {
+		logger.Info("parse block from %d to %d", parsedHeight+1, bcHeight)
+	}
+
+	return bcp.historyDB.SetParsedHeight(bcHeight)
```
gz-c: is it possible for `bcHeight < parsedHeight`? if so i'm not sure it is ok to call `SetParsedHeight` when that occurs. should check for `bcHeight < parsedHeight` as initial condition

src/visor/blockdb/unspent.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		return err
+	}
+
+	xorhash = xorhash.Xor(ux.SnapshotHash())
+	if err := up.setXorHash(xorhash); err != nil {
+		return err
+	}
+
+	return up.pool.Put([]byte(h.Hex()), v)
```
gz-c: what happens if `Put` fails here? `setXorHash` has succeeded

src/visor/blockdb/unspent.go
```diff
@@ -0,0 +1,248 @@
...
+			return coin.UxOut{}, false, fmt.Errorf("get uxout from pool failed: %v", err)
+		}
+		return ux, true, nil
+	}
+
+	return coin.UxOut{}, false, nil
+}
+
+// GetAll returns Pool as an array. Note: they are not in any particular order.
+func (up *UnspentPool) GetAll() (ua coin.UxArray, err error) {
```
gz-c: can remove the return values identifiers

src/visor/blockdb/unspent.go
```diff
@@ -0,0 +1,248 @@
...
+		arr = append(arr, uxout)
+		return nil
+	}); err != nil {
+		return coin.UxArray{}, err
+	}
+
+	return arr, nil
+}
+
+// delete removes an unsepnt from the pool by hash
```
gz-c: unsepnt -> unspent

src/visor/blockdb/unspent.go
```diff
@@ -0,0 +1,248 @@
...
+			return err
+		}
+
+		hash = hash.Xor(ux.SnapshotHash())
+		if err := up.setXorHash(hash); err != nil {
+			return err
+		}
+
+		// delete unspent hash from pool
+		if err := up.pool.Delete([]byte(h.Hex())); err != nil {
```
gz-c: same issue here, if `Delete` fails, the xorhash is already set

src/visor/blockdb/unspent.go
```diff
@@ -0,0 +1,248 @@
...
+		if err := encoder.DeserializeRaw(v, &ux); err != nil {
+			return err
+		}
+
+		if ux.Body.Address == addr {
+			uxs = append(uxs, ux)
+		}
+		return nil
+	}); err != nil {
+		return coin.UxArray{}, fmt.Errorf("get unsepnts of address %v failed: %v",
```
gz-c: unsepnts -> unspents

src/visor/blockdb/unspent.go
```diff
@@ -0,0 +1,248 @@
...
+			addr.String(), err)
+	}
+
+	return uxs, nil
+}
+
+// GetUnspentsOfAddrs returns unspent outputs map of given addresses,
+// the address as return map key, unspent outputs as value.
+func (up *UnspentPool) GetUnspentsOfAddrs(addrs []cipher.Address) (coin.AddressUxOuts, error) {
+	addrm := make(map[cipher.Address]byte, len(addrs))
```
gz-c: `map[Type]struct{}` is idiomatic for lookup tables,
the line below would be `addrm[a] = struct{}{}` instead of `addrm[a] = byte(1)`.  empty struct uses no memory

src/visor/historydb/historydb_test.go
```diff
@@ -52,37 +51,45 @@ var _ = func() int64 {
...
 }()
 
+// setup will create a random boltdb file in temp folder,
+// and return teardown function for later clean
 func setup(t *testing.T) (*bolt.DB, func(), error) {
-	dbName := fmt.Sprintf("%ddb", rand.Int31n(10000))
-	teardown := func() {}
-	tmpDir := filepath.Join(os.TempDir(), dbName)
+	dbName := fmt.Sprintf("%d.db", rand.Int31n(100))
+	close := func() {}
```
gz-c: although `close` is not reserved, should avoid assigning to builtin methods

src/visor/rpc.go
```diff
@@ -33,21 +34,23 @@ func (rpc RPC) GetBlockchainMetadata(v *Visor) *BlockchainMetadata {
 }
 
 // GetUnspent gets unspent
-func (rpc RPC) GetUnspent(v *Visor) coin.UnspentPool {
-	return v.Blockchain.GetUnspent().Clone()
+func (rpc RPC) GetUnspent(v *Visor) *blockdb.UnspentPool {
+	return v.Blockchain.Unspent
```
gz-c: i think this intentionally returned a copy of the unspent to avoid certain kinds of errors

src/visor/unconfirmed.go
```diff
@@ -414,50 +415,71 @@ func (utp *UnconfirmedTxnPool) GetKnown(txns []cipher.SHA256) coin.Transactions
 // SpendsForAddresses returns all unconfirmed coin.UxOut spends for addresses
 // Looks at all inputs for unconfirmed txns, gets their source UxOut from the
 // blockchain's unspent pool, and returns as coin.AddressUxOuts
-func (utp *UnconfirmedTxnPool) SpendsForAddresses(bcUnspent *coin.UnspentPool,
-	a map[cipher.Address]byte) coin.AddressUxOuts {
-	auxs := make(coin.AddressUxOuts, len(a))
+func (utp *UnconfirmedTxnPool) SpendsForAddresses(unspent *blockdb.UnspentPool,
+	addrs []cipher.Address) (coin.AddressUxOuts, error) {
+	addrm := make(map[cipher.Address]byte, len(addrs))
```
gz-c: can use `map[cipher.Address]struct{}` here too

src/visor/unconfirmed.go
```diff
@@ -414,50 +415,71 @@ func (utp *UnconfirmedTxnPool) GetKnown(txns []cipher.SHA256) coin.Transactions
...
-				if _, ok := a[ux.Body.Address]; ok {
-					auxs[ux.Body.Address] = append(auxs[ux.Body.Address], ux)
-				}
+			ux, ok, err := unspent.Get(h)
+			if err != nil {
+				return err
+			}
+
+			if !ok {
+				// unconfirm transaction's IN is not in the unspent pool, this should not happen
```
gz-c: log this?

src/visor/visor.go
```diff
@@ -214,10 +208,16 @@ func NewVisor(c Config) (*Visor, VsClose, error) {
 func (vs *Visor) Run(q chan struct{}) {
 	if vs.Blockchain.GetGenesisBlock() == nil {
 		vs.GenesisPreconditions()
-		b := vs.Blockchain.CreateGenesisBlock(
+		b, err := vs.Blockchain.CreateGenesisBlock(
 			vs.Config.GenesisAddress,
 			vs.Config.GenesisCoinVolume,
 			vs.Config.GenesisTimestamp)
+		if err != nil {
+			logger.Error("%v", err)
```
gz-c: `Error(err)`

src/gui/wallet.go
iketheadore: The`logger.Error("save wallet failed when generate new addresses:", err)` will log the message as: `2017/06/17 13:11:29 save wallet failed when generate new addresses:%!(EXTRA *errors.errorString=invalid address)`. The `"github.com/op/go-logging"` version we current use is too old, must have `format` string. We can create an issue to update the logger version and update all logs.

src/visor/blockchain_parser.go
iketheadore: Oh, yes, I will update to only call  `SetParsedHeight` when `parsedHeight <bcHeight`

src/visor/visor.go
iketheadore: This can't parse compile in current logger version

src/visor/blockchain_parser.go
iketheadore: Yes, I will remove it

src/visor/blockdb/unspent.go
iketheadore: I will add a rollback function to handle all problem like this.

src/visor/blockdb/unspent.go
iketheadore: I will change to use `struct{}`

src/visor/rpc.go
iketheadore: We used to have unspent pool in memory and is not thread safe, we return copy to avoid caller from changing, which is not necessary actually. Now our unspent pool is in db, it's thread safe.

src/visor/unconfirmed.go
iketheadore: hmm, I think need to return a error to describe this situation, in case of unexpected error.

src/visor/rpc.go
gz-c: ok, didnt know if it was that, or to prevent accidental modification via the pointer

iketheadore:
@gz-c code is updated, please have a review, thanks.

src/visor/blockdb/unspent.go
```diff
@@ -0,0 +1,297 @@
...
+		return err
+	}
+
+	_, err = tx.CreateBucket(up.meta.Name)
+	return err
+}
+
+func (up *UnspentPool) addWithTx(tx *bolt.Tx, ux coin.UxOut) (err error) {
+	// all updates will rollback if return is not nil
+	// in case of unexpected panic, we must catch it and return error
```
gz-c: what can cause panic here?

iketheadore:
encoder.Serialise

[Yunfei Mao - Contact Using Hop](http://GetHop.com/?_hmid=1497932855)

On June 20, 2017 at 4:25 GMT, gz c <notifications@github.com> wrote:

@gz-c commented on this pull request.
---------------------------------------------------------------

In [src/visor/blockdb/unspent.go](https://github.com/skycoin/skycoin/pull/363#discussion_r122878369):
> + _, err := tx.CreateBucket(up.pool.Name) + if err != nil { + return err + } + + if err := tx.DeleteBucket(up.meta.Name); err != nil { + return err + } + + _, err = tx.CreateBucket(up.meta.Name) + return err +} + +func (up *UnspentPool) addWithTx(tx *bolt.Tx, ux coin.UxOut) (err error) { + // all updates will rollback if return is not nil + // in case of unexpected panic, we must catch it and return error
what can cause panic here?

src/visor/blockdb/unspent.go
```diff
@@ -0,0 +1,297 @@
...
+
+		uxHash, err := meta.getXorHash()
+		if err != nil {
+			return err
+		}
+
+		uxHash = uxHash.Xor(ux.SnapshotHash())
+
+		// update uxhash
+		if err = meta.setXorHash(uxHash); err != nil {
```
gz-c: is the tx implicitly used here because of `tx.Bucket`? or does the tx need to be passed to this method as well

src/visor/blockchain_parser.go
```diff
@@ -93,41 +79,22 @@ func (bcp *BlockchainParser) Stop() {
...
-	return nil
-}
 
-// ParseNotifier sets the callback function that will be invoked
-// when block is parsed.
-func ParseNotifier(f func(height uint64)) ParserOption {
-	return func(p *BlockchainParser) {
-		p.parsedFunc = f
+	if parsedHeight < int64(bcHeight) {
+		return bcp.historyDB.SetParsedHeight(uint64(bcHeight))
```
gz-c: does this need to run in the same transaction as used in ProcessBlock?

iketheadore: Yes, I will move to ProcessBlock

src/visor/blockdb/unspent.go
iketheadore: The ux.Hash may panic, actually lots of function in encoder package might panic and it's function name does not have a `Must` prefix.

src/visor/blockdb/unspent.go
iketheadore: because of the function needs to have a `bolt.Tx` pass in as argument, all this function have `withTx` suffix are run in transaction that might be rollback if any error

iketheadore:
I think it's ready to merge in, @gz-c @skycoin-main

iketheadore:
@gz-c @skycoin-main code is updated, all outputs will be loaded into RAM when node start, and all requests for outputs will go through RAM.

Issue #373, opened 2017-06-23
Master node will shutdown if there're no transactions


iketheadore:
The master node with current codebase will shut down if no transactions exist.

iketheadore:
Fixed in #374

PR #374, opened 2017-06-23
Fix Zero transactions and Zero network bugs

iketheadore:
* Fix bug #373 , should continue instead of returning, which will break the daemon loop.
* Fix bug #375 , will be able to recover network from zero connections.

### DIFF
--- a/src/visor/blockchain.go
+++ b/src/visor/blockchain.go
@@ -228,7 +228,7 @@ func (bc Blockchain) NewBlockFromTransactions(txns coin.Transactions,
 	if len(txns) == 0 {
 		return nil, errors.New("No transactions")
 	}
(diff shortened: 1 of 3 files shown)


src/visor/blockchain.go
```diff
@@ -422,14 +422,14 @@ func (bc Blockchain) GetLastBlocks(num uint64) []coin.Block {
...
-	if arbitrating {
+	if bc.arbitrating {
 		txns = coin.SortTransactions(txns, bc.TransactionFee)
 	}
 	//TODO: audit
 	if len(txns) == 0 {
-		if arbitrating {
+		if bc.arbitrating {
 			return txns, nil
 		}
```
gz-c: when not arbitrating it returns `nil, errors.New("No transactions")`

PR #386, opened 2017-06-26
fix fatal bug that the unit test TestInitDataDirDefault deletes all user's data

zhiyuan2007:
1. TestInitDataDirDefault deletes all  data from the user's home directory (I have beed lost all my files)
2. TestInitDataDir can't passed, also modified two lines

codebugthatdeldata

### DIFF
--- a/src/util/file_test.go
+++ b/src/util/file_test.go
@@ -67,9 +67,8 @@ func TestInitDataDir(t *testing.T) {
 	d := "./.test/test"
 	assertDirNotExists(t, d)
 	dir := InitDataDir(d)
-	assert.Equal(t, dir, d)
 	assertDirExists(t, dir)
-	_, err := os.Stat(d)
+	_, err := os.Stat(dir)
 	assert.Nil(t, err)
 	os.RemoveAll(dir)
 }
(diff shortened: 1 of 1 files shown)


src/util/file_test.go
```diff
@@ -78,11 +77,12 @@ func TestInitDataDirDefault(t *testing.T) {
 	defaultDataDir := ".skycointestXCAWDAWD232232"
 	home := UserHome()
 	assertDirNotExists(t, filepath.Join(home, defaultDataDir))
-	dir := InitDataDir("")
+	dir := InitDataDir(defaultDataDir)
 	assert.NotEqual(t, dir, "")
 	assert.True(t, strings.HasSuffix(dir, defaultDataDir))
```
gz-c: We need to switch this to `require` instead of `assert`, so that it stops here.

src/util/file_test.go
```diff
@@ -78,11 +77,12 @@ func TestInitDataDirDefault(t *testing.T) {
...
 	assertDirExists(t, dir)
 	os.RemoveAll(dir)
```
gz-c: Maybe we should wrap `os.RemoveAll` for the tests, and make sure that aren't removing `/`, `$HOME`, etc

gz-c:
Maybe better off deleting these tests altogether

Issue #360, opened 2017-06-13
Node crash when ctrl+c while synchronizing blocks


iketheadore:
Node crashed when I ctrl+c while blocks are downloading:
```
[skycoin.daemon:INFO] Shutdown pool
[skycoin.daemon:INFO] Saving Peer List
[skycoin.daemon:INFO] Shutdown peers
[skycoin.visor:INFO] close db
unexpected fault address 0x7fd55565838c
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x1 addr=0x7fd55565838c pc=0x463d6a]

goroutine 39 [running]:
runtime.throw(0x88ed62, 0x5)
 /home/atomos/.gvm/gos/go1.7/src/runtime/panic.go:566 +0x95 fp=0xc420429590 sp=0xc420429570
runtime.sigpanic()
 /home/atomos/.gvm/gos/go1.7/src/runtime/sigpanic_unix.go:27 +0x288 fp=0xc4204295e8 sp=0xc420429590
runtime.memmove(0xc4202da050, 0x7fd55565838c, 0x44)
 /home/atomos/.gvm/gos/go1.7/src/runtime/memmove_amd64.s:186 +0x2da fp=0xc4204295f0 sp=0xc4204295e8
```

iketheadore:
This is a graceful shutdown bug, will fix it.

PR #387, opened 2017-06-27
Fix ctrl-c crash node when downloading blocks

iketheadore:
Fix bug #360

### DIFF
--- a/src/daemon/daemon.go
+++ b/src/daemon/daemon.go
@@ -283,7 +283,6 @@ func (dm *Daemon) Shutdown() {
 	dm.Pool.Shutdown()
 	dm.Peers.Shutdown()
 	dm.Visor.Shutdown()
-	gnet.EraseMessages()
 }
 
 // Run main loop for peer/connection management. Send anything to quit to shut it
--- a/src/daemon/daemon.go
+++ b/src/daemon/daemon.go
@@ -206,6 +206,8 @@ type Daemon struct {
 	ipCounts *IPCount
 	// Message handling queue
 	messageEvents chan MessageEvent
+	// quit channel
+	quitC chan chan struct{}
 }
 
 // NewDaemon returns a Daemon with primitives allocated
@@ -241,8 +243,8 @@ func NewDaemon(config Config) (*Daemon, error) {
 		connectionErrors:    make(chan ConnectionError, config.Daemon.OutgoingMax),
 		outgoingConnections: NewOutgoingConnections(config.Daemon.OutgoingMax),
 		pendingConnections:  NewPendingConnections(config.Daemon.PendingMax),
-		messageEvents: make(chan MessageEvent,
-			config.Pool.EventChannelSize),
+		messageEvents:       make(chan MessageEvent, config.Pool.EventChannelSize),
+		quitC:               make(chan chan struct{}),
 	}
 
 	d.Gateway = NewGateway(config.Gateway, d)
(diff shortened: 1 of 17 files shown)


src/daemon/gnet/pool.go
```diff
@@ -239,43 +239,56 @@ func (pool *ConnectionPool) Run(q chan struct{}) {
...
-	pool.addresses = make(map[string]*Connection)
 }
 
 // strand ensures all read and write action of pool's member variable are in one thread.
-func (pool *ConnectionPool) strand(f func()) {
+func (pool *ConnectionPool) strand(f func() error) (err error) {
+	defer func() {
+		// send on closed operation channel will panic.
+		if r := recover(); r != nil {
+			err = ErrConnectionPoolClosed
```
gz-c: log the original recover error?

src/daemon/gnet/pool.go
```diff
@@ -220,16 +221,15 @@ func (pool *ConnectionPool) Run(q chan struct{}) {
...
 			case <-pool.quit:
-				return
+				close(pool.ops)
+				return nil
 			default:
 				// without the default case the select will block.
-				logger.Error("%v", err)
-				close(q)
-				return
+				continue
```
gz-c: Before, upon error, it quit, now it tries again, is that ok? Also should log error?

src/daemon/visor.go
```diff
@@ -241,7 +241,14 @@ func (vs *Visor) RequestBlocksFromAddr(pool *Pool, addr string) error {
 	var err error
 	vs.strand(func() {
 		m := NewGetBlocksMessage(vs.v.HeadBkSeq(), vs.Config.BlocksResponseCount)
-		if !pool.Pool.IsConnExist(addr) {
+
```
gz-c:
```
var exist bool
exist, err = pool.Pool.IsConnExist(addr)
if err != nil {
    return
}
```

src/daemon/daemon.go
```diff
@@ -280,33 +282,36 @@ type MessageEvent struct {
 // over the quit channel provided to Init.  The Daemon run loop must be stopped
 // before calling this function.
 func (dm *Daemon) Shutdown() {
+	// close the daemon loop first
+	q := make(chan struct{}, 1)
```
iketheadore: The `Daemon` loop is listening the `quitC` channel, once received `q` from `quitC`,  will send `struct{}{}` to the `q` and stop the loop. Here we use buffered channel, so that the loop can return immediately, instead of waiting for the `q` channel receiver.

src/daemon/gnet/pool.go
iketheadore: Yes, before, we quit. The service is not closed gracefully by `Shutdown` method if the `q` channel is not closed before `Accept` error occurs, should have log, but need continue to try again.

src/daemon/gnet/pool.go
iketheadore: No need to log the recover, the panic is caused by sending on closed operations channel, and the channel will be closed after the connection pool is shutdown. Every methods calls after the connection pool is closed will cause this panic, but what the caller need to know is the connection pool is closed, don't try again.

src/daemon/visor.go
iketheadore: Oh, yes, will change

iketheadore:
If there're no other problems, we can merge in I think.

PR #388, opened 2017-06-27
Replace gvt with dep

gz-c:
gvt was having trouble and possibly deleted my gopath so I look for replacement.

`dep` (https://github.com/golang/dep) may be added to the go toolchain in the future and works better so far.

Use `dep help` for the manual.

I deleted vendor/ and reset it with `dep init`.  This cleans up a lot of dependencies pulled in from the `dht` package that skycoin does not use anymore.


iketheadore:
let's use dep.

PR #393, opened 2017-06-28
Remove util.DataDir; Remove os.RemoveAll from file tests

gz-c:
Restructure test to only test the folder path concatenation.  Does not test the folder creation, to avoid needing to clean up the folder.

Adds more checks for invalid dir and returns error if invalid.

Related: #386

### DIFF
--- a/src/util/file.go
+++ b/src/util/file.go
@@ -16,9 +16,6 @@ import (
 )
 
 var (
-	// DataDir app folder. A shared global value. Initialize it with InitDataDir.
-	DataDir = ""
-
 	EmptyDirectoryNameError = errors.New("data directory must not be empty")
 	DotDirectoryNameError   = errors.New("data directory must not be equivalent to .")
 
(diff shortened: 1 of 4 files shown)


src/util/file_test.go
```diff
@@ -62,32 +50,60 @@ func cleanup(fn string) {
...
-	dir := InitDataDir(defaultDataDir)
-	assert.NotEqual(t, dir, "")
-	assert.True(t, strings.HasSuffix(dir, defaultDataDir))
-	assertDirExists(t, dir)
-	os.RemoveAll(dir)
+func TestBuildDataDir(t *testing.T) {
+	require.Empty(t, DataDir)
+	defer resetDataDir()
+
+	dir := "./.test-skycoin/test"
```
gz-c: These tests do not create a folder anymore, to avoid needing removal.  So any value for `dir` is ok (except the banned values like `./`).

src/util/file.go
```diff
@@ -10,37 +10,67 @@ import (
...
 	"runtime"
+	"strings"
 
 	logging "github.com/op/go-logging"
 )
 
 var (
-	// DataDir app folder
+	// DataDir app folder. A shared global value. Initialize it with InitDataDir.
 	DataDir = ""
```
gz-c: thanks, I removed it and reverted the return value

src/util/file_test.go
iketheadore: 👍

Issue #383, opened 2017-06-25
Error during installation skycoin source


skybitman:
Following the guide in the main readme.

when I run go command:
_go get -v github.com/skycoin/skycoin/..._

I got an error:

_github.com/skycoin/skycoin (download)
package github.com/skycoin/skycoin/src/consensus/example
	**imports github.com/skycoin/skycoin/src/aether: cannot find package "github.com/skycoin/skycoin/src/aether" in any of:
	/home/username/.gvm/gos/go1.8/src/github.com/skycoin/skycoin/src/aether (from $GOROOT)
	/home/username/.gvm/pkgsets/go1.8/global/src/github.com/skycoin/skycoin/src/aether (from $GOPATH)**
github.com/golang/protobuf (download)
github.com/youtube/vitess (download)_

what am I missing?

iketheadore:
Hi @skybitman , please use `go get -v github.com/skycoin/skycoin`.

skybitman:
@iketheadore still not working,
I get another error:

_can't load package: package github.com/skycoin/skycoin: no buildable Go source files in /home/username/.gvm/pkgsets/go1.8/global/src/github.com/skycoin/skycoin_

gz-c:
can you try again with 
>go get github.com/skycoin/skycoin/...

iketheadore:
This is fixed by #399

PR #399, opened 2017-06-30
Fix go build ./src/... errors

gz-c:
#383 might be fixed after this


iketheadore:
There're conflicts.

gz-c:
#400

PR #427, opened 2017-07-06
Fixed #398 Do not save seeds in DOM

### LINKED ISSUE #398: Do not save seeds in DOM
Don't save seeds in DOM, also remove from DOM when leave the backup page.

vyloy:
Fixed #398 Do not save seeds in DOM


iketheadore:
Remove seed from DOM, not just hide it, we should not expose the seeds in the DOM.

vyloy:
I have checked in chrome developer tools, it will remove seed from DOM.

PR #428, opened 2017-07-06
Prepare for release 0.19.0

iketheadore:


src/gui/static/src/app/app.loadWallet.ts
```diff
@@ -564,8 +570,8 @@ export class LoadWalletComponent implements OnInit {
     getDateTimeString(ts) {
         return moment.unix(ts).format("YYYY-MM-DD HH:mm")
     }
-    getElapsedTime(ts) {
-        return moment().unix() - ts;
+    getElapsedTime() {
+        return moment().unix() - this.blockChain[0].header.timestamp;
```
gz-c: can `blockChain.length` be 0 here?

gz-c: or is this method not necessary now that `elapsedTime` is set as a property

iketheadore: this function is not longer used

iketheadore: will remove this function

PR #436, opened 2017-07-10
Replace aether in consensus pkg with daemon/gnet

iketheadore:


gz-c:
The commits are messed up

iketheadore:
wired, why there're so much commits, should have only one commit

iketheadore:
I will close and clean the commits

PR #438, opened 2017-07-11
Fix hard coded wallet folder api

iketheadore:


iketheadore:
This PR contains commits lots days ago, which is caused by rebase master branch, if you check the master branch, you will see master has abiyt 8 commits ahead of develop branch, which I think was caused by cheery-pick server days ago.

iketheadore:
Use filepath.Join in `GetWalletDir` method?

PR #447, opened 2017-07-16
Refactor /wallet/spend api and bug fixes

### LINKED ISSUE #371: Only one outstanding transaction
Wallet should only has one outstanding transaction to avoid mistakes, and wallet should not have new transaction until the previous unconfirmed transaction has been executed.

TODO:
- [ ] Check if there're unconfirmed pending transactions from local wallet in `/wallet/spend` api

### LINKED ISSUE #440: Add flag after peer connect log to indicator whether it's outgoing or incoming peer
After "connected to peer XYZ",  put "(outgoing)" or "(Incoming" after the peer ip/address

### LINKED ISSUE #443: Error predicted coins in balance api
The predicted balance field in spending result is always the same as confirmed field

image

### LINKED ISSUE #444: Bug in /blockchain/progress api
When start without network, the `highest` value in the response is `current`+1, and it's not the real highest value.

image

iketheadore:
* Refactor /wallet/spend api to Fix #371 , will return only when pending transaction is confirmed.
* Refactor to decouple wallet and visor package
* Fix #440 , will have "outgoing" or "incoming" flag in peer connected log
* Fix #443 , predict balance in /wallet/spend api's response is incorrect
* Fix #444 , bug in /blockchain/progress api
* Update tests that are above related


src/wallet/wallet.go
```diff
@@ -28,3 +36,387 @@ func NewWalletFilename() string {
...
+
+	//create transaction
+	txn.PushOutput(changeAddr, change.Coins, change.Hours)
+	txn.PushOutput(dest, amt.Coins, changeHours/2)
+	txn.SignInputs(toSign)
+	txn.UpdateHeader()
+	return &txn, nil
+}
+
+func createSpends(headTime uint64, uxa coin.UxArray,
```
gz-c: this function does not stop appending `UxOut` to `spending` once it has enough to satisfy amount?  all it seems to do is filter out invalid `UxOut` and return error if sum is less than balance

iketheadore: Yes, wallet will use all unspent outputs in wallet, lol, just copied the function from visor, will change to stop appending when get sufficient unspent outputs.

Issue #449, opened 2017-07-16
[V19 History Screen]address isn't getting searched


candyYMMA:
Scenario:
Perform some sending and receiving activities.
Open skycoin History tab.
Put the address you last got money from
Hit Search button

Expected:
History appears filtered - only transactions with FROM OR TO of the input address appears.

Actual:
The History displayed exactly as before with no filter.
Search isnt working for address 
![hist_tab](https://user-images.githubusercontent.com/28111034/28251055-fa7242c0-6a7d-11e7-895f-75b2586dc3a0.png)
After Searching:
![hist_after_hit_search](https://user-images.githubusercontent.com/28111034/28251068-3ac724a8-6a7e-11e7-8890-243c8849b2ab.png)

iketheadore:
Yes, the search function has broken for a long time.

xano8437:
Can be closed since search bar is removed. If search bar should be re-introduced, please open new request

PR #450, opened 2017-07-19
Fix #448 and #449

### LINKED ISSUE #448: [V19 UI Send Screen]Amount option to insert negative value
Scenario:
Open skycoin Send tab.
No send activity was performed yet.
Clicking on the arrow down at amount field.

Expected:
When clicked arrow down value shoudn't go under zero.
Can display a message to the user.

Actual:
When clicked arrow down at amount field the values become negative
image

vyloy:


gz-c:
New wallet is merged instead

PR #464, opened 2017-08-15
webrpc/CLI refactor

gz-c:
Refactoring to make some CLI functionality usable as a library instead of need to exec skycoin-cli.

Fixes bug with spawning 5 goroutines every webrpc call.

Create `webrpc.Client` to wrap webrpc calls, with tests.

CLI lacks tests. Manually tested every command except for:
* `send`
* `broadcastTx`
* `createRawTx`


src/api/cli/rpc.go
```diff
@@ -0,0 +1,132 @@
...
+	if err := c.Do(&uxouts, "get_address_uxouts", addrs, "1"); err != nil {
+		return nil, err
+	}
+
+	return uxouts, nil
+}
+
+func (c *RpcClient) GetLastBlocks(n uint64) (*visor.ReadableBlocks, error) {
+	if n <= 0 {
+		return nil, errors.New("block number must >= 0")
```
iketheadore: oh, the error should be `errors.New("block number must > 0")`

gz-c:
@iketheadore tell me when you think its ok or if other problems still

iketheadore:
I've checked the code, it's ok, but might have conflict with #447

Issue #445, opened 2017-07-15
[V19 UI Send Screen]Default value of Wallet field at Send screen is wrong.


candyYMMA:
Scenario:
Open skycoin Send tab.
No send activity was performed yet.
take a look at wallet field

Expected:
Value in DDL as default should be "select wallet" so user will know that in this field he should choose from which wallet to send skycoins.

Actual:
Undefined(All) - 0SKY appears.
When opening the DDL youcan only see the "Select Wallet".
user should see at at first and not when open the DDL or "Select Wallet" should be removed from the DDL.

[Link demonstracting the issue](https://ibb.co/hDvLpa)

iketheadore:
Thanks

btw, can upload image directly here

PR #480, opened 2017-08-20
Improve validation

xano8437:
- Do not allow negative values for send skycoin
- Do not allow values higher than wallet balance for send skycoin

Closes https://github.com/skycoin/skycoin/issues/448


xano8437:
Also closes https://github.com/skycoin/skycoin/issues/445

gz-c:
Does the backend also do this validation?

xano8437:
Back-end does validation on max amount, didn't try negative amount

PR #491, opened 2017-08-25
Refactor wallet

### LINKED ISSUE #466: [V19 generate_address Wallets ]Address generated within the same wallet is not unique
Run Wallet app
Blocks not yet fully loaded on main screen to your right
Click on Load wallet from Seed. 
Click on + sign

Expected:
Check another line with address created under your wallet.
Different address should be created

Actual:
Address created is the same value("address 2" & "address 3")

image

iketheadore:
* Remove global wallet variable in gui/http.go, it's not thread safe
* Add wallet service, provides safe apis to access wallet info
* Serve the wallet service through daemon.Gateway, all http apis should serve through it to keep safe, fix #466


src/wallet/service.go
```diff
@@ -0,0 +1,231 @@
...
+	// check dup
+	if id, ok := serv.firstAddrIDMap[w.Entries[0].Address.String()]; ok {
+		return Wallet{}, fmt.Errorf("duplicate wallet with %v", id)
+	}
+
+	if err := serv.wallets.Add(*w); err != nil {
+		return Wallet{}, err
+	}
+
+	if err := w.Save(serv.WalletDirectory); err != nil {
```
gz-c: Remove the added wallet from serv.wallets

src/wallet/service_test.go
```diff
@@ -0,0 +1,430 @@
...
+	rand.Seed(time.Now().Unix())
+}
+
+func prepareWltDir() (string, func()) {
+	tmpDir := os.TempDir()
+	rand.Seed(time.Now().Unix())
+	n := rand.Int31n(100)
+	dir := fmt.Sprintf("%s%d/wallets", tmpDir, n)
+	return dir, func() {
+		os.RemoveAll(dir)
```
gz-c: If its putting the test data in /tmp we should leave it there and not call RemoveAll() in case we have another stupid bug that deletes $HOME

src/wallet/wallets.go
```diff
@@ -139,11 +138,23 @@ func (wlts *Wallets) Remove(id string) {
...
 
+// Update updates the given wallet, return error if not exist
+func (wlts *Wallets) Update(wltID string, updateFunc func(Wallet) Wallet) error {
+	w, ok := (*wlts)[wltID]
+	if !ok {
+		return errWalletNotExist(wltID)
+	}
+
+	newWlt := updateFunc(*w)
+	(*wlts)[wltID] = &newWlt
```
gz-c: is the pointer dereference necessary?

src/wallet/service.go
iketheadore: Why remove the added wallet? if the wallet name is dup, this function should return error, the caller will rename and try again. If the first address is duplicate, we choose the old wallet, cause the wallet entry number in old wallet is >=1, but the new created wallet is 1.

src/wallet/wallets.go
iketheadore: Oh, yes, no need to use pointer (wlts *Wallets), the wallets is a map

src/wallet/service_test.go
iketheadore: I'm afraid of returning  same tmp wallet file, this will fail the test code

src/wallet/service.go
gz-c: oh i thought it would be disk error, then we would show a loaded wallet but it wouldnt be saved, so user would lose it mistakenly

src/wallet/service_test.go
gz-c: oh sorry i misread, i thought you were using this:
https://golang.org/pkg/io/ioutil/#TempDir
https://golang.org/pkg/io/ioutil/#TempFile

if you use this there will be no problem

src/wallet/service.go
iketheadore: Oh, yes, I've updated the code

PR #503, opened 2017-08-29
Fix #501 and #502

### LINKED ISSUE #501: [V20 Wallet] Add /version endpoint
Add /version endpoint to return wallet version info, including version number, HEAD commit id.

### LINKED ISSUE #502: Set default wallet's label as "Your Wallet"
No wallet name will be shown in GUI if it's the first time to run wallet, cause the default wallet label is empty.

iketheadore:
* Fix #501 , add /version endpoint, which will return the current node version number and the HEAD commit id when build the node
* Fix #502 , set default wallet's label as "Your Wallet"

### DIFF
--- a/cmd/skycoin/skycoin.go
+++ b/cmd/skycoin/skycoin.go
@@ -29,7 +29,10 @@ import (
 
 var (
 	// Version node version which will be set when build wallet by LDFLAGS
-	Version    = "0.0.0"
+	Version = "0.20.0"
+	// Commit id
+	Commit = ""
+
 	logger     = logging.MustGetLogger("main")
 	logFormat  = "[skycoin.%{module}:%{level}] %{message}"
 	logModules = []string{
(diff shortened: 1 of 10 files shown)


cmd/skycoin/skycoin.go
```diff
@@ -29,7 +29,10 @@ import (
 
 var (
 	// Version node version which will be set when build wallet by LDFLAGS
-	Version    = "0.0.0"
+	Version = "0.20.0"
```
gz-c: maybe can add "-dev" to this version

src/util/file/file.go
```diff
@@ -35,6 +35,12 @@ func InitDataDir(dir string) (string, error) {
 		return "", err
 	}
 
+	// check if dir already exist
+	if _, err := os.Stat(dir); !os.IsNotExist(err) {
```
gz-c: this also needs to check the `stat.IsDir()`

src/wallet/addresses.go
```diff
@@ -59,7 +59,7 @@ func GetSkycoinWalletEntry(pub cipher.PubKey, sec cipher.SecKey) ReadableEntry {
 	}
 }
 
-// Returns a ReadableEntry in Bitcoin format
+// GetBitcoinWalletEntry returns a ReadableEntry in Bitcoin format
```
gz-c: Does Golint require the function name repeated in the comment string?

cmd/skycoin/skycoin.go
iketheadore: yes, I was thinking add suffix as well.

src/wallet/addresses.go
iketheadore: Yes, every exported function's comment should start with the function name

src/util/file/file.go
iketheadore: ok, I will add the check

PR #506, opened 2017-09-01
Fix /wallet/newAddress bug

iketheadore:
Only the first generated address is saved since wallet start


gz-c:
Did you intend to close this?

iketheadore:
No, it's a mistake

gz-c:
Originally the wallet manager was meant to return only copies to avoid accidental modification. Make sure the code still is safe

iketheadore:
Yes, it's still safe now.

Issue #509, opened 2017-09-03
Verify signatures failed when node start


iketheadore:
The following error might occur when node upgrade from 0.18.0 to 0.19.0+,
```
[skycoin.main:ERROR] Invalid block signatures: Verify signature of block in depth: 
2708 failed: no sig for <SIG>
```
To fix this issue, we need to parse the block again to fill the lost signature when node start.

gz-c:
#512 fixes as much as it can be fixed

@iketheadore is it possible to detect when the data.db is corrupted and recreate it automatically?

iketheadore:
We can check and recreate the data.db if it's corrupted, but need to do this before other modules are loaded.

gz-c:
Might be important for next release, we still have people complaining about the error and they dont understand data.db deletion / hidden folder and might delete their wallet files too on accident

PR #512, opened 2017-09-06
Make block execute process atomic

iketheadore:
Our block execute process is not atomic, if the node closed unexpectedly when synchronizing the blocks, we might lost block signature or leaving used unconfirmed transaction in the pool. Make the execute block process atomic, and avoid #509 issue in the future.

### DIFF
--- a/src/visor/visor.go
+++ b/src/visor/visor.go
@@ -177,7 +177,7 @@ func NewVisor(c Config) (*Visor, VsClose, error) {
 	// var verifyOnce sync.Once
 	bp := NewBlockchainParser(history, bc)
 
-	bc.BindListener(bp.BlockListener)
+	bc.BindListener(bp.FeedBlock)
 
 	wltServ, err := wallet.NewService(c.WalletDirectory)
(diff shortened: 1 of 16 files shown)


src/visor/visor.go
```diff
@@ -239,11 +241,18 @@ func (vs *Visor) Run() error {
...
+		noSigBlockSeqs, err := vs.Blockchain.VerifySigs(vs.Config.BlockchainPubkey, vs.blockSigs)
+		if err != nil {
 			// TODO: if detect signature lost, we need to parse the specific block again to get signature.
-			errC <- fmt.Errorf("Invalid block signatures: %v", err)
+			errC <- fmt.Errorf("invalid block signatures: %v", err)
+			return
+		}
+
+		if len(noSigBlockSeqs) > 0 {
+			errC <- fmt.Errorf("no signature found in block %d", noSigBlockSeqs)
```
gz-c: wrong formatter %d

iketheadore: yes, will change to %v

PR #514, opened 2017-09-08
Check block signatures before node start

### LINKED ISSUE #509: Verify signatures failed when node start

iketheadore:
Remove the data.db and pull blocks again if found  signature lost.

This should be able to fix #509.

### DIFF
--- a/src/visor/blockchain.go
+++ b/src/visor/blockchain.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"sync"
-	"time"
 
 	"github.com/boltdb/bolt"
 	"github.com/skycoin/skycoin/src/cipher"
@@ -658,9 +657,6 @@ func (bc *Blockchain) VerifySigs() error {
 	if head == nil {
 		return nil
 	}
-
-	start := time.Now()
-	defer logger.Info("verify signature: %v block size: %v", time.Since(start), head.Seq())
 
 	seqC := make(chan uint64)
 
--- a/src/visor/blockchain.go
+++ b/src/visor/blockchain.go
@@ -705,10 +705,7 @@ func (bc *Blockchain) sigVerifier(seqC chan uint64) (func(), <-chan error) {
 	for i := 0; i < SigVerifyTheadNum; i++ {
 		wg.Add(1)
 		go func(id int) {
-			defer func() {
-				wg.Done()
-				logger.Critical("%d done", id)
-			}()
+			defer wg.Done()
 			for {
 				select {
 				case seq := <-seqC:
--- a/src/visor/blockchain.go
+++ b/src/visor/blockchain.go
@@ -3,6 +3,8 @@ package visor
 import (
 	"errors"
 	"fmt"
+	"sync"
+	"time"
 
 	"github.com/boltdb/bolt"
 	"github.com/skycoin/skycoin/src/cipher"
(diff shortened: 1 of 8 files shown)


src/visor/blockchain.go
```diff
@@ -615,36 +675,77 @@ func (bc Blockchain) TransactionFee(t *coin.Transaction) (uint64, error) {
...
+	for i := 0; i < SigVerifyTheadNum; i++ {
+		wg.Add(1)
+		go func(id int) {
+			defer wg.Done()
+			for {
+				select {
+				case <-c:
+					return
+				case seq := <-seqC:
+					b := bc.GetBlockInDepth(seq)
```
gz-c: Leave this as a separate function that can be call to check a block seq

src/visor/visor.go
```diff
@@ -157,19 +157,32 @@ func NewVisor(c Config) (*Visor, VsClose, error) {
...
+			return nil, nil, err
+		}
+		logger.Error("signature lost...")
+		closeDB()
 
-	// creates block signature bucket
-	sigs, err := blockdb.NewBlockSigs(db)
-	if err != nil {
-		return nil, nil, err
+		// remove the db
```
gz-c: can you make this logic a separate function

src/visor/blockchain.go
```diff
@@ -337,6 +370,27 @@ func (bc *Blockchain) ExecuteBlock(tx *bolt.Tx, b *coin.Block) error {
 	return nil
 }
 
+// ExecuteBlock attempts to append block to blockchain
+func (bc *Blockchain) ExecuteBlock(sb *coin.SignedBlock) error {
```
gz-c: If only test uses this, remove it, have test use bolt.Tx

src/visor/unconfirmed.go
```diff
@@ -339,28 +338,30 @@ func (utp *UnconfirmedTxnPool) InjectTxn(bc *Blockchain, t coin.Transaction) (bo
...
 		now := utc.Now().UnixNano()
 		tx.Received = now
 		tx.Checked = now
-
-		if unspentNotExistErr == nil {
-			tx.IsValid = 1
-		} else {
-			tx.IsValid = 0
-		}
+		tx.IsValid = 1
```
gz-c: We don't save unconfirmedtxns missing unspents now?

src/visor/visor.go
```diff
@@ -198,13 +182,48 @@ func NewVisor(c Config) (*Visor, VsClose, error) {
...
+// the db and create an empty blockchain.
+func load(dbpath string, pubkey cipher.PubKey, arbitrating bool) (*bolt.DB, *Blockchain, error) {
+	// creates blockchain instance
+	db, err := openDB(dbpath)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	bc, err := NewBlockchain(db, pubkey, Arbitrating(arbitrating))
+	if err != nil {
```
gz-c: This should only remove db if the error type is invalid db type error

src/visor/unconfirmed.go
iketheadore: 1. Yes, we don't save the tx that missing unspents now, actually we will check unconfirmed txs and remove those missing unspents txns when node start. 
2. VerifyTransactionFee method won't return ErrUnspentNotExist, and no one returns `ErrUnspentNotExist`.

src/visor/visor.go
```diff
@@ -198,13 +182,52 @@ func NewVisor(c Config) (*Visor, VsClose, error) {
...
+		if err != ErrSignatureLost {
+			return nil, nil, err
+		}
+
+		logger.Error("%v", err)
+		db.Close()
+
```
gz-c: Log message should say that it is removing database because invalid signatures

PR #524, opened 2017-09-14
Test signature lost db recreation

### LINKED ISSUE #523: Test signature verification and db recreation on failure
The database signatures are verified on start and the db is recreated if they are corrupted.

Need to add a unit test for this, using a corrupted db file.

gz-c:
Changed the `load()` logic to save a copy of the bad db file as `$file.corrupt.$hash` where hash is the first 8 bytes of the sha1 hash of the file, base64 encoded.  So if we encounter it again we can get a copy.

Also update main README go setup instructions, added table of contents.

Fixes #523


README.md
```diff
@@ -17,30 +17,83 @@ Skycoin is small part of OP Redecentralize and OP Darknet Plan.
...
+### Installing Go
+
+[Install go1.9+](https://golang.org/doc/install)
+
+### Set GOPATH in your environment
+
+Create a folder for your `$GOPATH`, for example:
+
+```sh
+mkdir ~/.gopath
```
iketheadore: why use hiden folder: ~/.gopath, the default GOPATH is $HOME/go since go1.8.

gz-c: oh didn't know they default now

gz-c: you don't need to set gopath anymore?

gz-c:
updated readme

Issue #530, opened 2017-09-19
Upgrade address generator tool


iketheadore:
- [ ] use bip39 as default seed and having option to use hex string as seed

- [ ] add option to only show address list, do not show the seed, pubkey, seckey, etc.

gz-c:
What is hex string seed for?

PR #531, opened 2017-09-19
Fix address generator issue #530

iketheadore:


cmd/address_gen/address_gen.go
```diff
@@ -24,6 +25,8 @@ func main() {
 	genCount := flag.Int("n", 1, "Number of addresses to generate")
 	hideSecKey := flag.Bool("s", false, "Hide the secret key from the output")
 	isBitcoin := flag.Bool("b", false, "Print address as a bitcoin address")
+	hexSeed := flag.Bool("x", false, "Use hex string as seed")
+	verbose := flag.Bool("v", false, "Show verbose info of generated addresses")
```
gz-c: Change `verbose` to `j,json`?

cmd/address_gen/address_gen.go
```diff
@@ -24,6 +25,8 @@ func main() {
 	genCount := flag.Int("n", 1, "Number of addresses to generate")
 	hideSecKey := flag.Bool("s", false, "Hide the secret key from the output")
 	isBitcoin := flag.Bool("b", false, "Print address as a bitcoin address")
+	hexSeed := flag.Bool("x", false, "Use hex string as seed")
```
gz-c: What is hex string seed for

cmd/address_gen/address_gen.go
iketheadore: It's for showing details of generated address info, including pubkey, seckey, seed, etc. So I think verbose is ok.

cmd/address_gen/address_gen.go
iketheadore: the generator use random hex string as default seed before, we just keep it here.

PR #533, opened 2017-09-19
Option to disable daemon.Visor networking

gz-c:
* Disables daemon.Visor networking functionality when daemon.Config.Disabled is true.
* Fix missing strand() call for AnnounceBlocks [fixed by #534 instead]
* Misc cleanups

The intent of this is to support the `-disable-networking` option, so that the node can be run in offline mode, but still accessed over the localhost APIs for some operations.

I don't want to merge this until after 0.20 is released in case of bugs.
Still needs to be tested that CLI/API methods work when daemon.Visor disabled, or return a correct error message (such as when using cli's broadcastTransaction, it should notify the user that networking is disabled).


gz-c:
See #541

PR #536, opened 2017-09-21
Add make run, run-help, test, lint, check

gz-c:


cmd/skycoin/skycoin.go
```diff
@@ -158,6 +160,7 @@ type Config struct {
 }
 
 func (c *Config) register() {
+	flag.BoolVar(&help, "help", false, "Show help")
```
iketheadore: flag package has default `-h` and `--help` flag, do we need to add a new help flag?

gz-c: If you don't set it then it exits with exit status 2 when using `-h`, `--help`, then `make` says failure

PR #564, opened 2017-10-03
Fix issue562

iketheadore:


src/visor/blockchain.go
```diff
@@ -362,11 +362,11 @@ func (bc Blockchain) GetLastBlocks(num uint64) []coin.SignedBlock {
 	}
 
 	end := bc.HeadSeq()
-	start := end - num + 1
+	start := int(end-num) + 1
```
gz-c: Does it need to be `int(end) - int(num)` or does go do it that way automatically?

iketheadore: Go will do it automatically

Issue #555, opened 2017-10-01
Apply new branding to v20


gz-c:
Need new icon swapped in, and new colors used

Issue #563, opened 2017-10-03
[wallet20-dev] explorer tab can't show genesis block


iketheadore:
Will show this in explorer tab when click the genesis block:

![image](https://user-images.githubusercontent.com/2589926/31136008-7620536c-a82c-11e7-848f-1e56d071091c.png)

montycrypto:
Solved in https://github.com/skycoin/skycoin/pull/565

PR #565, opened 2017-10-05
Apply new branding to v20

montycrypto:
Closes https://github.com/skycoin/skycoin/issues/555


montycrypto:
Also closes #563 now

iketheadore:
image

The logo and version number `0.20.0-dev` should be in the same horizontal line I think, and the white color `Skycoin` in the logo looks different from the version number as well.

gz-c:
The whole UI is being redesigned, so will not worry about its appearance yet

PR #575, opened 2017-10-10
Genere prod build for GUI

montycrypto:


montycrypto:
Improve dist files with latest angular cli

PR #578, opened 2017-10-10
Optimize flow and add yarn to gui

montycrypto:


montycrypto:
- Add yarn
- Move material icons to assets
- Update readme

Issue #535, opened 2017-09-21
daemon.Visor context deadline exceeded


iketheadore:
The expire time of `strand` method in daemon.Visor is `3 seconds`, the data protected under strand might be in unsafe state if timeout. We need to remove the timeout, and keep it thread safe.

PR #589, opened 2017-10-13
Remove context timeout

iketheadore:
Fix issue #535


gz-c:
i think i removed this in another pr that is waiting, i will update those

iketheadore:
ok, I will close this PR

PR #592, opened 2017-10-20
Fix blocks loading issue in windows

iketheadore:
The main thing that block the sync process is blockchain call notify method in bolt.Update, which will push the new block to a channel of history parser, the history parser also need to enter the bolt.Update method to process new block, if the channel is full and history parser is waiting to enter bolt.Update method, then will have a deadlock, cause blockchain won't leave bolt.Update if the notify method is blocked.

To fix this issue, just move the new block notification out of bolt.Update.

### DIFF
--- a/src/daemon/gnet/pool.go
+++ b/src/daemon/gnet/pool.go
@@ -179,6 +179,7 @@ type ConnectionPool struct {
 	ops chan func()
 	// quit channel
 	quit chan struct{}
+	wg   sync.WaitGroup
 }
 
 // NewConnectionPool creates a new ConnectionPool that will listen on Config.Port upon
@@ -211,10 +212,9 @@ func (pool *ConnectionPool) Run() error {
 
 	pool.listener = ln
 
-	var wg sync.WaitGroup
-	wg.Add(1)
+	pool.wg.Add(1)
 	go func() {
-		defer wg.Done()
+		defer pool.wg.Done()
 		for {
 			select {
 			case <-pool.quit:
@@ -244,9 +244,13 @@ loop:
 			}
 		}
 
-		go pool.handleConnection(conn, false)
+		pool.wg.Add(1)
+		go func() {
+			defer pool.wg.Done()
+			pool.handleConnection(conn, false)
+		}()
 	}
-	wg.Wait()
+	pool.wg.Wait()
 	return nil
 }
 
@@ -346,9 +350,10 @@ func (pool *ConnectionPool) handleConnection(conn net.Conn, solicited bool) {
 
 	wg := sync.WaitGroup{}
 	wg.Add(1)
+	qc := make(chan struct{})
 	go func() {
 		defer wg.Done()
-		if err := pool.readLoop(c, msgC); err != nil {
+		if err := pool.readLoop(c, msgC, qc); err != nil {
 			errC <- err
 		}
 	}()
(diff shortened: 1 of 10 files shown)


src/daemon/gnet/pool.go
```diff
@@ -331,36 +338,34 @@ func (pool *ConnectionPool) handleConnection(conn net.Conn, solicited bool) {
...
+	wg.Add(1)
 	go func() {
+		defer wg.Done()
 		errC <- readLoop(c, pool.Config.ReadTimeout, pool.Config.MaxMessageLength, msgC)
 	}()
 
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		errC <- pool.sendLoop(c, pool.Config.WriteTimeout)
```
gz-c: Should only send err if not nil

src/daemon/gnet/pool.go
```diff
@@ -331,36 +338,34 @@ func (pool *ConnectionPool) handleConnection(conn net.Conn, solicited bool) {
...
 					return
 				}
 
-				if err := pool.updateLastSent(c.Addr(), Now()); err != nil {
-					errC <- err
-					return
-				}
-			case msg := <-msgC:
 				if err := pool.receiveMessage(c, msg); err != nil {
 					errC <- err
```
gz-c: It can return here after error write. Don't need qc/c

src/daemon/gnet/pool.go
```diff
@@ -373,7 +378,7 @@ func (pool *ConnectionPool) handleConnection(conn net.Conn, solicited bool) {
 	q := make(chan struct{}, 1)
 	qc <- q
 	<-q
-
+	wg.Wait()
```
gz-c: Can remove qc/q, instead `close(msgC)` after err read from `errC`

src/daemon/gnet/pool.go
```diff
@@ -424,11 +429,37 @@ func readLoop(conn *Connection, timeout time.Duration, maxMsgLen int, msgChan ch
...
+		case pool.SendResults <- sr:
+		case <-time.After(3 * time.Second):
+			logger.Warning("push send result channel timeout")
+		}
+
+		if err != nil {
+			return err
+		}
+
+		if err := pool.updateLastSent(conn.Addr(), Now()); err != nil {
```
gz-c: This updates last sent even if timed out

src/daemon/gnet/pool.go
```diff
@@ -424,11 +429,37 @@ func readLoop(conn *Connection, timeout time.Duration, maxMsgLen int, msgChan ch
...
+	for m := range conn.WriteQueue {
+		if m == nil {
+			continue
+		}
+
+		err := sendMessage(conn.Conn, m, timeout)
+		sr := newSendResult(conn.Addr(), m, err)
+		select {
+		case pool.SendResults <- sr:
+		case <-time.After(3 * time.Second):
```
gz-c: timeout should be const at top of file

src/daemon/visor.go
```diff
@@ -116,7 +116,9 @@ func (vs *Visor) Run() error {
 			return err
 		case req := <-vs.reqC:
 			func() {
-				req()
+				cxt, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Minute))
```
gz-c: I thought context deadline exiting was unsafe?

src/daemon/gnet/pool.go
iketheadore: `close(msgC)` might cause `panic: send on closed channel`  in readLoop

src/daemon/visor.go
iketheadore: Yes, hmm, might have thread safe problem, will remove this, lol

src/daemon/gnet/pool.go
iketheadore: The `timeout` here is caused by sending `SendResult` to daemon, the data actually have been sent if the err is nil, so it should update the last sent of this connection.

src/daemon/gnet/pool.go
gz-c: readLoop needs to check a quit channel then, and close msgC when its done

PR #611, opened 2017-11-05
skycoin crashed when getblocksbyseq

zhiyuan2007:
skycoin crashed when run command getblocksbyseq and seq > blockcount,  because return value is incorrect, error should not nil

skycoin-crashed


iketheadore:
Do not return that error, the `GetBlocksInDepth` need to check the nil, and `GetBlockBySeq` is designed to return nil on not found.

zhiyuan2007:
crashed have to handled by other manner

gz-c:
The caller of `GetBlockBySeq` needs to check for `nil` blocks

gz-c:
#617 #616

Issue #616, opened 2017-11-12
[webrpc] get_blocks_by_seq will crash the node if required seq > blockchain height


iketheadore:

gz-c:
@iketheadore is this fixed by #617 ?

iketheadore:
yes, this is fixed by #617.

PR #617, opened 2017-11-12
Fix issue #616

iketheadore:


src/daemon/gateway.go
```diff
@@ -198,11 +198,16 @@ func (gw *Gateway) GetBlocksInDepth(vs []uint64) (*visor.ReadableBlocks, error)
 
 	gw.strand("GetBlocksInDepth", func() {
 		for _, n := range vs {
-			b, err := gw.vrpc.GetBlockBySeq(gw.v, n)
```
gz-c: Leave `err`, change `:=` to `=`, add `var b *coin.SignedBlock`

PR #622, opened 2017-11-13
Refactor visor to use bolt.Tx everywhere; add unparam linter

### LINKED ISSUE #621: Replace UnspentPool struct cache with bolt.DB
Mixing the struct cache with bolt.DB transaction is complex, convert to bolt.DB

gz-c:
* Refactor visor to use `bolt.Tx` everywhere

Systematic use of `bolt.Tx` is applied to the `visor` package. `bolt.Tx` must be used for all database operations; never use a bucket directly.

`visor.Visor` is responsible for starting all DB transactions. The exceptions are in constructors to create and/or initialize buckets, or in tests.

Care must be taken not to call multiple transaction-based `visor.Visor` methods in the same function from an external package. For example, if a `daemon.Gateway` method needs to call 2 `visor.Visor` method, you must create a new method on `visor.Visor` that wraps both in a transaction.  Otherwise, two separate transactions will be used which can cause subtle bugs.

External packages should not be calling methods underneath `visor.Visor`, e.g. they should not call `visor.Unconfirmed.Foo(tx)`.  They can't call it, because it needs a `bolt.Tx`, and only `visor.Visor` has a `bolt.DB` to create a `bolt.Tx`.  Instead, add a wrapper method `visor.Visor.Foo()` that performs the boltdb transaction.

The `bucket.Bucket` database abstraction layer is replaced with the `dbutil` package, which has wrappers for accessing a bucket's methods through a `bolt.Tx`.

* Remove unspent and blockchain cache

The extra in-memory cache for unspents and blockchain complicates things and probably has some edge cases that could result in deadlocks.  These are removed in favor of bolt.DB directly.  Fixes #621 

* Add `unparam` linter to discover unused arguments

Note: `unparam` is too aggressive in some cases, use `// nolint: unparam` to disable on certain lines.  Do not use `// nolint: unparam` if not necessary. 

Unparam has 3 kinds of errors:

* `parameter foo is unused` [argument is not used]
* `result foo (type) is never used` [return value is not used]
* `parameter foo always receives bar` [this function is always called with the same argument]

Never ignore the first one, unless it is a mock struct method in tests to satisfy an interface.
The other two can be ignored in test code, but never ignored in live code.

Fixes #621


src/visor/visor.go
```diff
@@ -339,210 +409,390 @@ func (vs *Visor) SignBlock(b coin.Block) coin.SignedBlock {
...
+		}
+
+		headSeq, ok, err := vs.Blockchain.HeadSeq(tx)
+		if err != nil {
+			return err
+		} else if !ok {
+			return errors.New("Blockchain is empty but history has transactions")
+		}
+
+		b, err := vs.GetSignedBlockBySeq(htxn.BlockSeq)
```
iketheadore: is it ok to call `vs.GetSignedBlockBySeq` in `vs.DB.View`?, the `vs.GetSignedBlockBySeq` method will call `vs.DB.View` as well.

mvdan:
@gz-c if you have found any unparam false positives, please do report them - it will help make the tool better :) I'm not familiar with the codebase, so it is harder for me to tell why a warning would be a false positive.

Also note that I've made some improvements in the past week, so some of them may have already been fixed.

gz-c:
@mvdan There weren't any false positives but we're primarily looking for unused function arguments because of a bug due to a shadowed function argument.  The majority of unparam warnings were "result is never used" or "always receives" warnings in test code helper methods, which I think are fine to ignore.  I would disable these warnings for our linting if there was the option, since they seem more like code cleanliness than bugs.  Or I might run `gometalinter` and `gometalinter --tests` separately if the others have trouble with the warnings (I don't want others to get in the habit of using `// nolint`).

src/visor/visor.go
gz-c: It might work if it uses an RLock but still, not ok, inside these visor methods it should call the storage's`*tx` method and not another top-level visor method

mvdan:
@gz-c you can disable this in the tool directly via `unparam -tests=false`, which will ignore the test files - it won't even parse nor type-check them.

gz-c:
Replaced by #1440

PR #623, opened 2017-11-14
Fix cli insufficient balance error message

iketheadore:
Differ insufficient balance in wallet and insufficient confirmed balance, cli should give different message to indicate whether the wallet does really not have sufficient balance, or just the balance have not confirmed yet.

### DIFF
--- a/src/api/cli/create_rawtx.go
+++ b/src/api/cli/create_rawtx.go
@@ -425,8 +425,10 @@ func getKeys(wlt *wallet.Wallet, outs []UnspentOut) ([]cipher.SecKey, error) {
 }
 
 func getSufficientUnspents(unspents *webrpc.OutputsResult, coins uint64) ([]UnspentOut, error) {
+	// get spendable outputs, which are all confirmed outputs without
+	// the spending outputs that are in unconfirmed tx pool.
 	spendableOuts := unspents.Outputs.SpendableOutputs()
-	var spendCoins uint64
+	var spendableCoins uint64
 	var spendOuts []UnspentOut
 	for i, out := range spendableOuts {
 		c, err := droplet.FromString(out.Coins)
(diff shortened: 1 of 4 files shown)


src/api/cli/create_rawtx.go
```diff
@@ -430,36 +424,36 @@ func getKeys(wlt *wallet.Wallet, outs []UnspentOut) ([]cipher.SecKey, error) {
...
+	unconfirmedBalance, err := unconfirmedOuts.Balance()
+	if err != nil {
+		return nil, fmt.Errorf("get unconfirmed balance failed: %v", err)
+	}
 
-			if totalCoins >= coins {
-				return outs, nil
-			}
-		}
+	if unconfirmedBalance.Coins+spendCoins < coins {
```
gz-c: `if unconfirmedBalance.Coins+spendCoins > coins` ?

iketheadore: oh, should rename the unconfirmedBalance to unconfirmedIncoming, and spendCoins to spendableCoins

gz-c: Can you fix it, the code seems wrong

src/api/cli/create_rawtx.go
```diff
@@ -430,36 +424,39 @@ func getKeys(wlt *wallet.Wallet, outs []UnspentOut) ([]cipher.SecKey, error) {
...
+	uncfmIncoming, err := uncfmIncomingOuts.Balance()
+	if err != nil {
+		return nil, fmt.Errorf("get unconfirmed balance failed: %v", err)
+	}
 
-			if totalCoins >= coins {
-				return outs, nil
-			}
-		}
+	if spendableCoins+uncfmIncoming.Coins < coins {
```
gz-c: Are you sure these two errors are not in the wrong order?

iketheadore: Yes, if the spendable coins plus the unconfirmed incoming coins still less than the coins you want to spend, that means you don't have enough coins in wallet.

PR #624, opened 2017-11-18
Add /wallet/load endpoint to load wallet from seed and scan ahead N addresses

### LINKED ISSUE #597: Importing wallet from seed only imports the original address.
I encountered an issue today when importing a wallet from seed. It was showing a zero balance because it only imported one address, not the second address that I have within the same wallet.

On import, we need to check up to N addresses on import to see if there's more than one address per wallet and, if so, import them all.

iketheadore:
Add `/wallet/load` endpoint, which will load wallet from given seed and scan ahead N addresses to check balance. 

Arguments:
    `seed`: wallet seed
    `label`: wallet label
    `scan-num`: the ahead number of address that will scan, if not specified, we will use 100 by default.

Fixes #597

### DIFF
--- a/src/coin/outputs.go
+++ b/src/coin/outputs.go
@@ -159,14 +159,23 @@ func (ua UxArray) Swap(i, j int) {
 	ua[i], ua[j] = ua[j], ua[i]
 }
 
-// CoinHours returns the total coins and hours
-func (ua UxArray) CoinHours(headTime uint64) (coins uint64, hours uint64) {
+// Coins returns the total coins
+func (ua UxArray) Coins() uint64 {
+	var coins uint64
 	for _, ux := range ua {
 		coins += ux.Body.Coins
-		hours += ux.CoinHours(headTime)
 	}
 
-	return
+	return coins
+}
+
+// CoinHours returns the total coin hours
+func (ua UxArray) CoinHours(headTime uint64) uint64 {
+	var hours uint64
+	for _, ux := range ua {
+		hours += ux.CoinHours(headTime)
+	}
+	return hours
 }
 
 // AddressUxOuts maps address with uxarray
--- a/src/coin/outputs.go
+++ b/src/coin/outputs.go
@@ -159,6 +159,16 @@ func (ua UxArray) Swap(i, j int) {
 	ua[i], ua[j] = ua[j], ua[i]
 }
 
+// CoinHours returns the total coins and hours
+func (ua UxArray) CoinHours(headTime uint64) (coins uint64, hours uint64) {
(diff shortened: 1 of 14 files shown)


src/wallet/service.go
```diff
@@ -92,9 +97,71 @@ func (serv *Service) CreateWallet(wltName string, options ...Option) (Wallet, er
 	return *w, nil
 }
 
+// LoadAndScanWallet loads wallet from seed and scan the first N address
+func (serv *Service) LoadAndScanWallet(wltName string, seed string, scanN uint64, bg BalanceGetter, options ...Option) (Wallet, error) {
+	ops := make([]Option, 0, len(serv.options)+len(options))
+	ops = append(ops, serv.options...)
+	ops = append(ops, options...)
+	ops = append(ops, OptSeed(seed))
```
gz-c: We need to remove `Option` later and use config struct

src/wallet/service.go
```diff
@@ -92,9 +97,71 @@ func (serv *Service) CreateWallet(wltName string, options ...Option) (Wallet, er
...
+	for i := len(bals) - 1; i >= 0; i-- {
+		if bals[i].Confirmed.Coins > 0 || bals[i].Predicted.Coins > 0 {
+			keepNum = uint64(i + 1)
+			break
+		}
+	}
+
+	// reset the wallet if scan number > 1 and not equal to the keep number
+	if scanN > 1 && keepNum != scanN {
+		// reset wallet and
```
gz-c: extra comment

src/wallet/wallet.go
```diff
@@ -283,6 +287,12 @@ func (wlt *Wallet) AddEntry(entry Entry) error {
 	return nil
 }
 
+// Reset resets the wallet entries and move the lastSeed to origin
+func (wlt *Wallet) Reset() {
+	wlt.Entries = wlt.Entries[0:0]
```
gz-c: Allocate new array

gz-c:
Update api README

src/coin/outputs.go
```diff
@@ -159,6 +159,16 @@ func (ua UxArray) Swap(i, j int) {
 	ua[i], ua[j] = ua[j], ua[i]
 }
 
+// CoinHours returns the total coins and hours
+func (ua UxArray) CoinHours(headTime uint64) (coins uint64, hours uint64) {
```
gz-c: make two methods,  `Coins` and `CoinHours`

src/wallet/service.go
gz-c: remove this `// reset wallet and` comment

PR #625, opened 2017-11-19
Expose logs over API

hanyh2004:
add log http interface to front gui
provides 4 args:
1,lines
how many lines to fetch
curl "http://127.0.0.1:6420/logs?lines=20
2,include
must include in log
curl "http://127.0.0.1:6420/logs?include=keyword

3,exclude
must not include in log
curl "http://127.0.0.1:6420/logs?exclude=sss
4,reg
curl "http://127.0.0.1:6420/logs?reg=regexpr

### DIFF
--- a/cmd/skycoin/skycoin.go
+++ b/cmd/skycoin/skycoin.go
@@ -318,7 +318,7 @@ var devConfig = Config{
 	// Will force it to connect to this ip:port, instead of waiting for it
 	// to show up as a peer
 	ConnectTo:   "",
-	LogBuffSize: 16384, //2*1024*8
+	LogBuffSize: 8388608, //1024*1024*8
 }
 
 // Parse prepare the config
@@ -584,11 +584,9 @@ func Run(c *Config) {
 					logger.Info("Logbuff service closed normally")
 					return
 				case <-time.After(1 * time.Second): //insure logbuff size not exceed required size, like lru
-					logger.Debug("logbuffer size %d", buf.Len())
 					for buf.Len() > c.LogBuffSize {
 						_, err := buf.ReadString(byte('\n')) //discard one line
 						if err != nil {
-							logger.Info("read logbuffer err %v", err)
 							continue
 						}
 					}
--- a/cmd/skycoin/skycoin.go
+++ b/cmd/skycoin/skycoin.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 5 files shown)


cmd/skycoin/skycoin.go
```diff
@@ -562,6 +571,32 @@ func Run(c *Config) {
...
+	}
+	if c.Logtogui {
+		go func(buf *bytes.Buffer, quit chan struct{}) {
+			for {
+				select {
+				case <-quit:
+					logger.Info("Logbuff service closed normally")
+					return
+				case <-time.After(1 * time.Second): //insure logbuff size not exceed required size, like lru
+					logger.Debug("logbuffer size %d", buf.Len())
```
gz-c: This log line adds too much noise, can you remove

cmd/skycoin/skycoin.go
```diff
@@ -312,7 +317,8 @@ var devConfig = Config{
 	HTTPProf: false,
 	// Will force it to connect to this ip:port, instead of waiting for it
 	// to show up as a peer
-	ConnectTo: "",
+	ConnectTo:   "",
+	LogBuffSize: 16384, //2*1024*8
```
gz-c: I think 1MB might be better size

src/gui/http.go
```diff
@@ -252,3 +257,84 @@ func versionHandler(gateway *daemon.Gateway) http.HandlerFunc {
...
+			}
+		}
+		keyword := r.FormValue("include")
+		excludeKeyword := r.FormValue("exclude")
+		regstr := r.FormValue("reg")
+		logs := []string{}
+
+		reg, err := regexp.Compile(regstr)
+		if err != nil {
+			logger.Info("reg express error:%v", err)
```
gz-c: Should be `wh.Error400` if user provides bad regex

gz-c: Regex is not really necessary for this though, you can remove it

src/gui/http.go
```diff
@@ -252,3 +257,84 @@ func versionHandler(gateway *daemon.Gateway) http.HandlerFunc {
...
+}
+func getLogsHandler(logbuf *bytes.Buffer) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			wh.Error405(w)
+			return
+		}
+
+		var err error
+		defaultLineNum := 10 // default line numbers
```
gz-c: Default should be much larger, at least 1000

gz-c:
The main purpose of this is so that we can add a debug panel to the GUI wallet to show logs, so that regular users can share the logs with us easily if there are problems.

Preferably they can show us the entire log, but if it is too large, then it can filter `DEBUG` messages and/or truncate.  But unless there is a good reason, a large log buffer should be used so that we can get as much of the logs as possible

hanyh2004:
1,get rid of regex filter 
2,increase bufsize 
3,keep logs in memory until buf is full

PR #666, opened 2017-11-25
Implement byte array encryption function

### LINKED ISSUE #663: Implement string and byte array encryption function
Wallet encryption:

Split data into 256 bits(32 bytes) blocks (pad to 32 bytes with nulls at end)

Each block is encrypted by XORing the unencrypted block with SHA256(SHA256(password), SHA256(index, SHA256(nonce))

index is 256 bit number, and is 0 for the first block of 32 bytes, 1 for the second block of 32 bytes, 2 for third block

Combine those encrypted blocks in hex with comma, SHA256 nonce with it as checksum for error detection.

Use base64 encoded  <checksum(32 bytes)><nonce(32 bytes)> as the finally encrypted string

iketheadore:
Fix #663

### DIFF
--- a/src/cipher/encrypt.go
+++ b/src/cipher/encrypt.go
@@ -1,4 +1,4 @@
-package wallet
+package cipher
 
 import (
 	"bytes"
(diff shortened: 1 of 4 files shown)


gz-c:
Encrypt/decrypt methods and review moved to #698

Issue #660, opened 2017-11-25
Measure daemon run loop call times


gz-c:
Log a warning if more than 200ms passes for any operation called by the daemon.Run select {}

Ch3ck:
@gz-c I'll love to take on this issue.

gz-c:
This was one finished by #677

Issue #659, opened 2017-11-25
Measure networking loop cycle time


gz-c:
Measure time between calls to update/dispatch of networking messages
Log a warning if more than 500 ms passes between updates

gz-c:
https://github.com/skycoin/skycoin/blob/develop/src/daemon/gnet/pool.go#L444-L465

PR #677, opened 2017-11-27
Elapse too long operations:

Vavilen:
- readLoop
- sendLoop
- receiveMessage
- daemon Run.select{} loop

Fix [#659](https://github.com/skycoin/skycoin/issues/659), [#660](https://github.com/skycoin/skycoin/issues/660)

### DIFF
--- a/src/daemon/daemon.go
+++ b/src/daemon/daemon.go
@@ -15,9 +15,9 @@ import (
 	"github.com/skycoin/skycoin/src/daemon/gnet"
 	"github.com/skycoin/skycoin/src/daemon/pex"
 
+	"github.com/skycoin/skycoin/src/util/elapse"
 	"github.com/skycoin/skycoin/src/util/logging"
 	"github.com/skycoin/skycoin/src/util/utc"
-	"github.com/skycoin/skycoin/src/util/elapse"
 )
 
 /*
@@ -56,8 +56,8 @@ var (
 
 const (
 	// MaxDropletPrecision represents the precision of droplets
-	MaxDropletPrecision = 1
-	MaxDropletDivisor   = 1e6
+	MaxDropletPrecision        = 1
+	MaxDropletDivisor          = 1e6
 	daemonRunDurationThreshold = time.Millisecond * 200
 )
 
@@ -388,9 +388,8 @@ func (dm *Daemon) Run() error {
 
 loop:
 	for {
+		elapser.CheckForDone()
 		select {
-		case <-elapser.Done:
-			elapser.Elapsed()
 		case <-dm.quitC:
 			break loop
 			// Remove connections that failed to complete the handshake
@@ -522,8 +521,6 @@ loop:
 
 	return err
 }
-
-
 
 // GetListenPort returns the ListenPort for a given address.
 // If no port is found, 0 is returned.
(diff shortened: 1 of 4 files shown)


gz-c:
Need to run `make check` and fix linter warnings

Files must be formatted with `goimports` (you can enable this by default with your editor)

src/util/elapse/elapser.go
```diff
@@ -0,0 +1,61 @@
...
+	elapser := &Elapser{
+		elapsedThreshold: elapsedThreshold,
+		Done:             make(chan bool, 100),
+		logger:           logger,
+	}
+	return elapser
+}
+
+func (e *Elapser) CheckForDone() {
+	if len(e.Done) > 0 {
```
gz-c: Use non-blocking select
```go
select {
case <-e.Done:
    e.Elapsed()
default:
}
```

src/daemon/gnet/pool.go
```diff
@@ -435,34 +444,40 @@ func (pool *ConnectionPool) readLoop(conn *Connection, msgChan chan []byte, qc c
...
 func (pool *ConnectionPool) sendLoop(conn *Connection, timeout time.Duration, qc chan struct{}) error {
+	var elapser = elapse.NewElapser(sendLoopDurationThreshold, logger)
 	for {
 		select {
 		case <-pool.quit:
 			return nil
 		case <-qc:
 			return nil
 		case m := <-conn.WriteQueue:
+			elapser.Register("conn.WriteQueue")
```
gz-c: Why not put `defer elapser.CheckForDone()` before the `for` loop like in `readLoop`?

If it needs to be in this block, wrap the block into a function and use the `defer`, instead of calling it for every return

src/daemon/daemon.go
```diff
@@ -382,18 +384,24 @@ func (dm *Daemon) Run() error {
 	}
 
 	var err error
+	var elapser = elapse.NewElapser(daemonRunDurationThreshold, logger)
+
 loop:
 	for {
 		select {
+		case <-elapser.Done:
```
gz-c: Do you need this? Can't you can `elapser.CheckForDone` at the beginning of the `for` loop?

gz-c:
You can use `make format` to format the code

src/daemon/strand/strand.go
```diff
@@ -97,7 +97,7 @@ func WithQuit(logger *logging.Logger, c chan Request, name string, f func() erro
 			if elapsed > logDurationThreshold {
 				logger.Warning("%s took %s", name, elapsed)
 			} else {
-				 //logger.Debug("%s took %s", name, elapsed)
+				//logger.Debug("%s took %s", name, elapsed)
```
gz-c: this should have double space between `//` and the text

src/daemon/daemon.go
```diff
@@ -382,18 +384,23 @@ func (dm *Daemon) Run() error {
...
+	var elapser = elapse.NewElapser(daemonRunDurationThreshold, logger)
+
 loop:
 	for {
+		elapser.CheckForDone()
 		select {
 		case <-dm.quitC:
 			break loop
-		// Remove connections that failed to complete the handshake
+			// Remove connections that failed to complete the handshake
```
gz-c: Why is the comment indentation changed? It's doesn't make sense to read now

Issue #653, opened 2017-11-24
Connection Pool: Connection Pool Size


skycoin-manager:
Increase connection pool size (double to 32 or 64)

gz-c:
This should be a configurable option if it is not

Vavilen:
It looks like MaxConnections parameter changing does not give any effect.
The connections count increase without any control even if I pass the parameter here
https://github.com/skycoin/skycoin/blob/1fec1f06083f8046e9b66f17501d44b5629a490a/src/daemon/gnet/pool.go#L278
The connections counter just increments and nothing more.
Am I right?

gz-c:
How many connections are there?

gz-c:
When you run it, `MaxConnections` is set to 16, see cmd/skycoin/skycoin.go

Vavilen:
Here [click](https://github.com/skycoin/skycoin/blob/29974b4a517efdf976efe9432caa49b2d55b432e/src/daemon/gnet/pool.go#L93) we can see 128 connections, but the parameter is never used, if I`m not mistaken.

Vavilen:
I made changeable the MaxConnections param from cmd/skycoin/skycoin.go and passed as an config field into the daemon/gnet/pool.go, but it cannot limit the connections count.

iketheadore:
The peerlist max size is managed here: https://github.com/skycoin/skycoin/blob/develop/src/daemon/pex/pex.go#L139:2

gz-c:
#709

PR #687, opened 2017-11-28
Connection Pool: Connection Pool Size

Vavilen:
Fix [#653](https://github.com/skycoin/skycoin/issues/653)

Just a notice:
In fact, connections count can be exceeded at first seconds after the application start and be remain at ~ constant level, because outgoing connections count is checked only here:
```
// Fill up our outgoing connections
		case <-outgoingConnectionsTicker:
```
but there is other case with peer connecting:
`case <-privateConnectionsTicker:`


iketheadore:
Change outgoing connection pool size won't fix the peer list full issue. We need to change the pool size of incoming connection pool size, and it's default value 1000, managed in src/daemon/pex/pex.go.

Vavilen:
@iketheadore, ok, I`ll add this param as command line param. What is the best default value for this parameter?

cmd/skycoin/skycoin.go
```diff
@@ -486,6 +490,7 @@ func configureDaemon(c *Config) daemon.Config {
 	dc := daemon.NewConfig()
 	dc.Pex.DataDirectory = c.DataDirectory
 	dc.Pex.Disabled = c.DisablePEX
+	dc.Pex.Max = c.MaxIncomingConnections
```
gz-c: Is this the right setting to apply? Or do we need a new setting just for peer list size?

gz-c: The peerlist size controls the max incoming connections?

iketheadore: I think need a new setting, the peer list doesn't control the incoming connections directly.

gz-c:
#709

Issue #674, opened 2017-11-27
CLI option to control peerlist downloading


gz-c:
Add command line option in networking tab to disable "fetch bootstrap peers from skycoin.net".

e.g. `--download-peers-list=false`

Disable this by default when running from source.

Enable this by default in the distributed GUI release. You may have to edit the electron launcher file in the `electron/` folder to do this.

Use this option to enable/disable peers bootstrap file downloading as implemented in #673

Issue #673, opened 2017-11-27
Download a peerlist from the website


gz-c:
* Add a function to download a peerlist from https://downloads.skycoin.net/blockchain/peers.txt
* On node startup, download the peerlist, parse it into a peers structure and add it to the internal peers list
* If the download fails, log the error but do not exit the application
* Make this feature optional inside the peers code using a config flag

PR #694, opened 2017-11-29
Download a peerlist from the website

Vavilen:
Fix [#673](https://github.com/skycoin/skycoin/issues/673), [#674](https://github.com/skycoin/skycoin/issues/674)


src/daemon/pex/pex.go
```diff
@@ -215,6 +229,33 @@ func New(cfg Config, defaultConns []string) (*Pex, error) {
...
+	}
+
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	for _, address := range strings.Split(string(body), "\n") {
+		if validateAddress(address, px.Config.AllowLocalhost) {
```
gz-c: Never allow localhost in this peerlist, use `false`

src/daemon/pex/pex.go
```diff
@@ -215,6 +229,33 @@ func New(cfg Config, defaultConns []string) (*Pex, error) {
...
+			err := px.AddPeer(address)
```
gz-c: Parse the peerlist into a list of addresses, then shuffle the list before adding

src/daemon/pex/pex.go
```diff
@@ -215,6 +229,33 @@ func New(cfg Config, defaultConns []string) (*Pex, error) {
 	return pex, nil
 }
 
+func (px *Pex) download() error {
+	resp, err := http.Get(PeerDatabaseUrl)
```
gz-c: Make the download and read into a separate function at package level, that accepts a url and returns a string

src/daemon/pex/pex.go
```diff
@@ -215,6 +229,33 @@ func New(cfg Config, defaultConns []string) (*Pex, error) {
...
+		return err
+	}
+
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	for _, address := range strings.Split(string(body), "\n") {
```
gz-c: Make a separate function at package level, for parsing a string peerlist into a list of peers

src/daemon/pex/pex.go
gz-c: Later, this will be moved to util and have backoff options

gz-c:
Replaced by #701

PR #698, opened 2017-12-01
Encrypt seed and private keys in wallet

### LINKED ISSUE #664: Encrypt seed and private keys in wallet files

iketheadore:
Fixes #664 

Changes:
- Implemented sha256xor way to do encryption/decryption, which use secp156k1 as key derivation, and encrypt/decrypt data with sha256xor. Check the sha256xor details in `src/cipher/encrypt/sha256xor.go`.
- Implemented scryptChacha20poly1305 way to do encryption/decryption, which use scrypt as key derivation, and encrypt/decrypt data with chacha20poly1305. Check the scryptChacha20poly1305 details in `src/cipher/encrypt/scrypt_chacha20poly1305.go`.
- Encrypt all secrets(seed, lastSeed, and pub/sec keys in all entries) together, and store the encrypted value in `wallet.meta.secrets`.

Yes.


gz-c:
What frontend changes are needed?

src/daemon/gateway.go
```diff
@@ -516,48 +517,58 @@ func (gw *Gateway) Spend(wltID string, amt wallet.Balance, dest cipher.Address)
...
-		headTime := gw.v.Blockchain.Time()
-		tx, err = wlt.CreateAndSignTransaction(sv, unspent, headTime, amt, dest)
+	gw.strand("EncryptWallet", func() {
+		wlt, err = gw.v.Wallets.Encrypt(wltName, password)
 	})
-	return tx, err
+	return wlt, err
 }
 
+// CreateSpendingTransaction creates spending transactions
```
gz-c: Delete this if unused

src/api/cli/generate_addrs.go
```diff
@@ -63,7 +64,13 @@ func generateAddrs(c *gcli.Context) error {
 		return err
 	}
 
-	addrs, err := GenerateAddressesInFile(w, num)
+	fmt.Println("enter password: ")
```
gz-c: need cli flag to disable encryption

gz-c:
Need an endpoint to say if a wallet is encrypted or not, so the client knows if it needs to ask for password

gz-c:
Need endpoint to encrypt/decrypt wallet
Need cli command to encrypt/decrypt wallet

src/wallet/wallet.go
```diff
@@ -96,40 +119,181 @@ func NewWallet(wltName string, opts Options) (*Wallet, error) {
...
+// If the wallet is not encrypted, it would return ErrWalletNotEncrypted error
+func (w *Wallet) guard(password []byte, f func(w *Wallet) error) (err error) {
+	if !w.IsEncrypted() {
+		return ErrWalletNotEncrypted
+	}
 
-	w := Wallet{
-		Meta:    r.Meta,
-		Entries: ets,
+	if password == nil {
```
gz-c: `len(password) == 0`

src/wallet/wallet.go
```diff
@@ -96,40 +119,181 @@ func NewWallet(wltName string, opts Options) (*Wallet, error) {
...
+	for i, e := range w.Entries {
+		se, err := Encrypt(e.Secret[:], password)
+		if err != nil {
+			return err
+		}
+
+		// Set the encrypted seckey value
+		w.Entries[i].EncryptedSeckey = se
+		// Clear the entry.Secret
+		w.Entries[i].Secret = cipher.SecKey{}
```
gz-c: Also do

`for i := Secret { Secret[i] = 0}`to wipe memory before overwriting with empty seckey

src/cipher/encrypt.go
```diff
@@ -0,0 +1,173 @@
...
+	"encoding/binary"
+	"errors"
+	"strings"
+)
+
+const (
+	// the data size of each block
+	blockSize = 32 // 32 bytes
+	// the data length prefix size
+	lenPrefixSize = 4 // 4 bytes
```
gz-c: uint64 uses 8 bytes?

src/cipher/encrypt.go
```diff
@@ -0,0 +1,173 @@
...
+
+// hash(password, hash(index, hash(nonce)))
+func hashPwdIndexNonce(password []byte, index int64, nonce []byte) SHA256 {
+	// convert index to 256bit number
+	indexBytes := make([]byte, 32)
+	binary.PutVarint(indexBytes, index)
+
+	// hash(index, hash(nonce))
+	nonceHash := SumSHA256(nonce)
+	indexNonceHash := SumSHA256(append(indexBytes, nonceHash[:]...))
```
gz-c: Can use AddSHA256 here too?

src/cipher/encrypt.go
```diff
@@ -0,0 +1,173 @@
...
+		h := hashPwdIndexNonce(password, int64(i), nonce)
+		bh := SHA256(blocks[i])
+		encryptedHash := bh.Xor(h)
+		// encode the encrypted hash in base64
+		s := base64.StdEncoding.EncodeToString(encryptedHash[:])
+		encryptedBlocks = append(encryptedBlocks, s)
+	}
+
+	encryptedData := strings.Join(encryptedBlocks, ",")
+	nonceAndDataBytes := append(nonce, []byte(encryptedData)...)
```
gz-c: Doesn't the nonce need to be base64 encoded too

src/cipher/encrypt.go
```diff
@@ -0,0 +1,173 @@
...
+	buf := bytes.NewBuffer(data)
+
+	checkSumBytes := make([]byte, blockSize)
+	n, err := buf.Read(checkSumBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	if n != blockSize {
+		return nil, errors.New("decode checksum failed")
```
gz-c: "invalid checksum length"

src/cipher/encrypt.go
```diff
@@ -0,0 +1,173 @@
...
+		// encode the encrypted hash in base64
+		s := base64.StdEncoding.EncodeToString(encryptedHash[:])
+		encryptedBlocks = append(encryptedBlocks, s)
+	}
+
+	encryptedData := strings.Join(encryptedBlocks, ",")
+	nonceAndDataBytes := append(nonce, []byte(encryptedData)...)
+	checkSum := SumSHA256(nonceAndDataBytes)
+	var buf bytes.Buffer
+	_, err := buf.Write(checkSum[:])
```
gz-c: base64 encode?

src/cipher/encrypt.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if n != blockSize {
+		return nil, errors.New("decode checksum failed")
+	}
+
+	var checkSum SHA256
+	copy(checkSum[:], checkSumBytes)
```
gz-c: base64 decode?

src/cipher/encrypt.go
```diff
@@ -0,0 +1,173 @@
...
+	nonce := RandByte(blockSize)
+	var encryptedBlocks []string
+	// encode the blocks
+	for i := range blocks {
+		// hash(password, hash(index, hash(nonce)))
+		h := hashPwdIndexNonce(password, int64(i), nonce)
+		bh := SHA256(blocks[i])
+		encryptedHash := bh.Xor(h)
+		// encode the encrypted hash in base64
+		s := base64.StdEncoding.EncodeToString(encryptedHash[:])
```
gz-c: I think the base64 encoding should be postponed until later, and base64 encode everything together before writing, after added nonce and checksum.

Checksum should be calculated on the raw bytes

src/cipher/encrypt.go
```diff
@@ -0,0 +1,173 @@
...
+	if n != blockSize {
+		return nil, errors.New("decode checksum failed")
+	}
+
+	var checkSum SHA256
+	copy(checkSum[:], checkSumBytes)
+
+	// verify the checksum
+	csh := SumSHA256(buf.Bytes())
+	if csh.Hex() != checkSum.Hex() {
```
gz-c: Isn't there an sha256 compare method? shouldn't need to convert to string first

src/cipher/encrypt.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// hash(password, hash(index, hash(nonce)))
+func hashPwdIndexNonce(password []byte, index int64, nonce []byte) SHA256 {
+	// convert index to 256bit number
+	indexBytes := make([]byte, 32)
+	binary.PutVarint(indexBytes, index)
+
+	// hash(index, hash(nonce))
+	nonceHash := SumSHA256(nonce)
```
gz-c: you can pass the hashed nonce as argument since the nonce does not change

src/cipher/encrypt.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	nonce := make([]byte, blockSize)
+	n, err = buf.Read(nonce)
+	if err != nil {
+		return nil, err
+	}
+
+	if n != blockSize {
+		return nil, errors.New("decode nonce failed")
```
gz-c: "invalid nonce length"

src/cipher/encrypt.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	buf = bytes.NewBuffer(decodeData)
+	l, err := binary.ReadUvarint(bytes.NewReader(decodeData[:lenPrefixSize]))
+	if err != nil {
+		return nil, ErrInvalidPassword
+	}
+
+	if l > uint64(len(decodeData[lenPrefixSize:])) || l < 32 {
+		return nil, ErrInvalidPassword
```
gz-c: is this `ErrInvalidPassword` or a different error?

src/wallet/entry.go
```diff
@@ -1,56 +1,13 @@
...
-// VerifyPublic checks that the public key is associated with the address
-func (we *Entry) VerifyPublic() error {
-	if err := we.Public.Verify(); err != nil {
-		return err
-	}
-	return we.Address.Verify(we.Public)
+	Address         cipher.Address
+	Public          cipher.PubKey
+	Secret          cipher.SecKey
+	EncryptedSeckey string
```
gz-c: EncryptedSecret

src/wallet/readable.go
```diff
@@ -16,11 +18,18 @@ type ReadableEntry struct {
...
+		secret = w.Secret.Hex()
+	} else {
+		secret = w.EncryptedSeckey
+	}
+
 	return ReadableEntry{
 		Address: w.Address.String(),
 		Public:  w.Public.Hex(),
-		Secret:  w.Secret.Hex(),
+		Secret:  secret,
```
gz-c: Why not add `EncryptedSecret` to `ReadableEntry`?
The way here, is not easy to check if the value is encrypted or not

gz-c: The `Entry` should know if it is encrypted or not. The caller does not need to specify.  If `EncryptedSecret` is not empty then it is encrypted (and `Secret` should be empty).

gz-c: If it makes sense to leave `isEncrypted` argument, then check that `w.EncryptedSecret` is not empty

src/wallet/readable.go
```diff
@@ -118,19 +153,32 @@ func LoadReadableWallet(filename string) (*ReadableWallet, error) {
...
+	return w, nil
 }
 
-// Save saves to filename
+// Save saves to filename, remove .bak file if exist
 func (rw *ReadableWallet) Save(filename string) error {
-	// logger.Info("Saving readable wallet to %s with filename %s", filename,
-	// 	self.Meta["filename"])
-	return file.SaveJSON(filename, rw, 0600)
+	if err := file.SaveJSON(filename, rw, 0600); err != nil {
```
gz-c: Make new method that does not save .bak file, instead of creating one then deleting it

src/wallet/wallet.go
```diff
@@ -427,6 +586,35 @@ func (w *Wallet) CreateAndSignTransaction(vld Validator, unspent blockdb.Unspent
 	return &txn, nil
 }
 
+func (w *Wallet) setEncrypted(encrypt bool) {
+	if encrypt {
+		w.Meta["encrypted"] = "true"
```
gz-c: why is this saved as string? use bool

src/cipher/encrypt.go
iketheadore: We actually only need to base64 encode the encrypted block hashes,  the reason we do base64 encoding is we need to compose the encrypted hashes with comma. There's no need to base64 encode the nonce, checksum, etc.

src/cipher/encrypt.go
iketheadore: Only encrypted block hashes need to be base64 encoded.

src/cipher/encrypt.go
iketheadore: If the decrypted prefix length is not matched, we can say the password is invalid,  return "invalid prefix length" doesn't make sense to the user I think.

src/cipher/encrypt.go
iketheadore: The index here is not SHA256 hashed.

src/wallet/readable.go
iketheadore: There's a `encrypted` field in wallet meta info, the entries in the wallet would be either all encrypted or not encrypted. The `secret` field in `ReadableEntry` is string type, so we can decode it into `secret` or `encryptedSecret` in `Entry` struct base on the `encrypted` meta info in wallet,

gz-c: I am not sure we should have two different formats for the same field and toggle them based on a 3rd value

gz-c: Leave it as is for now, maybe entry can also have an encrypted bool field

src/wallet/wallet.go
iketheadore: The `Meta` is `map[string][string]`, maybe should use `map[string]interface{}`

gz-c: Oh, i thought it was `map[string]interface{}`,
leave this then

iketheadore: I changed it to `map[string]interface{}`, so that the meta can have data type.

src/wallet/wallet.go
```diff
@@ -96,40 +119,227 @@ func NewWallet(wltName string, opts Options) (*Wallet, error) {
...
+// unlock decrypts the wallet into a temporary decrypted copy of the wallet
+// It returns an error if decryption fails
+// The temporary decrypted wallet should be erased from memory when done.
+func (w *Wallet) unlock(password []byte) (*Wallet, error) {
+	if !w.IsEncrypted() {
+		return nil, ErrWalletNotEncrypted
+	}
+
+	if len(password) == 0 {
+		return nil, errors.New("password is required to decrypt wallet")
```
gz-c: `ErrRequirePassword`

src/wallet/wallet.go
```diff
@@ -96,40 +119,227 @@ func NewWallet(wltName string, opts Options) (*Wallet, error) {
...
+		return fmt.Errorf("unlock wallet failed: %v", err)
+	}
+
+	defer func() {
+		if lockErr := wlt.lock(password); lockErr != nil {
+			err = fmt.Errorf("lock wallet failed: %v", lockErr)
+			return
+		}
+
+		*w = *wlt
```
gz-c: Change this to do:

if `f(wlt)` succeeds, then lock(wlt), copy to w.  lock(wlt) should return a copy of wlt, if not, then clone wlt before copy to w.

Add an erase method to wipe secret fields.  In the defer, call erase() on wlt

iketheadore: We use the `seed` field to store either raw seed or encrypted seed. Once the `unlock` method is called, the encrypted seed value will be replaced with decrypted seed value, so if `f(wlt)` failed, calling `erase()` could only wipe the secrets, can't set back the original encrypted seed value, so does the `lastSeed`.

To implement this, I will add `encryptedSeed` and `encryptedLastSeed` fields, call `unlock` method to decrypt them and set the decrypted value in `seed` and `lastSeed` fields, call `erase` method to wipe the `seed` and `lastSeed` value.

src/cipher/encrypt.go
gz-c: Messages for the user should be transformed at the API layer, if necessary.  We do not care about the user in this method.

src/cipher/encrypt.go
```diff
@@ -0,0 +1,202 @@
...
+
+var (
+	// ErrInvalidPassword represents the invalid password error
+	ErrInvalidPassword = errors.New("invalid password")
+)
+
+// Encrypt encrypts the data with password
+//
+// 1> Add 32 bits length prefix to indicate the length of data. <length(4 bytes)><data>
+// 2> Pad the length + data to 32 bytes will nulls at end
```
gz-c: "with nulls"

src/cipher/encrypt.go
```diff
@@ -0,0 +1,202 @@
...
+// 2> SHA256(<length(4 bytes)><data><padding>) and prefix the hash. <hash(32 bytes)><length(4 bytes)><data><padding>
+// 3> Split the whole data(hash+length+data+padding) into 256 bits(32 bytes) blocks
+// 4> Each block is encrypted by XORing the unencrypted block with SHA256(SHA256(password), SHA256(index, SHA256(nonce))
+// 	  - index is 0 for the first block of 32 bytes, 1 for the second block of 32 bytes, 2 for third block
+// 5> Prefix nonce and SHA256 the nonce with blocks to get checksum, and prefix the checksum
+// 6> Finally, the data format is: <checksum(32 bytes)><nonce(32 bytes)><block0.Hex(), block1.Hex()...>
+func Encrypt(data []byte, password []byte) ([]byte, error) {
```
gz-c: return error if len(password) == 0

src/cipher/encrypt.go
```diff
@@ -0,0 +1,202 @@
...
+	// set data length prefix
+	dataLenBytes := make([]byte, lengthSize)
+	binary.LittleEndian.PutUint32(dataLenBytes, uint32(len(data)))
```
gz-c: Need to check that len(data) <= math.MaxUint32

src/cipher/encrypt.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if n != lengthSize {
+		return nil, errors.New("read data length failed")
+	}
+
+	l := binary.LittleEndian.Uint32(dataLenBytes)
+	if l > uint32(buf.Len()) {
```
gz-c: Need to check that buf.Len() <= math.MaxUint32

src/cipher/encrypt.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	// Reads out the raw data
+	rawData := make([]byte, l)
+	n, err = buf.Read(rawData)
+	if err != nil {
+		return nil, err
+	}
+
+	if uint32(n) != l {
```
gz-c: check n for math.MaxUint32

src/cipher/encrypt.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	return rawData, nil
+}
+
+// hash(password, hash(index, hash(nonce)))
+func hashPwdIndexNonce(password SHA256, index int64, nonceHash SHA256) SHA256 {
+	// convert index to 256bit number
+	indexBytes := make([]byte, 32)
+	binary.PutVarint(indexBytes, index)
```
gz-c: Should index also be hashed first?

src/cipher/encrypt.go
```diff
@@ -0,0 +1,202 @@
...
+
+	if uint32(n) != l {
+		return nil, fmt.Errorf("read data failed, expect %d bytes, but get %d bytes", l, n)
+	}
+
+	return rawData, nil
+}
+
+// hash(password, hash(index, hash(nonce)))
+func hashPwdIndexNonce(password SHA256, index int64, nonceHash SHA256) SHA256 {
```
gz-c: rename password to passwordHash

src/cipher/encrypt_test.go
```diff
@@ -0,0 +1,71 @@
...
+			"pwd13",
+			nil,
+		},
+		{
+			"encrypt data size > 2*32, with key",
+			"9d4a8e0b9fe800e40790cab00710ef2a49dbdb6be2b0e0950c739fe0799d0",
+			"8JMkCPphe73NQvGhmab",
+			nil,
+		},
+	}
```
gz-c: Need more test cases, generate a test for each data length between 32 and 64 to test all padding cases

src/wallet/entry.go
```diff
@@ -1,56 +1,13 @@
...
-	return &Entry{
-		Address: a,
-		Public:  cipher.PubKeyFromSecKey(s),
-		Secret:  s,
-	}, nil
-}
-
-// Verify checks that the public key is derivable from the secret key,
-// and that the public key is associated with the address
-func (we *Entry) Verify() error {
```
gz-c: Can we keep verification code for unencrypted wallets?

src/wallet/readable.go
```diff
@@ -118,19 +156,37 @@ func LoadReadableWallet(filename string) (*ReadableWallet, error) {
...
+	if err := w.validate(); err != nil {
+		return nil, fmt.Errorf("invalid wallet %s: %v", w.Filename(), err)
 	}
-	return *w, nil
+
+	return w, nil
 }
 
-// Save saves to filename
+// Save saves to filename, remove .bak file if exist
```
gz-c: Update comment

src/wallet/service.go
```diff
@@ -84,89 +86,270 @@ func (serv *Service) CreateWallet(wltName string, options Options) (Wallet, erro
...
+		w := serv.wallets[i].clone()
+		if !w.IsEncrypted() {
+			continue
+		}
+
+		// Checks if the wallet can be decrypted with the provided password,
+		// fails if can't.
+		decryptedWlt, err := w.unlock(password)
+		if err != nil {
+			return fmt.Errorf("decrypt wallet %s failed: %v", w.Filename(), err)
```
gz-c: log error with this message, but return the original error (same in `Encrypt`), so that the API handlers can check the error and return the correct status code

src/wallet/wallet.go
```diff
@@ -45,34 +56,61 @@ const (
...
+//      seed
+//      encryptedSeed - encrypted seed
+//      lastSeed - seed for generating next address
+// .    encryptedLastSeed - encrypted last seed
+//      tm - timestamp when creating the wallet
+//      type - wallet type
+//      coin - coin type
 type Wallet struct {
-	Meta    map[string]string
+	Meta    map[string]interface{}
```
gz-c: can keep it at `string`, that is fine

src/wallet/wallet.go
```diff
@@ -87,137 +125,416 @@ func NewWallet(wltName string, opts Options) (*Wallet, error) {
...
+			metaSeed:     seed,
+			metaLastSeed: seed,
+			metaTm:       fmt.Sprintf("%v", time.Now().Unix()),
+			metaType:     "deterministic",
+			metaCoin:     string(coin),
 		},
 	}
 
+	// Generate addresses
+	if _, err := w.GenerateAddresses(opts.AddressNum); err != nil {
```
gz-c: I removed this functionality already, it is in service, why is it back here

iketheadore: Without this, if we want to create an encrypted wallet with addresses, we need to first create an encrypted wallet, then unlock the wallet, generate addresses, then lock wallet. And now we just need to set options.AddressNum, and the NewWallet method will generate addresses, then locking the wallet.

src/wallet/entry.go
iketheadore: If wallet is encrypted, require password to unlock the wallet and do verification.

src/wallet/wallet.go
iketheadore: Seems this is not necessary, I will remove this.

iketheadore:
@gz-c updated, please have a check

iketheadore:
Moved the `cipher.Encrypt()` and `cipher.Decrypt()` methods into subfolder `sha256xor`, and defined the `Crypto` interface in wallet package. Also defined a `cryptoTable` in wallet, which records all supported crypto methods. Wallet can choose crypto from the table base on it's meta info `cryptoType`. 

If want to add new crypto method, just implement the interface of `Crypto` and register it in the `cryptoTable`.

@gz-c check again please.

iketheadore:
Moved the `cipher.Encrypt()` and `cipher.Decrypt()` methods into subfolder `sha256xor`, and defined the `Crypto` interface in wallet package. Also defined a `cryptoTable` in wallet, which records all supported crypto methods. Wallet can choose crypto from the table base on it's meta info `cryptoType`. 

If want to add new crypto method, just implement the interface of `Crypto` and register it in the `cryptoTable`.

@gz-c check again please.

iketheadore:
@gz-c Add scrypt chacha20poly1305, check please.

src/wallet/crypto.go
```diff
@@ -0,0 +1,405 @@
...
+	if !ok {
+		return nil, fmt.Errorf("can not find crypto %v in crypto table", cryptoType)
+	}
+
+	return c, nil
+}
+
+// scryptParams records the scrypt paramenters
+type scryptParams struct {
+	N       int    `json:"N"`       // scrypt N parament
```
gz-c: `parament` -> `parameter`

src/wallet/crypto.go
```diff
@@ -0,0 +1,405 @@
...
+		return ErrMissingPassword
+	}
+
+	if w.IsEncrypted() {
+		return ErrWalletEncrypted
+	}
+
+	wlt := w.clone()
+
+	// Derives password key
```
gz-c: panic if salt len is <8

src/wallet/crypto.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	aead, err := chacha20poly1305.New(dk)
+	if err != nil {
+		return err
+	}
+
+	nonce := cipher.RandByte(chacha20poly1305.NonceSize)
+	// Generates additional data
+	ad := cipher.RandByte(additionalDataLen)
```
gz-c: what is random additional data for? the scrypt params need to go in here, random data is useless to put there. the point is that the additional data is also authenticated (but not encrypted)

src/wallet/crypto.go
```diff
@@ -0,0 +1,405 @@
...
+	if err != nil {
+		return err
+	}
+
+	aead, err := chacha20poly1305.New(dk)
+	if err != nil {
+		return err
+	}
+
+	nonce := cipher.RandByte(chacha20poly1305.NonceSize)
```
gz-c: `aead.NonceSize()`

src/wallet/crypto.go
```diff
@@ -0,0 +1,405 @@
...
+		se := aead.Seal(nil, nonce, e.Secret[:], ad)
+
+		// Set the encrypted seckey value
+		wlt.Entries[i].EncryptedSecret = base64.StdEncoding.EncodeToString(se)
+	}
+
+	// Sets wallet as encrypted
+	wlt.setEncrypted(true)
+	// Sets authenticated data
+	wlt.setAuthenticated(authenticated{
```
gz-c: no, the parameters go into the additional data

src/cipher/scryptChacha20poly1305/scryptChacha20poly1305.go
```diff
@@ -0,0 +1,144 @@
+package scryptChacha20poly1305
```
gz-c: package name, folder name, file name, does not follow naming standards

src/wallet/service.go
```diff
@@ -352,7 +352,12 @@ func (serv *Service) CreateAndSignTransaction(wltID string, password []byte, vld
 	}
 
 	if w.IsEncrypted() {
-		if err := w.guard(password, f); err != nil {
+		dw, err := w.unlock(password)
```
gz-c: why does guard need to be removed? guard doesn't need to relock anything

iketheadore: The guard will lock the wallet at the end. I need to call defer `dw.erase()` here.

gz-c: guard:

unlock wallet, creates new wallet copy with unencrypted data

src/cipher/encrypt/scrypt_chacha20poly1305.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Scyrpt derives key from password
+	salt := cipher.RandByte(saltSize)
+	dk, err := scrypt.Key(password, salt, s.n, s.r, s.p, s.keyLen)
+	if err != nil {
+		return nil, err
+	}
+
+	// Prepare metadata
+	m := meta{s.n, s.r, s.p, s.keyLen, salt, cipher.RandByte(chacha20poly1305.NonceSize)}
```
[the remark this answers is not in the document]
gz-c: nevermind, " []byte encodes as a base64-encoded string", from https://golang.org/pkg/encoding/json/, this is fine

iketheadore:
@gz-c check again please

src/wallet/wallet.go
```diff
@@ -45,179 +67,561 @@ const (
...
 	if walletType != "deterministic" {
 		return errors.New("wallet type invalid")
 	}
 
-	if _, ok := w.Meta["coin"]; !ok {
+	if _, ok := w.Meta[metaCoin]; !ok {
 		return errors.New("coin field not set")
 	}
 
+	if _, ok := w.Meta[metaCryptoType]; !ok {
```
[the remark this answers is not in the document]
iketheadore: The meta.cryptoType will be set as `DefaultCryptoType` when loading from v1 wallet.

gz-c:
builds failing

iketheadore:
Updated, @gz-c check please

src/cipher/encrypt/sha256xor.go
```diff
@@ -0,0 +1,246 @@
...
+	"io"
+	"math"
+
+	"github.com/skycoin/skycoin/src/cipher"
+	secp256k1 "github.com/skycoin/skycoin/src/cipher/secp256k1-go"
+)
+
+const (
+	// Data size of each block
+	blockSize = 32 // 32 bytes
```
gz-c: prefix all of these with `sha256Xor` so they don't get mixed up with others

src/cipher/encrypt/scrypt_chacha20poly1305_test.go
```diff
@@ -0,0 +1,88 @@
...
+			[]byte("pwd"),
+			[]byte("wrong password"),
+			errors.New("chacha20poly1305: message authentication failed"),
+		},
+		{
+			"missing password",
+			[]byte("plaintext"),
+			[]byte("pwd"),
+			nil,
+			errors.New("missing password"),
```
gz-c: tests should include the encrypted data string

iketheadore:
@gz-c check please

iketheadore:
Rebased and merged the updates from develop

gz-c:
merge conflict

gz-c:
merge conflict

src/wallet/wallets.go
```diff
@@ -106,98 +74,68 @@ func backupWltFile(src, dst string) error {
...
-func (wlts Wallets) ToReadable() []*ReadableWallet {
-	return wlts.toReadable(NewReadableWallet)
+// Update updates the given wallet, return error if not exist
+func (wlts Wallets) update(id string, updateFunc func(*Wallet) (*Wallet, error)) error {
+	w, ok := wlts[id]
+	if !ok {
+		return ErrWalletNotExist
+	}
+
+	newWlt, err := updateFunc(w.clone())
```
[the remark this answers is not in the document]
iketheadore: We update the clone wallet, in case of interrupting the original wallet in the updateFunc. Actually the returned `newWlt` is the cloned wallet, and will be used to replace the old wallet, therefore, call erase of the cloned wallet would make the wallet lose seeds, address keys, etc. 

I will change the signature of `updateFunc`, it shouldn't have *Wallet returned, which might confuse user, and don't know if the input and returned wallet are related.

PR #699, opened 2017-12-01
Require coinhours

gz-c:
*Note: in the future we may keep transactions that are invalid due to insufficient coinhours in the unconfirmed pool. In this case, a user could submit a transaction early and wait for it to confirm later.  This requires certain UI/UX prompts, since the user may not want to wait so long.  For now, require them to have the coinhours available first.*

## Change BurnFactor to 3

1/3 of coinhours are required for the transaction now, instead of 1/2.

## Fix change hours calculation

A mistake in the change hours calculation sent 1/8 of coinhours to the change address and 1/8 to the destination addresses (or 1/4 to the destination addresses if there were no coins as change).

Fix this to send 1/4 to change and 1/4 to the destination addresses (or 1/2 to the destination addresses).

The initial division by `fee.BurnFactor` for the coinhours burn rounds up in favor of the burn.  For example, 1 coin hour is split into a 1 coin hour fee and 1 coin hour in change.  3 are split into 2 coin hour fee and 1 for change.

When splitting between change and destination addresses, the change address is favored for the extra hour, if the remaining coin hours are odd.

When splitting amongst multiple destination addresses, each one is allocated an extra coin hour in sequence from the remainder of the final division.  

For example:

* `101 coin hours input, 2 destination addresses -> fee=51, change=25, dest1=13, dest2=12`
* `100 coin hours input, 1 destination address -> fee=50, change=25, dest1=25`
* `99 coin hours input, 3 destination addresses -> fee=50, change=25, dest1=8, dest2=8, dest3=8`
* `1 coin hour input, 1 destination address -> fee=1, change=0, dest1=0`
* `23 coin hours input, no change, 2 destination address -> fee=12, change=0, dest1=6, dest2=5`

## Require at least one transaction input to have non-zero coinhours

Any transaction that does not burn any coinhours is rejected, to enforce spam restrictions.  This is important now that decimals will be extended.

## Change uxout spend selection

* First choose the uxout with the highest coins and non-zero coinhours
* Then choose uxouts with zero coinhours, from either highest coins first or lowest coins first
* Then choose uxouts with non-zero coinhours, from either highest coins first or lowest coins first

The first step is to ensure that at least one input has coin hours.
The second step is to maximize spending of uxouts that don't have coinhours, since these can't be spent unless combined with a uxout that has coinhours.
The third step uses the remaining uxouts available.

When uxouts are sorted highest-first, it minimizes the number of uxouts spent.  This is the default for cli's `create_rawtx` command.  Users of this command likely want to retain as many uxouts with coinhours as possible, in case of frequent sending, such as an exchange.  Although they should use multisend transactions to better avoid the problem of unconfirmed unspent outputs.

When uxouts are sorted lowest-first, it maximizes the number of uxouts spent.  This is the default for spends through the wallet service API, which the GUI clients use.  This is chosen as the default, to avoid creating uxouts excessively.

## Add `block_seq` to `get_outputs` and `/outputs` API response


src/wallet/wallet.go
```diff
@@ -648,5 +570,93 @@ func cmpUxOutByHash(a, b UxBalance) bool {
...
+		return nil, errors.New("zero spend amount")
+	}
+
+	if len(uxa) == 0 {
+		return nil, errors.New("no unspents to spend")
+	}
+
+	for _, ux := range uxa {
+		if ux.Coins == 0 {
+			logger.Panic("UxOut coins are 0, can't spend")
```
iketheadore: Panic here? this function has return error value

src/wallet/wallet.go
```diff
@@ -648,5 +570,93 @@ func cmpUxOutByHash(a, b UxBalance) bool {
 		logger.Panic("Duplicate UxOut when sorting")
 	}
 	return cmp < 0
+}
+
+// ChooseSpends chooses uxouts from a list of uxouts.
+// It first chooses the uxout with the most number of coins that has nonzero coinhours.
+// It then chooses uxouts with zero coinhours, ordered by sortStrategy
+// It then chooses remaining uxouts with nonzero coinhours, ordered by sortStrategy
+func ChooseSpends(uxa []UxBalance, amt Balance, sortStrategy func([]UxBalance)) ([]UxBalance, error) {
```
iketheadore: So whatever the strategy is, It always choose the uxout that has highest coins with none-hours first?

src/wallet/wallet.go
```diff
@@ -384,31 +381,93 @@ func (wlt *Wallet) CreateAndSignTransaction(
...
+		// If remainingHours is an odd number, give the extra hour to the change output
+		if remainingHours%2 == 1 {
+			changeHours++
+		}
+
+	}
+
+	// Distribute the remaining hours equally amongst the destination outputs
+	remainingAddrHours := remainingHours - changeHours
+	addrHoursShare := remainingAddrHours / nAddrs
```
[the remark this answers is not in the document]
gz-c: It shouldn't be called with 0, I figure tests should discover

src/wallet/wallet.go
gz-c: This panic was in the old function too, I think it panics because this should never happen and means something is corrupt internally

src/wallet/wallet.go
gz-c: Yes, because you need at least 1 input with coinhours.  So it looks at the one with the highest coins, since that is most likely to satisfy the send amount.

Then it tries to get the rest from inputs with no coinhours.  Keep in mind, most of the time they will have coinhours. Only someone spending frequently will have 0 coin hours, at least until we allow people to send outputs with 0 coinhours.

If it still needs more, then it gets from the remaining outputs with coinhours.

Issue #610, opened 2017-11-03
create_rawtx tests


gz-c:
The create_rawtx command in the CLI api needs tests.

Related to: https://github.com/skycoin/skycoin/pull/609

gz-c:
Replaced by #939

PR #702, opened 2017-12-03
Add test for create_rawtx #610

jayphbee:
add some test for create_rawtx.go


jayphbee:
unexpected operation, closed

Issue #695, opened 2017-11-30
TestLoadPeersFromFile non-deterministic failure


gz-c:
```
--- FAIL: TestLoadPeersFromFile (0.12s)
    --- FAIL: TestLoadPeersFromFile/one_addr (0.06s)
	Error Trace:	peerlist_test.go:129
	Error:      	Not equal: 
	            	expected: pex.Peer{Addr:"112.32.32.14:7200", LastSeen:1511876126, Private:false, Trusted:false, HasIncomingPort:false, RetryTimes:0}
	            	actual: pex.Peer{Addr:"112.32.32.14:7200", LastSeen:1511876127, Private:false, Trusted:false, HasIncomingPort:false, RetryTimes:0}
	            	
	            	Diff:
    --- FAIL: TestLoadPeersFromFile/two_addr (0.02s)
	Error Trace:	peerlist_test.go:129
	Error:      	Not equal: 
	            	expected: pex.Peer{Addr:"112.32.32.14:7200", LastSeen:1511876126, Private:false, Trusted:false, HasIncomingPort:false, RetryTimes:0}
	            	actual: pex.Peer{Addr:"112.32.32.14:7200", LastSeen:1511876127, Private:false, Trusted:false, HasIncomingPort:false, RetryTimes:0}
```

The `LastSeen` value need to check within 1 second, not exact match

Vavilen:
Please, look at PR: [#723](https://github.com/skycoin/skycoin/pull/723)

PR #723, opened 2017-12-07
The LastSeen value is checked within 1 second

Vavilen:
Fix [#695](https://github.com/skycoin/skycoin/issues/695)

### DIFF
--- a/src/daemon/pex/peerlist_test.go
+++ b/src/daemon/pex/peerlist_test.go
@@ -511,12 +511,9 @@ func TestPeerJSONParsing(t *testing.T) {
 }
 
 func peersEqualWithSeenAllowedDiff(t *testing.T, expected Peer, actual Peer) {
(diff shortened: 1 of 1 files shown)


src/daemon/pex/peerlist_test.go
```diff
@@ -510,6 +510,15 @@ func TestPeerJSONParsing(t *testing.T) {
 	check(p)
 }
 
+func peersEqualWithSeenAllowedDiff(t *testing.T, expected Peer, actual Peer) {
```
gz-c: Change this to something like

e.g.
```go
require.WithinDuration(t, time.Unix(expected.LastSeen, 0), time.Unix(actual.LastSeen, 0), 1*time.Second)
expected.LastSeen = actual.LastSeen
require.Equal(t, expected, actual)
```

So that if we add a new field to peer, we don't have to remember to update this test method to make sure the new field is compared.

Issue #631, opened 2017-11-22
Refactor gui http interface into a struct


gz-c:
The gui http interface still uses some global vars,
the package should expose a configurable struct,
and provide a proper shutdown mechanism

Vavilen:
@gz-c , could you give me more details about this issue?
Which global vars do you mean? Just for example.

gz-c:
The logger can stay as a global

The `Launch*` methods should not call `serve()`

The `Shutdown()` method should be a struct method

The caller (in `cmd/skycoin/skycoin.go`) should do something like:

```go
s := Create()
...
go s.Serve()
....
s.Shutdown()
```

And in the `go s.Serve()` goroutine, manage it with a `sync.WaitGroup` (you will see other goroutines do this, e.g. `wg.Add(1)`)

Vavilen:
@gz-c , please, look at my PR: [#727](https://github.com/skycoin/skycoin/pull/727)

gz-c:
#791

PR #727, opened 2017-12-08
Refactor gui http interface into a struct

Vavilen:
PR for [#631](https://github.com/skycoin/skycoin/issues/631)

@gz-c , what is the point of quit channel here:
https://github.com/skycoin/skycoin/blob/130913a893a075975e22dc26431dd6366868b3e3/cmd/skycoin/skycoin.go#L571
It is used only here for closing the channel:
https://github.com/skycoin/skycoin/blob/130913a893a075975e22dc26431dd6366868b3e3/cmd/skycoin/skycoin.go#L576


gz-c:
That quit channel is used elsewhere in the run method, it is for handling ctrl+c shutdown

cmd/skycoin/skycoin.go
```diff
@@ -725,6 +722,7 @@ func Run(c *Config) {
 			}()
 		}
 	*/
+	go s.Serve()
```
gz-c: Need to use the waitgroup like the other goroutines

src/gui/http.go
```diff
@@ -27,44 +28,67 @@ var (
 	quit     chan struct{}
 )
 
+type Server struct {
+	wg *sync.WaitGroup
```
gz-c: Do not pass waitgroup or pointers to waitgroup around

src/gui/http.go
```diff
@@ -74,22 +98,25 @@ func LaunchWebInterfaceHTTPS(host, staticDir string, daemon *daemon.Daemon, cert
...
 
 	// Runs http.Serve() in a goroutine
-	serve(listener, NewGUIMux(appLoc, daemon), quit)
 	return nil
+
 }
 
-func serve(listener net.Listener, mux *http.ServeMux, q chan struct{}) {
+func (s *Server) Serve() {
+	mux := NewGUIMux(s.appLoc, s.daemon)
```
gz-c: The main thread calls serve in a goroutine, serve should not also run a goroutine

cmd/skycoin/skycoin.go
```diff
@@ -434,6 +434,56 @@ func catchDebug() {
 	}
 }
 
+func createGUI(c *Config, d *daemon.Daemon, wg sync.WaitGroup) *gui.Server {
```
gz-c: sync.WaitGroup cannot be used like this


>A WaitGroup waits for a collection of goroutines to finish. The main goroutine calls Add to set the number of goroutines to wait for. Then each of the goroutines runs and calls Done when finished. At the same time, Wait can be used to block until all goroutines have finished.
>A WaitGroup must not be copied after first use.

https://golang.org/pkg/sync/#WaitGroup

You must use it from the same function it is created

cmd/skycoin/skycoin.go
```diff
@@ -725,6 +722,11 @@ func Run(c *Config) {
 			}()
 		}
 	*/
+	wg.Add(1)
```
gz-c: Start the gui goroutine in the same sequence that it previously was started in

src/gui/http.go
```diff
@@ -27,44 +27,65 @@ var (
...
 
-	listener, err = net.Listen("tcp", host)
+	s.listener, err = net.Listen("tcp", host)
 	if err != nil {
 		return err
 	}
 
 	// Runs http.Serve() in a goroutine
-	serve(listener, NewGUIMux(appLoc, daemon), quit)
+	//s.serve(NewGUIMux(appLoc, daemon, &s))
```
gz-c: delete old lines of code

src/gui/http.go
```diff
@@ -74,33 +95,32 @@ func LaunchWebInterfaceHTTPS(host, staticDir string, daemon *daemon.Daemon, cert
...
-	go func() {
-		for {
-			if err := http.Serve(listener, mux); err != nil {
-				select {
-				case <-q:
-					return
-				default:
-				}
-				continue
+func (s *Server) Serve() {
```
gz-c: return error (except when quit)
log error in the goroutine that calls `Serve()` from the main thread

src/gui/http.go
```diff
@@ -74,33 +95,32 @@ func LaunchWebInterfaceHTTPS(host, staticDir string, daemon *daemon.Daemon, cert
...
+			continue
 		}
-	}()
+	}
 }
 
 // Shutdown close http service
-func Shutdown() {
+func (s *Server) Shutdown() {
 	if quit != nil {
```
gz-c: Remove the quit != nil check

src/gui/http.go
gz-c: Add a `done` channel to the `Server` struct and defer `close(s.done)` here
In the shutdown channel, wait for `<-s.done` before exiting the `Shutdown` method

src/gui/http.go
```diff
@@ -74,33 +95,32 @@ func LaunchWebInterfaceHTTPS(host, staticDir string, daemon *daemon.Daemon, cert
...
 			}
+			continue
 		}
-	}()
+	}
 }
 
 // Shutdown close http service
-func Shutdown() {
+func (s *Server) Shutdown() {
```
gz-c: Add logging here similar to other Shutdown methods

src/gui/http.go
```diff
@@ -74,33 +95,32 @@ func LaunchWebInterfaceHTTPS(host, staticDir string, daemon *daemon.Daemon, cert
...
-		for {
-			if err := http.Serve(listener, mux); err != nil {
-				select {
-				case <-q:
-					return
-				default:
-				}
-				continue
+func (s *Server) Serve() {
+	mux := NewGUIMux(s.appLoc, s.daemon)
```
gz-c: Add logging here about http server starting and defer a "server closed" log statement, similar to other `Run` methods

src/gui/http.go
```diff
@@ -27,44 +27,65 @@ var (
...
 	return nil
 }
 
 // LaunchWebInterfaceHTTPS begins listening on https://$host, for enabling remote web access
 // Uses HTTPS
-func LaunchWebInterfaceHTTPS(host, staticDir string, daemon *daemon.Daemon, certFile, keyFile string) error {
-	quit = make(chan struct{})
+func (s *Server) launchWebInterfaceHTTPS(host, staticDir string, daemon *daemon.Daemon, certFile, keyFile string) error {
+	var err error
 	logger.Info("Starting web interface on https://%s", host)
```
gz-c: Move this log line to "Serve"
Add new log line that says "Web interface tcp listening on %s host"

gz-c: Do the same for `launchWebInterfaceHTTP`

.appveyor.yml
```diff
@@ -23,7 +23,7 @@ install:
   - set PATH=%GOPATH%\bin;c:\go\bin;%PATH%
   - go version
   - go env
-  - go get github.com/gz-c/gox
+  - go get github.com/mitchellh/gox
```
gz-c: why is this changed

.travis.yml
```diff
@@ -26,7 +26,7 @@ matrix:
 
 install:
   # Install gox
-  - go get github.com/gz-c/gox
+  - go get github.com/mitchellh/gox
```
gz-c: should not be changed

cmd/skycoin/skycoin.go
gz-c: This needs to be before the commented out code

cmd/skycoin/skycoin.go
```diff
@@ -725,6 +706,30 @@ func Run(c *Config) {
...
+	go func() {
+		defer wg.Done()
+		err = s.Serve()
```
gz-c: This needs to be `err := s.Serve()`

src/gui/http.go
```diff
@@ -22,49 +22,64 @@ import (
...
 	devDir      = "dev/"
 	indexPage   = "index.html"
 )
 
+func Create(needHttps bool, host string, guiDirectory string, daemon *daemon.Daemon, cert string, key string) *Server {
+	s := Server{
+		daemon: daemon,
+		done:   make(chan struct{}),
+	}
+	if needHttps {
```
gz-c: Return error??

src/gui/http.go
```diff
@@ -74,39 +89,40 @@ func LaunchWebInterfaceHTTPS(host, staticDir string, daemon *daemon.Daemon, cert
...
-				case <-q:
-					return
-				default:
-				}
-				continue
+func (s *Server) Serve() error {
+	defer logger.Info("Server closed")
+	logger.Info("Starting web interface on %s", s.listener.Addr())
+	mux := NewGUIMux(s.appLoc, s.daemon)
+	for {
```
gz-c: What is this for loop for

src/gui/http.go
```diff
@@ -74,39 +89,40 @@ func LaunchWebInterfaceHTTPS(host, staticDir string, daemon *daemon.Daemon, cert
...
-	if quit != nil {
-		// must close quit first
-		close(quit)
-		listener.Close()
-		listener = nil
-	}
+func (s *Server) Shutdown() {
+	logger.Info("Shutting down Server")
+	// must close done first
+	close(s.done)
```
gz-c: You misunderstood, the `Serve()` loop closes a `done` channel. At the end of `Shutdown`, wait for it, i.e. `<-s.done`

gz-c:
#791

Issue #636, opened 2017-11-23
Test /wallet/spend


gz-c:

Vavilen:
@gz-c , I think the test covers this case: https://github.com/skycoin/skycoin/blob/41be95959901370b6fd7766b2c7c5a978c5476c3/src/wallet/service_test.go#L573

gz-c:
@Vavilen that covers the real spending logic but doesn't test the handler itself, to check the query arguments processing.

Vavilen:
@gz-c so, is it a test for HTTP handler not the CreateAndSignTransaction function exactly as and others /wallet/* tests?

gz-c:
Yes, only the http handlers, the wallet package will test its own methods.

We may need to use an interface for `Gateway` to do this testing easier

Vavilen:
@gz-c , please, assign this issue to me.

PR #743, opened 2017-12-13
Test /wallet/spend

Vavilen:
Fix [[#636]](https://github.com/skycoin/skycoin/issues/636)


src/api/cli/check_balance_test.go
```diff
@@ -1,266 +0,0 @@
-package cli
```
gz-c: why is this test removed

src/gui/wallet_test.go
```diff
@@ -0,0 +1,330 @@
...
+		{
+			"200 - OK",
+			"POST",
+			"/wallet/spend",
+			&httpBody{
+				Id:    "1234",
+				Dst:   "2konv5no3DZvSMxf2GPVtAfZinfwqCGhfVQ",
+				Coins: "12",
+			},
+			http.StatusOK,
```
gz-c: should not be `StatusOK`

src/gui/wallet_test.go
```diff
@@ -0,0 +1,330 @@
...
+		{
+			"200 - gw GetWalletBalance error",
+			"POST",
+			"/wallet/spend",
+			&httpBody{
+				Id:    "1234",
+				Dst:   "2konv5no3DZvSMxf2GPVtAfZinfwqCGhfVQ",
+				Coins: "12",
+			},
+			http.StatusOK,
```
gz-c: should not be `StatusOK`

src/gui/wallet_test.go
```diff
@@ -0,0 +1,330 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		rr := httptest.NewRecorder()
+		handler := http.HandlerFunc(WalletSpendHandler(gateway))
+
+		handler.ServeHTTP(rr, req)
+
+		status := rr.Code
```
gz-c: use `require.Equal`

src/gui/wallet_test.go
```diff
@@ -0,0 +1,330 @@
...
+		if status != tc.status {
+			t.Errorf("case: %s, handler returned wrong status code: got `%v` want `%v`",
+				tc.name, status, tc.status)
+		}
+		if status != http.StatusOK {
```
gz-c: use `require.Equal`

src/gui/wallet_test.go
Vavilen: Then we need make some changes [here](https://github.com/skycoin/skycoin/blob/6634757f5dbf49be580e8ad7a0862d25bcd2dedd/src/gui/wallet.go#L141), because here we receive SpendResult struct with Error field. But the status is 200 OK.

gz-c:
```
--- FAIL: TestWalletSpendHandler (0.00s)
	Error Trace:	wallet_test.go:318
	Error:      	Not equal: 
	            	expected: "400 Bad Request - invalid destination address: Invalid base58 character"
	            	actual: "400 Bad Request - invalid destination address: Invalid address length"
	Messages:   	case: 400 - bad dst addr, handler returned wrong error message: got `400 Bad Request - invalid destination address: Invalid base58 character`| %!s(int=400), want `400 Bad Request - invalid destination address: Invalid address length`
```

gz-c:
#793

Issue #637, opened 2017-11-23
Test /wallet/balance


gz-c:

Vavilen:
@gz-c , please, assign this issue to me.

Vavilen:
@gz-c , please, look at my [PR](https://github.com/skycoin/skycoin/pull/747)

PR #747, opened 2017-12-14
Test /wallet/balance

Vavilen:
Fix [#637](https://github.com/skycoin/skycoin/issues/637)
One of the tests fails, because [here](https://github.com/skycoin/skycoin/blob/2b1a1908cb1a173b85f16a28acdc3c019658868d/src/gui/wallet.go#L89) just logs the error and returns 200 - OK.


gz-c:
Can you return a 400 error where the error is missing.

Actually the error should be a 404 or 500 depending on the error; if the wallet is missing then its a 404 otherwise a 500.  It returns an error created by `errWalletNotExist()` if the wallet is missing, but we can't check this without inspecting the string value of the error, which we should do.

Vavilen:
@gz-c You mean, that I need to change [here](https://github.com/skycoin/skycoin/blob/2b1a1908cb1a173b85f16a28acdc3c019658868d/src/gui/wallet.go#L89) or  wallet_test.go?

src/gui/wallet_test.go
```diff
@@ -432,3 +433,132 @@ func TestWalletSpendHandler(t *testing.T) {
...
+					tc.name, status, tc.status)
+			}
+			if status != http.StatusOK {
+				if errMsg := rr.Body.String(); strings.TrimSpace(errMsg) != tc.err {
+					t.Errorf("case: %s, handler returned wrong error message: got `%v`| %s, want `%v`",
+						tc.name, errMsg, status, tc.err)
+				}
+			} else {
+				var msg wallet.BalancePair
+				err = json.Unmarshal(rr.Body.Bytes(), &msg)
```
gz-c: require.NoError

src/gui/wallet_test.go
```diff
@@ -432,3 +433,132 @@ func TestWalletSpendHandler(t *testing.T) {
...
+			handler.ServeHTTP(rr, req)
+
+			status := rr.Code
+			require.Equal(t, tc.status, status, "case: %s, handler returned wrong status code: got `%v` want `%v`", tc.name, status, tc.status)
+			if status != tc.status {
+				t.Errorf("case: %s, handler returned wrong status code: got `%v` want `%v`",
+					tc.name, status, tc.status)
+			}
+			if status != http.StatusOK {
+				if errMsg := rr.Body.String(); strings.TrimSpace(errMsg) != tc.err {
```
gz-c: require.Equal

src/gui/wallet.go
```diff
@@ -48,6 +48,7 @@ func walletBalanceHandler(gateway Gatewayer) http.HandlerFunc {
 		b, err := gateway.GetWalletBalance(wltID)
 		if err != nil {
 			logger.Error("Get wallet balance failed: %v", err)
+			wh.Error400(w, "get wallet balance failed")
```
gz-c: switch err
case wallet.ErrWalletNotExist: Error404
default: Error500

Issue #640, opened 2017-11-23
Test /wallet/newAddress


gz-c:

Vavilen:
PR [#757](https://github.com/skycoin/skycoin/pull/757)

PR #757, opened 2017-12-18
Test /wallet/newAddress

Vavilen:
For [#640](https://github.com/skycoin/skycoin/issues/640)


src/gui/wallet_test.go
```diff
@@ -15,6 +13,9 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"bytes"
```
gz-c: sort the imports properly

src/gui/wallet_test.go
```diff
@@ -548,10 +555,191 @@ func TestWalletGet(t *testing.T) {
...
+				Id:  "foo",
+				Num: "1",
+			},
+			http.StatusBadRequest,
+			"400 Bad Request - gateway.NewAddresses error",
+			"foo",
+			1,
+			make([]cipher.Address, 0),
+			errors.New("gateway.NewAddresses error"),
+			Addresses{},
```
gz-c: duplicate test case?

src/gui/wallet_test.go
```diff
@@ -548,10 +555,191 @@ func TestWalletGet(t *testing.T) {
...
+				Id:  "foo",
+				Num: "1",
+			},
+			http.StatusOK,
+			"",
+			"foo",
+			1,
+			emptyAddrs,
+			nil,
+			responseEmptyAddresses,
```
gz-c: what is this case testing?

src/gui/wallet_test.go
Vavilen: Not quite.

src/gui/wallet_test.go
Vavilen: This case shows that the result of gateway.NewAddresses can be an empty []cipher.Address, but the response status is 200.

src/gui/wallet_test.go
Vavilen: fixed

gz-c:
merge conflict

src/gui/wallet_test.go
gz-c: ok, only when num is 0

src/gui/wallet_test.go
gz-c: it looks the same as the one about it

src/gui/wallet_test.go
gz-c: it should only return empty addrs if num is 0

src/gui/wallet_test.go
gz-c: *above it

Vavilen: You're right. I will remove it.

src/gui/wallet_test.go
Vavilen: Changed 1 -> 0

PR #771, opened 2017-12-21
Update /wallet apis to support encrypted wallets

### LINKED ISSUE #665: Update wallet API to support encrypted wallets
`GET /wallet`
    - Don't return sensitive data
    - Fix the response data format to be the same as POST /wallet/create response

`POST /wallet/create`
    - Add encrypt (bool) and password arguments
    - Don't return sensitive data

`POST /wallet/newAddress`
    - Add password argument

`POST /wallet/spend`
    - Add password argument

New endpoint:

```
GET /wallet/seed
Args:
    id - wallet ID
    password - wallet password
```

This endpoint will only be enabled if the node is configured to run with it. Add a CLI flag, `-enable-seed-api` and use it when calling the electron app.

If the wallet is not encrypted, return 403 Unauthorized
    - User can't find seed of unencrypted wallet through the API with this
    - In the UI, call /wallet/folderName and tell the user where to find their wallet file, and suggest they encrypt the wallet

Returns:

```json
{
    "filename": "...",
    "label": "...",
    "seed": "..."
}
```

New endpoint:

```
POST /wallet/encrypt
Args:
    id - wallet ID
    password - wallet password
```

If the wallet is already encrypted, return 400 Bad Request

Returns the same response as `GET /wallet`

New endpoint:

```
POST /wallet/decrypt
Args:
    id - wallet ID
    password - wallet password
```

If the wallet is not encrypted, return 400 Bad Request

Returns the same response as `GET /wallet`

iketheadore:
1. Add "encrypt" and "password" paraments in /wallet/create api. "encrypt" value must be 0 or 1, and "password" must be provided if "encrypt" is 1.
2. Add "password" paramenter in the following apis:
        /wallet/newAddress
        /wallet/spend
3. Reorder the api handler functions, the same as they are registered in RegisterWalletHandlers function.
4. Update api comments

Fix #665

### DIFF
--- a/src/api/cli/add_private_key.go
+++ b/src/api/cli/add_private_key.go
@@ -86,7 +86,7 @@ func AddPrivateKey(wlt *wallet.Wallet, key string) error {
 	return wlt.AddEntry(entry)
 }
 
-// Adds a private key to a wallet based on filename.  Will save the wallet after modifying.
+// AddPrivateKeyToFile adds a private key to a wallet based on filename.  Will save the wallet after modifying.
 func AddPrivateKeyToFile(walletFile, key string) error {
 	wlt, err := wallet.Load(walletFile)
 	if err != nil {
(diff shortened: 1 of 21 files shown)


gz-c:
merge conflicts

samuelvisscher:
@iketheadore please let me know when this is ready for front-end development

Issue #765, opened 2017-12-19
Add vet linter from gometalinter


Vavilen:

Vavilen:
#718

Vavilen:
go vet ./...
src/api/cli/wallet_history.go:26: struct field coins has json tag but is not exported
exit status 1
src/cipher/chacha20/go-chacha20/chacha/chacha20_ref.go:1: possible malformed +build comment
exit status 1
src/cipher/secp256k1-go/secp256k1-go2/sig.go:206: arg len(r) for printf verb %s of wrong type: int
exit status 1
src/consensus/blockstat.go:295: unreachable code
src/consensus/blockstat.go:576: unreachable code
exit status 1
src/visor/readable_test.go:1: +build comment must appear before package clause and be followed by a blank line
exit status 1
make: *** [lint] Error 1

PR #772, opened 2017-12-21
fixed go vet errors

Vavilen:
For [#765](https://github.com/skycoin/skycoin/issues/765)


src/cipher/chacha20/go-chacha20/chacha/chacha20_ref.go
```diff
@@ -1,5 +1,3 @@
-// This was originally "+build !cgo" but the cgo file is always ignored in build now
```
gz-c: Leave this comment
Figure out a different way to suppress the error

Can try adding `//` as the first line of the file

Makefile
```diff
@@ -35,6 +35,7 @@ test: ## Run tests
 lint: ## Run linters. Use make install-linters first.
 	vendorcheck ./...
 	gometalinter --disable-all -E goimports --tests --vendor ./...
```
gz-c: Enable `vet` in this line, don't add a new line

src/consensus/blockstat.go
```diff
@@ -290,11 +289,6 @@ func (self *BlockStat) try_add_hash_and_sig(
 	self.debug_count += 1
 	self.debug_usage += 1
 
-	if !(action_update || action_skip || action_insert) {
```
gz-c: Comment out this code instead (normally you should delete it, but this code in the `consensus/` folder is reference code and I do not want to remove what was there)

src/consensus/blockstat.go
```diff
@@ -571,11 +565,6 @@ func (self *BlockStatQueue) try_append_to_BlockStatQueue(
 	}
 	n = -1 // guard
 
-	if !(action_update || action_skip || action_insert) {
```
gz-c: Comment out this code instead (normally you should delete it, but this code in the `consensus/` folder is reference code and I do not want to remove what was there)

Issue #749, opened 2017-12-15
Rewrite gui/wallet.go:Spend function


gz-c:
`Spend()` in gui/wallet.go should return an error.

Also remove the for loop that it uses, it is unnecessary.

If a caller of `Spend` needs to write the `SpendResult` as json to the user, it should add the error to the json result itself.

PR #774, opened 2017-12-22
refactor spend function and add error to result

therealssj:
Refactor gui/wallet.go:Spend function 

return `balance`, `transaction`, `error` instead of `SpendResult`
also added some documentation

Resolves: #749

### DIFF
--- a/src/gui/wallet.go
+++ b/src/gui/wallet.go
@@ -23,48 +23,55 @@ type SpendResult struct {
 	Error       string                     `json:"error,omitempty"`
 }
 
-// Spend spends coins from specific wallet
-func Spend(gateway *daemon.Gateway, walletID string, coins uint64, dest cipher.Address) *SpendResult {
+// Spend spends coins from given wallet id
+// Args:
+//  walletID    string          ID of wallet to spend from
+//  coins       uint64          amount of coins to spend
+//  dest        ciper.Address   recipient address
+// Return:
+//  balance     *wallet.BalancePair         latest balance
+//  transaction *visor.ReadableTransaction  readable transaction
+//  error       error                       error in spending the coins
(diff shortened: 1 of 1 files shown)


src/gui/wallet.go
```diff
@@ -23,46 +23,44 @@ type SpendResult struct {
 	Error       string                     `json:"error,omitempty"`
 }
 
-// Spend spend coins from specific wallet
+// Spend spends coins from specific wallet
 func Spend(gateway *daemon.Gateway, walletID string, coins uint64, dest cipher.Address) *SpendResult {
```
gz-c: Return `(balance, transaction, error)` from this method
The caller of this should pack it into the `SpendResult` json struct

src/gui/wallet.go
```diff
@@ -23,50 +23,55 @@ type SpendResult struct {
...
-		}
 
-		var txStr string
-		txStr, err = visor.TransactionToJSON(*tx)
-		if err != nil {
-			break
-		}
-
-		logger.Info("Spend: \ntx= \n %s \n", txStr)
+	if tx, err = gateway.Spend(walletID, coins, dest); err != nil {
```
gz-c: write it as 

```go
tx, err := gateway.Spend(...)
if err != nil { 
    ...
    return nil, nil, err 
}
```

Remove the `var` declarations at the top

therealssj: `return nil, nil, err` ?
Should I not return the wallet balance and transaction when it fails?

gz-c: No, you should return empty values for the initial arguments, when there is an error

gz-c: No, you should return empty values for the initial arguments, when there is an error

Issue #762, opened 2017-12-19
Replace base58 with a better library


gz-c:
See #760 for the original discussion/problem

Our base58 library had a problem with base58 parsing, it treated any leading invalid base58 character as a `0` character (only `"1"` is a valid leading `0` character).

In case it has other problems, it should be replaced with a more readable and maintained/tested library.

This may be a suitable replacement library:
https://github.com/itchyny/base58-go

Before replacing the base58 library, add many valid addresses to the test cases. Compare the two libraries: verify that for valid addresses, the decoded result is the same, and that they encode to the same string values.

therealssj:
Have you taken a look at [this](https://github.com/mr-tron/base58) library
Its fast base58 implementation is significantly faster than of the library mentioned above.

gz-c:
Yes but the code is more complicated so its harder to audit, I'm not sure we should use it

therealssj:
Fair point. 
https://github.com/itchyny/base58-go also provides a neat cli just to quickly test some base58 addresses.

Out of context but you guys have some contributing guidelines?

gz-c:
Regarding contribution guidelines, travis will enforce most of the rules.  Descriptive commit messages or pull requests are helpful

therealssj:
I played with both libraries a bit more and I take back my opinion of going with https://github.com/itchyny/base58-go
The way that library implements the base58 decode is fine but the end result is not the hex value but the ascii value of hex values which makes it more complicated to work as the current codebase is around hex the values.

PR #775, opened 2017-12-23
[WIP]Replace old base58 library

### LINKED ISSUE #763: Test base58 address encoder/decoder thoroughly
We need more thorough test cases for base58 address encoding and decoding.

See #760, #761 and #762 for more background.

We need many hardcoded address test cases, and also fuzzing or property testing, to test many addresses intelligently.

therealssj:
Replace old base58 library with https://github.com/m0t0k1ch1/base58
Resolves: #762

TODO:
- [ ] Add Tests to resolve #763

Opening this halfway just to get more of an opinion on the change


gz-c:
What I don't like about this library is that it allocates the charmap every time

gz-c:
We could fork it and copy the source into our tree and modify it though, the rest of the code seems simple enough

src/cipher/address.go
```diff
@@ -153,12 +167,20 @@ func (addr Address) Verify(key PubKey) error {
 // version is first byte in binary format
 // in printed address its key, version, checksum
 func (addr Address) String() string {
-	return string(base58.Hex2Base58(addr.Bytes()))
+	str, err := b58.EncodeToString(addr.Bytes())
+	if err != nil {
```
gz-c: The library returns `err` if `bytes.Buffer.WriteBytes` returns an error. However, this method never returns an error. See: https://golang.org/pkg/bytes/#Buffer.WriteByte
So `EncodeToString` will never return an error.
Furthermore, that library does not check the `WriteBytes` error for one of its calls, which is inconsistent (although it never needs to check the error).

I'm starting to lean towards having our own base58 implementation based upon one of these libraries, but fixing the minor issues.

Still, in a case like this (where our method does not return an error), it should call `panic(err)` (although `err` will always be `nil`).

therealssj: I could try to fix the issues with the existing library but I am also more inclined towards creating a new library at this point

therealssj:
Instead of having `DecodeBase58Address` and `MustDecodeBase58Address` 
Why don't we just have `DecodeBase58Address(addr string, strict bool)`?

gz-c:
@therealssj bools do not belong in APIs and `Must` methods are idiomatic for go, where the function will panic if there is an error

samuelvisscher:
Has this PR been abandoned?

gz-c:
No, we'll get back to it later

src/cipher/address_test.go
```diff
@@ -1,24 +1,77 @@
...
+// Generate random string of length n
+func RandStringBytes(n int) []byte {
+	b := make([]byte, n)
+	for i := range b {
+		b[i] = base58.BTCALPAHBET[rand.Intn(len(base58.BTCALPAHBET))]
+	}
+
+	return b
+}
+
```
gz-c: We should pull all addresses from blockchain and save them to a file, and use that as test data for a test of the skycoin address parser

src/cipher/address.go
```diff
@@ -192,7 +212,12 @@ func BitcoinAddressFromPubkey(pubkey PubKey) string {
 	b3 := append([]byte{byte(0)}, b2[:]...)
 	b4 := DoubleSHA256(b3)
 	b5 := append(b3, b4[0:4]...)
-	return string(base58.Hex2Base58(b5))
+
+	str, err := base58.Encode(b5)
+	if err != nil {
+		logger.Error("Error in encoding address to base58: %s", err)
```
gz-c: panic

src/cipher/address.go
```diff
@@ -153,12 +165,20 @@ func (addr Address) Verify(key PubKey) error {
...
+	}
+	return str
 }
 
 // BitcoinString convert bitcoin address to hex string
 func (addr Address) BitcoinString() string {
-	return string(base58.Hex2Base58(addr.BitcoinBytes()))
+	str, err := base58.Encode(addr.BitcoinBytes())
+	if err != nil {
+		logger.Error("Error in encoding address to base58: %s", err)
```
gz-c: panic

src/cipher/address.go
```diff
@@ -153,12 +165,20 @@ func (addr Address) Verify(key PubKey) error {
 // version is first byte in binary format
 // in printed address its key, version, checksum
 func (addr Address) String() string {
-	return string(base58.Hex2Base58(addr.Bytes()))
+	str, err := base58.Encode(addr.Bytes())
+	if err != nil {
+		logger.Error("Error in encoding address to base58: %s", err)
```
gz-c: panic

gz-c:
Is this still WIP?

therealssj:
not really. I kinda need to go over the tests again I guess

PR #792, opened 2017-12-28
Add /transactions endpoint

iketheadore:
Add `/transactions` endpoint, which can be used to query multiple addresses related transactions.
example:
1. Get all addresses related transactions, return all transactions if $addrs is not provided:
```
/transactions?addrs=$addrs
```
2. Get addresses related confirmed transactions:
```
/transactions?addrs=$addrs&confirmed=1
```
3. Get addresses related unconfirmed transactions:
```
/transactions?addrs=$addrs&confirmed=0
```

**TODO:**

- [x] Write test code

### DIFF
--- a/src/gui/transaction.go
+++ b/src/gui/transaction.go
@@ -28,7 +28,7 @@ func RegisterTxHandlers(mux *http.ServeMux, gateway *daemon.Gateway) {
 	// Returns transactions that match the filters.
 	// Method: GET
 	// Args:
-	//     addrs: Comma seperated addresses [optional, returns all transactions if no address provided]
+	//     addrs: Comma seperated addresses [optional, returns all transactions if no address is provided]
 	//     confirmed: Whether the transactions should be confirmed [optional, must be 0 or 1; if not provided, returns all]
 	mux.HandleFunc("/transactions", getTransactions(gateway))
 	//inject a transaction into network
--- a/src/gui/transaction.go
+++ b/src/gui/transaction.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/skycoin/skycoin/src/cipher"
 	"github.com/skycoin/skycoin/src/coin"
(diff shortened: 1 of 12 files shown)


gz-c:
What if you want confirmed and unconfirmed

src/gui/transaction.go
```diff
@@ -130,6 +138,97 @@ func getTransactionByID(gate *daemon.Gateway) http.HandlerFunc {
...
+			return nil, err
+		}
+
+		addrs = append(addrs, a)
+	}
+
+	return addrs, nil
+}
+
+// Parses the confirmed string int bool value, must be "0" or "1".
```
gz-c: Make a `parseBool` method

`"0", "f", "F", "false", "FALSE", ""` are false
`"1", "t", "T", "true", "TRUE"` are true

will use this for other APIs

src/visor/historydb/historydb.go
```diff
@@ -6,6 +6,7 @@ import (
 	"errors"
 
 	"github.com/boltdb/bolt"
+	"github.com/rencatoken/rnc/src/cipher/encoder"
```
gz-c: bad import

iketheadore: Will fix this

src/gui/transaction.go
iketheadore: ok

iketheadore:
The 'confirmed' parameter is optional, if you don't set it, this api would return confirmed and unconfirmed transactions to you.

src/gui/transaction.go
gz-c: nevermind, can rewrite this to use `strconv.ParseBool`

gz-c: https://golang.org/pkg/strconv/#ParseBool

iketheadore:
@gz-c updated, please check again

src/visor/blockchain_test.go
```diff
@@ -43,15 +43,18 @@ func makeFeeCalc(fee uint64) coin.FeeCalculator {
...
 	require.NoError(t, err)
 	gbSig := cipher.SignHash(gb.HashHeader(), genSecret)
 
+	bcc, ok := bc.(*Blockchain)
+	require.True(t, ok)
+
 	// add genesis block to blockchain
-	bc.db.Update(func(tx *bolt.Tx) error {
-		return bc.store.AddBlockWithTx(tx, &coin.SignedBlock{
+	bcc.db.Update(func(tx *bolt.Tx) error {
```
gz-c: should check err result

src/visor/unconfirmed.go
```diff
@@ -275,7 +275,7 @@ func (utp *UnconfirmedTxnPool) createUnconfirmedTxn(t coin.Transaction) Unconfir
 // InjectTxn adds a coin.Transaction to the pool, or updates an existing one's timestamps
 // Returns an error if txn is invalid, and whether the transaction already
 // existed in the pool.
-func (utp *UnconfirmedTxnPool) InjectTxn(bc *Blockchain, t coin.Transaction) (bool, error) {
+func (utp *UnconfirmedTxnPool) InjectTxn(bc blockchainer, t coin.Transaction) (bool, error) {
```
gz-c: blockchainer should be Blockchainer since the rest is public

src/visor/visor.go
```diff
@@ -159,17 +161,73 @@ func (c Config) Verify() error {
 	return nil
 }
 
+// historyer is the interface that provides methods for accessing history data that are parsed from blockchain.
+type historyer interface {
```
gz-c: make this public too, Historyer

iketheadore: The `historyer` interface is only used in the visor package, so don't have to make it public,. And go prefer expose struct instead of interface, expose interface only if the package have public methods or functions that have the interface as parameter, otherwise should keep it in the package.

PR #820, opened 2018-01-04
Add click to copy to wallet gui

therealssj:
When you click on an address it will copied to clipboard

Here is a gif to show the change http://g.recordit.co/d3Dv105Ylt.gif


gz-c:
Don't recompile dist, we decided only core team will recompile it

PR #852, opened 2018-01-15
Update peers.txt

asxtree:


gz-c:
This peer was missing the port, can you submit it again with the port number

PR #864, opened 2018-01-18
advanced spend builder

therealssj:
# Motivation
Allow the user to tweak his `coin hours`  and `unspents` in a transaction

# Info
- By default in a transaction, a user loses `75%` coin hours wherein `50%` is burned and `25%` is sent to destination addrs.
- The default strategy to select unspents priority is `Max skycoins` > `zero coin hours` > `non-zero coin hours` 

# Tasks
- [x] choose coin hours to be sent to destination addrs
- [x]  choose unspents to be used
- [x]  send from multiple addresses
- [x]  create `advancedCreateRawTx` command
- [x] Update changelog
- [x] Document use cases
- [ ] Tests 

# Use Cases
- This will allow users to send the desired number of coinhours to the receiver.
- Select unspent with lower coin hours to decrease transaction fees and hence preserving coinhours

### DIFF
--- a/cmd/cli/README.md
+++ b/cmd/cli/README.md
@@ -143,16 +143,29 @@ Use `skycoin-cli send -h` to see the subcommand usage.
 ### Advanced send
 
 ```bash
-$ skycoin-cli advancedSend -s $coinhours_distribute -u $unspent_hash $recipient_address $amount
+$ skycoin-cli advancedSend -$a $from_address $recipient_address $amount $hours
 ```
 
-The above `advancedSend` command will send coins from your node's default wallet: `$HOME/.skycoin/wallets/skycoin_cli.wlt` if it has `$unspent_hash`. 
-It willt send `$coinhours_distribute` to `$recipient_address`
+The above `advancedSend` command will send `$amount` coins and `$hours` coinhours from your node's default wallet: `$HOME/.skycoin/wallets/skycoin_cli.wlt`
+
 You can also send from the wallet as you want, just use the `-f` option flag, example:
-
 ```bash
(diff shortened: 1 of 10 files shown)


therealssj:
Could use some suggestions on the whole "different strategies to choose unspents" scenario

My main motivation behind this was the fact that we can decrease coin hour loss by using an unspent with fewer `coin hours` which has enough skycoins to satisfy the transaction but is not the one with the highest number of skycoins

cmd/cli/README.md
```diff
@@ -135,6 +140,22 @@ $ skycoin-cli send -f $WALLET_PATH $recipient_address $amount
...
 Use `skycoin-cli send -h` to see the subcommand usage.
 
+### Advanced send
+
+```bash
+$ skycoin-cli advancedSend -s $coinhours_distribute -u $unspent_hash $recipient_address $amount
+```
+
+The above `advancedSend` command will send coins from your node's default wallet: `$HOME/.skycoin/wallets/skycoin_cli.wlt` if it has `$unspent_hash`. 
+It willt send `$coinhours_distribute` to `$recipient_address`
```
gz-c: `willt` typo

gz-c: The `$coinhours_distribute` parameter is not clear

gz-c:
See #872, I wrote up an initial spec

therealssj:
The example seems to be unclear but the command does let you use multiple unspents

gz-c:
@therealssj should we close this?

gz-c:
We'll redo this to use the `POST /wallet/transaction` API

Issue #865, opened 2018-01-18
API to return total coin hours


gz-c:
This might fit into an existing API call, such as the endpoint that returns coin supply information for the explorer

Issue #867, opened 2018-01-18
Add begin and end pagination parameters to /explorer/address


gz-c:
See: https://github.com/skycoin/skycoin-explorer/issues/89

Senyoret1:
It will also be necessary to create a new api endpoint to get the number of transactions of an address. If this endpoint is not created, it will not be possible to use /explorer/address to correctly create a multi-page ui.

Additionally, to avoid negatively affecting any system currently using /explorer/address, it would be advisable to make the two new parameters optional.

therealssj:
I have tried to solve pagination in  #869 
though to maintain backward compatibility I don't return the total length in the response

if you give it parameters more than largest index then it sets the value to the largest index

Senyoret1:
As you say, adding the total number of transactions to the response could break the compatibility, that's why it would be good to create a totally new endpoint to check the number of transactions of an address.

The objective would be to add pagination to the list of transactions in the explorer. Currently, the explorer has pagination in the blocks list and looks like this:

![blocks](https://user-images.githubusercontent.com/34079003/35179258-07c22a08-fd6d-11e7-83e0-2d7642abee98.png)

If there is no way to get the total number of transactions, there is no way to create a control like the one in the block list. Additionally, the data is necessary because the browser shows the user how many transactions an address has.

What I do not know is if creating a new endpoint just to check the number of transactions of an address would have a high cost in CPU processing time. In that case it would be necessary to consider what options there are, such as creating a new endpoint, apart from /explorer/address, to obtain the transactions of an address with paging and returning the total number of transactions, in order to be able to make one request instead of two.

gz-c:
Adding new fields to the response shouldn't break compatibility.  Changing field names or field value types would

therealssj:
I dont think I can return the total number of transactions without breaking compatibility
```
[
    {
        "status": {
           //stuff
        },
        "length": 220,
        "type": 0,
        "txid": "86cdee14f1b9cc06710815f51e5a546a8a33c4179433e047ed50d17b3a7a734e",
        "inner_hash": "45ade9ec2b7618f782a869796f021486dda3856bf009dc6ee633d1840fd08a75",
        "timestamp": 1516000192,
        "sigs": [
            // stuff
        ],
        "inputs": [
            {
              //stuff
            }
        ],
        "outputs": [
            {
              //stuff
            },
            {
               //stuff
            }
        ]
    }
]
```
This is how the current response looks

[...]

Issue #866, opened 2018-01-18
Include coins and hours in /explorer/address inputs


gz-c:
See: https://github.com/skycoin/skycoin-explorer/issues/90

PR #869, opened 2018-01-18
Include total coinhours in /coinSupply

therealssj:
Resolves: #865

### DIFF
--- a/src/gui/explorer.go
+++ b/src/gui/explorer.go
@@ -42,6 +42,8 @@ type CoinSupply struct {
 	TotalSupply string `json:"total_supply"`
 	// MaxSupply is the maximum number of coins to be distributed ever
 	MaxSupply string `json:"max_supply"`
+	// CoinHourSupply is the total number of coinhours in the distribution addresses
+	CoinHourSupply uint64 `json:"coinhour_supply"`
 	// Distribution addresses which count towards total supply
 	UnlockedAddresses []string `json:"unlocked_distribution_addresses"`
 	// Distribution addresses which are locked and do not count towards total supply
@@ -123,10 +125,18 @@ func coinSupply(gateway *daemon.Gateway, w http.ResponseWriter, r *http.Request)
 		return nil, nil
 	}
 
+	totalCoinHours, err := gateway.GetTotalCoinhours()
+	if err != nil {
+		logger.Errorf("Failed to get total coinhours: %v", err.Error())
+		wh.Error500(w)
+		return nil, nil
+	}
+
 	cs := CoinSupply{
 		CurrentSupply:     currentSupplyStr,
 		TotalSupply:       totalSupplyStr,
 		MaxSupply:         maxSupplyStr,
+		CoinHourSupply:    totalCoinHours,
 		UnlockedAddresses: unlockedAddrs,
 		LockedAddresses:   visor.GetLockedDistributionAddresses(),
 	}
--- a/src/gui/explorer.go
+++ b/src/gui/explorer.go
@@ -16,22 +16,11 @@ func RegisterExplorerHandlers(mux *http.ServeMux, gateway *daemon.Gateway) {
 	// get set of pending transactions
 	mux.HandleFunc("/explorer/address", getTransactionsForAddress(gateway))
 
-	mux.HandleFunc("/explorer/getEffectiveOutputs", getEffectiveOutputs(gateway))
-
 	mux.HandleFunc("/coinSupply", getCoinSupply(gateway))
 
 	mux.HandleFunc("/richlist", getRichlist(gateway))
 
 	mux.HandleFunc("/addresscount", getAddressCount(gateway))
-}
-
-// DeprecatedCoinSupply records the coin supply info
-type DeprecatedCoinSupply struct {
-	CoinSupply
-	DeprecatedCurrentSupply                           uint64   `json:"coinSupply"`
-	DeprecatedCoinCap                                 uint64   `json:"coinCap"`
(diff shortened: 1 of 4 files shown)


src/visor/readable.go
```diff
@@ -127,6 +127,8 @@ type ReadableTransactionOutput struct {
 type ReadableTransactionInput struct {
 	Hash    string `json:"uxid"`
 	Address string `json:"owner"`
+	Coins   string `json:"coins"`
+	Hours   uint64 `json:"hours"`
```
gz-c: hours should always be strings

src/gui/explorer.go
```diff
@@ -123,10 +125,18 @@ func coinSupply(gateway *daemon.Gateway, w http.ResponseWriter, r *http.Request)
 		return nil, nil
 	}
 
+	totalCoinHours, err := gateway.GetTotalCoinhours()
```
gz-c: Should call GetUnspentOutputs() directly and reuse the result to calculate total coin hours and also filter the unlocked addresses which is done in the lines above

src/visor/readable.go
therealssj: aren't hours like elapsed time?
they are stored as `uint64` everywhere else

src/gui/explorer.go
therealssj: when u say filter, do u mean skip the unlocked addresses?

src/visor/readable.go
gz-c: API policy going forward is to using strings for hours

Senyoret1:
Please, see https://github.com/skycoin/skycoin/issues/867#issuecomment-359140328

src/gui/explorer.go
```diff
@@ -155,21 +152,66 @@ func getTransactionsForAddress(gateway *daemon.Gateway) http.HandlerFunc {
 			return
 		}
 
+		// Get start and end pagination index
+		start := r.FormValue("start")
```
gz-c: would be better written as 

```go
startIndex := 0
start := r.FormValue("start")
if start != "" {
    // atoi, check err, check startIndex >= 0
}
```

src/gui/explorer.go
```diff
@@ -155,21 +152,66 @@ func getTransactionsForAddress(gateway *daemon.Gateway) http.HandlerFunc {
...
+		}
+
 		cipherAddr, err := cipher.DecodeBase58Address(addr)
 		if err != nil {
 			wh.Error400(w, "invalid address")
 			return
 		}
 
 		txns, err := gateway.GetAddressTxns(cipherAddr)
+		totalLen := len(txns.Txns)
```
gz-c: This reduces the amount of data sent in the request but the real bottleneck is the db query, is it possible to improve query performance with the range parameters?

src/visor/readable.go
```diff
@@ -145,10 +147,12 @@ func NewReadableTransactionOutput(t *coin.TransactionOutput, txid cipher.SHA256)
 }
 
 // NewReadableTransactionInput creates readable transaction input
-func NewReadableTransactionInput(uxID string, ownerAddress string) ReadableTransactionInput {
+func NewReadableTransactionInput(uxID string, ownerAddress string, coins string, hours string) ReadableTransactionInput {
```
gz-c: This function signature can be rewritten as `(uxID, ownerAddress, coins, hours string)`

gz-c:
The pagination changes should be a separate PR from showing the total coin hours (which would have been merged if not for the pagination changes which may take a while)

therealssj:
@gz-c I have removed the pagination changes

therealssj:
does this need to go into the changelog?
otherwise looks good for the merge

src/visor/readable.go
```diff
@@ -127,6 +127,8 @@ type ReadableTransactionOutput struct {
 type ReadableTransactionInput struct {
 	Hash    string `json:"uxid"`
 	Address string `json:"owner"`
+	Coins   string `json:"coins"`
```
gz-c: this is not related to coinhour_supply changes

gz-c:
Remove the changes for #866 too, one PR per issue

therealssj:
removed the other changes and updated changelog and readme

src/gui/explorer.go
```diff
@@ -42,6 +31,8 @@ type CoinSupply struct {
 	TotalSupply string `json:"total_supply"`
 	// MaxSupply is the maximum number of coins to be distributed ever
 	MaxSupply string `json:"max_supply"`
+	// CoinHourSupply is the total number of coinhours in the distribution addresses
+	CoinHourSupply string `json:"coinhour_supply"`
```
gz-c: Can we do `current_coinhour_supply` (exclude all distribution addresses) and `total_coinhour_supply` (same as `coinhour_supply` is here) to match `current_supply` and `total_supply`?

src/gui/explorer.go
```diff
@@ -104,31 +106,43 @@ func coinSupply(gateway *daemon.Gateway, w http.ResponseWriter, r *http.Request)
...
+
+	// Get coin hours from confirmed unspents only
 	var totalCoinHours uint64
 	for _, out := range lockedOutputs.HeadOutputs {
 		totalCoinHours += out.Hours
 	}
 
+	// get current coin hours which excludes all distribution addresses
+	coinHourFilter = append(coinHourFilter, daemon.FbyAddressesNotIncluded(unlockedAddrs))
+	normalOutputs, err := gateway.GetUnspentOutputs(coinHourFilter...)
```
gz-c: Is it not possible to apply the filters ourselves? It's querying the full database 3 times in this endpoint now

gz-c:
need to do `make format`, one of the files is not formatted

PR #877, opened 2018-01-21
Add -disable-wallet-api option to skycoin cmd

scboffspring:
Related to #648 

Add "disable-wallet-api" option to the skycoin cmd, which will disable all wallet API from the gateway. 
Return 403 when a call to any API requiring the wallet is done from the GUI.

----
_Donation is not required, but highly appreciated (in case you find PR useful of course)._

_Skycoin Wallet: VfF9T7dWkE5TRHjwi84cnbg189vVRtbYQL_

### DIFF
--- a/src/gui/wallet.go
+++ b/src/gui/wallet.go
@@ -404,7 +404,12 @@ func walletsHandler(gateway *daemon.Gateway) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		wlts, err := gateway.GetWallets()
 		if err != nil {
-			wh.Error403(w)
+			switch err {
+			case wallet.ErrWalletApiDisabled:
+				wh.Error403(w)
+			default:
+				wh.Error500(w)
+			}
 			return
 		}
 		wh.SendOr404(w, wlts.ToReadable())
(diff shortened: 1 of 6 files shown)


src/gui/wallet.go
```diff
@@ -375,8 +402,12 @@ func walletTransactionsHandler(gateway Gatewayer) http.HandlerFunc {
 // Returns all loaded wallets
 func walletsHandler(gateway *daemon.Gateway) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		wlts := gateway.GetWallets().ToReadable()
-		wh.SendOr404(w, wlts)
+		wlts, err := gateway.GetWallets()
+		if err != nil {
```
gz-c: should do the switch like the others

scboffspring: Done

scboffspring:
@gz-c Thanks for merging. Just noticed I haven't added anything to the changelog. Should we add it? It's an "important" feature for any node administrator

gz-c:
@scboffspring yes, please add it. we're not consistent about updating the changelog yet so we update it at release time, but would be best to have it updated with each PR

PR #878, opened 2018-01-21
Include coins and hours in /explorer/address inputs

therealssj:
Resolves: #866

### DIFF
--- a/src/visor/readable.go
+++ b/src/visor/readable.go
@@ -127,6 +127,8 @@ type ReadableTransactionOutput struct {
 type ReadableTransactionInput struct {
 	Hash    string `json:"uxid"`
 	Address string `json:"owner"`
+	Coins   string `json:"coins"`
+	Hours   uint64 `json:"hours"`
 }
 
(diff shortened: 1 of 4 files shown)


therealssj:
fixed conflict

gz-c:
Do you remember the issue/PR in explorer that needed this info?

src/visor/readable.go
```diff
@@ -145,10 +147,12 @@ func NewReadableTransactionOutput(t *coin.TransactionOutput, txid cipher.SHA256)
 }
 
 // NewReadableTransactionInput creates readable transaction input
-func NewReadableTransactionInput(uxID string, ownerAddress string) ReadableTransactionInput {
+func NewReadableTransactionInput(uxID, ownerAddress, coins, hours string) ReadableTransactionInput {
```
gz-c: Probably better to have `coins, hours uint64` and do the string conversions in here

therealssj: But then will have to do error handling for `coins` inside the function

gz-c: that's fine

PR #895, opened 2018-01-23
Test gui/transactions.go endpoints

Vavilen:
- /pendingTxs
- /transaction
- /injectTransaction
- /resendUnconfirmedTxns
- /rawtx
- /transactions


gz-c:
I merged #891 

Update your tests to match the new style, and add invalid Host 403 check to each test.

Vavilen:
@gz-c, done.

src/gui/transaction_test.go
```diff
@@ -0,0 +1,794 @@
...
+	invalidTxn.Txn.Out = append(invalidTxn.Txn.Out, coin.TransactionOutput{
+		Coins: math.MaxInt64 + 1,
+	})
+
+	tt := []struct {
+		name                          string
+		method                        string
+		url                           string
+		status                        int
+		error                         string
```
gz-c: use `err` not `error`, otherwise syntax highlighting is wrong because `error` has special meaning

PR #901, opened 2018-01-26
fixes #828 skycoin docker testnet

### LINKED ISSUE #828: Testnet deploy scripts
#Implement script to deploy Docker containers running Skycoin (mainnet / testnet) . The configuration would be similar to this 

- [X] a ~(Python / Bash)~ Golang script to generate a (parameterized, e.g. number of nodes) docker-compose.yml
- [X] the docker-compose.yml to setup the testnet network consisting in :
- [X] a virtual network
- [X] <N> nodes running with no browser
- [X] 1 container running browser 
- [x] detect testnet script path Cc @olemis
- [ ] 1 container running a centralized logging server ( ~[logstash](https://www.elastic.co/guide/en/logstash/current/docker.html)~ vs ~[gollum]( https://github.com/trivago/gollum)~ vs [logspout](https://github.com/gliderlabs/logspout)+[oklog](https://github.com/oklog/oklog) 👍  ) Cc @stdevMauricio
- [x] merge with testnet branch Cc @olemis see comments in skycoin/services#12
- [x] load default connections (peers) from txt file Cc @olemis
- [ ] separate logging output (docker-compose stdout + [logspout](https://github.com/gliderlabs/logspout) ) Cc @stdevMauricio
- [x] container to run https://github.com/skycoin/skycoin-explorer Cc @stdevMauricio see skycoin/skycoin-explorer#110
- [ ] option to load existing db Cc @stdevMauricio
- [x] `dependency not vendored: gopkg.in/yaml.v2` Need to update dependencies following these [instructions](https://github.com/skycoin/skycoin#dependency-management) Cc @stdevMauricio 
- [x] Do `gofmt -w docker/images/testnet/testnet.go` as part of `make format` Cc @olemis
- ~deployment options (e.g. local, AWS, Azure, ...)~
- [ ] Replace all relative paths in `testnet.go` script (... and beyond ...) by using `projectPath` variable pointing at a working copy of skycoin/skycoin repository checked out in local file system Cc @stdevMauricio
- [x] Fix [testnet issue with pubkey](https://github.com/skycoin/skycoin/issues/828#issuecomment-364476592) Cc @olemis
- [ ] Enforce usage of different address `Version` for mainnet vs testnet ( Cc @olemis )
- [ ] Options to log events in [logspout format](https://github.com/gliderlabs/logspout#raw-format) (.Cc @olemis ) 
- [ ] Further info when `-version` flag (see #1390) is specified.
  * [ ] Adress prefix (see #988 )
  * [ ] Address version (mainnet vs testnet vs simtestnet)
- [ ] initialize genesis block for testnet
- [ ] create new blockchain from scratch using genesis creation tool (see #1468 )

stdevMauricio:
- Generates the Dockerfiles and the docker-compose file
- Creates a test network, the volumes for the containers and runs the testnet.
There are some things missing
- A webserver to serve the peers list peers.txt
- A log server to save the logs
- Some polishment.
Right now the testnet should be run offline, with a webserver running on the host serving the peers.txt. This is because the daemon downloads the peers list, but it adds those peers to the hardcoded list it has in the source code.

### DIFF
--- a/docker/images/testnet/templates/Dockerfile
+++ b/docker/images/testnet/templates/Dockerfile
@@ -44,4 +44,4 @@ VOLUME $DATA_DIR
 
 EXPOSE 6000 6420 6430
 
-CMD ["/usr/bin/skycoin", "{{ .SkyCoinParameters }}"]
+CMD ["/usr/bin/skycoin"{{ range .SkyCoinParameters }}, "{{ . }}"{{ end }}]
--- a/docker/images/testnet/templates/Dockerfile
+++ b/docker/images/testnet/templates/Dockerfile
@@ -1,0 +1,47 @@
+# skycoin build
+# reference https://github.com/skycoin/skycoin
+FROM golang:1.9 AS build
+
+# dirs
+RUN mkdir -p $GOPATH/src/github.com/skycoin/skycoin
+
+# copy code to build
+ADD src $GOPATH/src/github.com/skycoin/skycoin/src
+ADD cmd $GOPATH/src/github.com/skycoin/skycoin/cmd
+ADD vendor $GOPATH/src/github.com/skycoin/skycoin/vendor
+
+# install cli
+RUN cd $GOPATH/src/github.com/skycoin/skycoin/cmd/cli && \
+  CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o /go/bin/skycoin-cli .
+
+# install skycoin
+RUN cd $GOPATH/src/github.com/skycoin/skycoin/cmd/skycoin && \
+  CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o /go/bin/skycoin .
+
+# install address_gen
+RUN cd $GOPATH/src/github.com/skycoin/skycoin/cmd/address_gen && \
+  CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o /go/bin/address_gen .
+
+
+# skycoin image
+FROM alpine:3.7
+
+ENV COIN="skycoin" \
+    RPC_ADDR="0.0.0.0:6430" \
+    DATA_DIR="/root/.$COIN" \
+    WALLET_DIR="/wallet" \
+    WALLET_NAME="$COIN_cli.wlt"
+
+# copy all the binaries
+COPY --from=build /go/bin/* /usr/bin/
+
+# copy assets
+COPY --from=build /go/src/github.com/skycoin/skycoin/src/gui/static /usr/local/skycoin/static
+
+# volumes
+VOLUME $WALLET_DIR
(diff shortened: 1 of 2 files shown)


gz-c:
`docker/images/testnet/testnet.go:1::warning: file is not goimported`

Need to do `gofmt -w docker/images/testnet/testnet.go`

`make format` is supposed to take care of this, but it may need updating to format this new `docker/` folder's code

gz-c:
Also, 

`dependency not vendored: gopkg.in/yaml.v2`

Need to update dependencies following these instructions:

https://github.com/skycoin/skycoin#dependency-management

gz-c:
There is a flag for `-localhost-only` connections, if that works for `172.16.200.*` (depends on the `IsLocalhost` code).  If that doesn't work, we could add an ipmask filter to restrict the range of IPs that can connect to the node.  We want to avoid connecting outside of the network if possible, even on accident.

Could also have an option to specify the peerlist from disk, and another to ignore the default peers, if that would help.

iketheadore:
Node will connect to default nodes(mainnet nodes) when start, need option to stop connecting the default nodes.

olemis:
Yes @iketheadore this is something I am working on right now . Added in the TODO list in #828 . I am not sure of whether we should merge this code with develop (e.g. this option should **only** be in testnet binary, never in mainnet). @iketheadore @gz-c what do you think ?

olemis:
@iketheadore indeed the right TODO bullet point is `load default connections (peers) from txt file` because that's exactly what is needed , replace hard coded default connections list with IPs stored in a txt file .

olemis:
@gz-c loading default connection `IP:address` is already in TODO list (like I said before) but instead of disabling default peers I'd rather remove default connections completely and implement peers initialization **only** from file only for testnet . Makes sense ?

I'd recommend to implement network isolation outside containers themselves . @stdevMauricio in generated `docker-compose.yml` ?

olemis:
@stdevMauricio I also see the travis-ci build failing . Beware of that . Not including it in TODO list since this is obvious .

gz-c:
@olemis default peers from file (separate from our `peers.txt`) would be ok. If not from a file, then a comma separated list of ip:port.  Then it can be hardcoded as the default config in cmd/skycoin/skycoin.go.  This will help with creating skyledger coins as well.

iketheadore:
Nodes are not connected with each other. I changed the default peers in skycoin.go to 172.16.200.*:6000(172.16.200.2:6000, 172.16.200.3:6000,...), all nodes in containers are up, but they're not connected.

olemis:
@iketheadore thanks for your feedback . We are taking a look into this .

olemis:
@stdevMauricio I am gonna need you to update your branch with the latest code I uploaded so that changes be reflected in this PR .

PR #922, opened 2018-01-30
Fix 32bit overflow

iketheadore:


src/coin/transactions.go
```diff
@@ -535,7 +535,7 @@ func addUint32(a, b uint32) (uint32, error) {
 }
 
 func intToUint32(a int) (uint32, error) {
-	if a > math.MaxUint32 {
+	if uint(a) > math.MaxUint32 {
```
gz-c: oh, this method should also check a < 0 first

iketheadore: The `intToUint32` function is not used, maybe we can delete it, or change the parameter type `int` to `int64`. In 32bit architecture, `int` is `int32`, and the test code in [coin/transactions_test.go](https://github.com/skycoin/skycoin/blob/develop/src/coin/transactions_test.go#L943):
```go
func TestIntToUint32(t *testing.T) {
	n, err := intToUint32(10)
	require.Equal(t, uint32(10), n)
	require.NoError(t, err)

	_, err = intToUint32(math.MaxUint32 + 1)
	require.Error(t, err)
}
```
will fail, cause the `math.MaxUint32` overflows the `int32`, also no `int32` value would be able to overflow the `math.MaxUint32` in 32 bit architecture.

gz-c: Can remove it, was going to use it for checking txn.Size(), but maybe not necessary

iketheadore: Ok, I will remove it.

Issue #919, opened 2018-01-30
Transaction inputs data


Senyoret1:
Various api endpoints returning transaction data return only the hash of the inputs, but it is necessary to return more data, so that the api can be used in a more practical way. The inputs data should include the owning address, the coins and the hours. The endpoints with the problem are `/block`, `/blocks`, `/transaction` and `/pendingTxs` (there could be others, but those are the ones I've been using).

However, the change could break the compatibility with the systems that are currently using the api, because the current response have the following format:

```
"inputs": [
    "0922a7b41d1b76b6b56bfad0d5d0f6845517bbd895c660eab0ebe3899b5f63c4",
    "d73cf1f1d04a1d493fe3480a00e48187f9201bb64828fe0c638f17c0c88bb3d9",
    "16dd81af869743599fe60108c22d7ee1fcbf1a7f460fffd3a015fbb3f721c36d"
],
```

It would be ideal to change them to respond with the same format used by `/explorer/address`:

```
"inputs": [{
    "uxid": "0922a7b41d1b76b6b56bfad0d5d0f6845517bbd895c660eab0ebe3899b5f63c4",
    "owner": "2Q2VViWhgBzz6c8GkXQyDVFdQUWcBcDon4L",
    "coins": "7.000000",
    "hours": "851106"
}],
```

However, this change could give compatibility problems. Therefore, another alternative would be the following, which should work correctly:

```
"inputs": [
    "0922a7b41d1b76b6b56bfad0d5d0f6845517bbd895c660eab0ebe3899b5f63c4",
    "d73cf1f1d04a1d493fe3480a00e48187f9201bb64828fe0c638f17c0c88bb3d9",
],
"inputs_data": [
    {
        "owner": "2Q2VViWhgBzz6c8GkXQyDVFdQUWcBcDon4L",
        "coins": "7.000000",
        "hours": "851106"
    },
    {
        "owner": "2Q2VViWhgBzz6c8GkXQyDVFdQUWcBcDon4L",
        "coins": "8.000000",
        "hours": "751106"
    }
],
```

I would like to know @gz-c 's opinion regarding which option to take.

Additionally, before making any changes, it is necessary to take into account the problem that `/explorer/address` currently has and wait to see what alternative is available: https://github.com/skycoin/skycoin/issues/918

gz-c:
We can add "inputs_data", but each object in that array should have "uxid" in it.

When we clean up the API we'll remove "inputs" and rename "inputs_data" to "inputs"

[...]

Issue #918, opened 2018-01-30
Problems with /explorer/address


Senyoret1:
Doing some changes to the explorer I found that `/explorer/address` is returning an incorrect number of hours in the inputs. In the following image (taken from a modified version of the explorer) you can see an example:
![input](https://user-images.githubusercontent.com/34079003/35548473-04f210fe-0556-11e8-91cc-07fae095dd21.png)

In the image you can see that q6UjfkBA... received an output with 380 coins and 269 on 1/25/18, 7:35 AM. On 1/26/18, 5:48 AM the address used that same output as an input in a new transaction and it says that the input had 269 hours, the same amount as when it was received, but the outputs resulting from the transaction have more than 2000 hours each, which means that the input clearly had much more than 269 hours when it was used.

The results shown in the image correspond to transactions 6 and 7 of the followind call to the node's api: http://127.0.0.1:8001/api/address?address=q6UjfkBAhB8s9DjWnLKjgkTVs8j5a6bt4f

Modifications should be made so that `/explorer/address` returns the correct number of hours of the inputs or, if that is not possible, eliminate the number of hours of the inputs in the responses returned by the endpoint. Additionally, I would like to know if it is viable to solve the problem, to make some decisions in the development of the explorer.

I'm not sure about how to fix the problem, but the following data may be helpfull: the code that handles the `/explorer/address` endpoint is on line 147 of /src/gui/explorer.go. The problem seems to be on line 185, because `gateway.GetUxOutByID(id)` (function that is called to obtain the data of the inputs) always returns the number of hours the outputs had when they were received.

gz-c:
Must not be recalculating the CoinHours based upon block execution time

gz-c:
#920

gz-c:
@therealssj can you confirm if this is done?

gz-c:
#919 and #920 cover this issue

The problem stated here is because the hours are the original input hours and not the calculated hours

Issue #915, opened 2018-01-30
Remove the JSON RPC interface


gz-c:
The RPC interface is a reimplemented subset of the HTTP API.

The CLI tool communicates with the RPC interface.

Remove the RPC interface.

Requirements:

The CLI tool must process the same arguments and return the same data structures as it currently does.  This may require some data transformation, or in the worst case, new endpoints.

The CLI tool should communicate to the HTTP API, not the RPC API.  This requires a new Client object for communicating with the HTTP API (similar to the Client object used to communicate to the HTTP API).

Remove any `EXPOSE` directives in Docker containers for port 6430 (used by the RPC interface).

PR #928, opened 2018-02-01
Transaction inputs data

Senyoret1:
For solving https://github.com/skycoin/skycoin/issues/919, the inputs data was included in the following api endpoints: `/wallet/spend`, `/pendingTxs`, `/transaction`, `/block`, `/blocks`, `/last_blocks` and `/transactions`.

The new information was added in a field identified as "inputs_data". The responses now have the following format:

```
"txn": {
    "length": 317,
    "type": 0,
    "txid": "89578005d8730fe1789288ee7dea036160a9bd43234fb673baa6abd91289a48b",
    "inner_hash": "cac977eee019832245724aa643ceff451b9d8b24612b2f6a58177c79e8a4c26f",
    "sigs": [
        "3f084a0c750731dd985d3137200f9b5fc3de06069e62edea0cdd3a91d88e56b95aff5104a3e797ab4d6d417861af0c343efb0fff2e5ba9e7cf88ab714e10f38101"
    ],
    "inputs": [
        "bb89d4ed40d0e6e3a82c12e70b01a4bc240d2cd4f252cfac88235abe61bd3ad0"
    ],
    "inputs_data": [
        {
            "uxid": "bb89d4ed40d0e6e3a82c12e70b01a4bc240d2cd4f252cfac88235abe61bd3ad0",
            "owner": "nu7eSpT6hr5P21uzw7bnbxm83B6ywSjHdq",
            "coins": "59.000000",
            "hours": "418"
        }
    ],
    "outputs": [
        {
            "uxid": "ec9cf2f6052bab24ec57847c72cfb377c06958a9e04a077d07b6dd5bf23ec106",
            "dst": "nu7eSpT6hr5P21uzw7bnbxm83B6ywSjHdq",
            "coins": "60.000000",
            "hours": 2458
        }
    ]
}
```

The `inputs` field was not eliminated, to avoid compatibility problems with the systems currently using the Api. The input hash is repeated in `inputs_data`, as it was suggested in https://github.com/skycoin/skycoin/issues/919#issuecomment-361491658.

The way in which the api returns the transactions was changed, for that reason the modification included  more end points than those that were mentioned in https://github.com/skycoin/skycoin/issues/919.

As the modification was relatively long, I will place some comments in the code, to make it easier to understand the motivation of some changes.

Important: the changes have affected the rpc api, now the transactions returned by various of the endpoints contain an empty `inputs_data` array. I did not fix it because that should not create compatibility problems and the there are plans to make important changes to the rpc api (https://github.com/skycoin/skycoin/issues/915). However, if necessary, I can make the changes to the rpc api code so that the array returns the correct data.
 
Important 2: all the data added in this pr suffers from the problem indicated in https://github.com/skycoin/skycoin/issues/918. However, when creating a solution, it will be easy to apply it to all the api endpoints simply by modifying the `GetInputData` function in src/daemon/gateway.go.


src/daemon/gateway.go
```diff
@@ -183,14 +184,99 @@ func (gw *Gateway) GetBlockBySeq(seq uint64) (block coin.SignedBlock, ok bool) {
 	return
 }
 
+// GetInputData returns the data of an inputs
```
Senyoret1: `GetInputData`, `GetTransactionInputsData`, `GetBlockInputsData`, `GetSignedBlockInputsData` y `GetSignedBlocksInputsData` were added as a centralized way to access the transactions inputs data.

src/gui/explorer.go
```diff
@@ -182,15 +182,9 @@ func getTransactionsForAddress(gateway Gatewayer) http.HandlerFunc {
 					return
 				}
 
-				uxout, err := gateway.GetUxOutByID(id)
```
Senyoret1: This function creates the response of `/explorer/address` and already included the inputs data. It was modified to use the functions that were added in gateway.go and thus comply with DRY.

src/gui/transaction.go
```diff
@@ -160,7 +181,7 @@ func getTransactions(gateway Gatewayer) http.HandlerFunc {
 		}
 
 		// Converts visor.Transaction to visor.TransactionResult
-		txRlts, err := visor.NewTransactionResults(txns)
```
Senyoret1: `visor.NewTransactionResults(txns)` is in rpc.go. A new version was created at the end of this file (modified to include the inputs data) to mantain it separated from the rpc code.

src/visor/readable.go
```diff
@@ -443,7 +454,7 @@ func NewGenesisReadableTransaction(t *Transaction) (*ReadableTransaction, error)
 }
 
 // NewReadableTransaction creates readable transaction
-func NewReadableTransaction(t *Transaction) (*ReadableTransaction, error) {
+func NewReadableTransaction(t *Transaction, inData []*historydb.UxOut) (*ReadableTransaction, error) {
```
Senyoret1: By adding the `inData` parameter to `NewReadableTransaction`, now all the functions that want to obtain a transaction in the appropriate format for a JSON response must send an slice with the details of the inputs. This requirement was put in place to mantain consistency in the api and was the reason why all the endpoints that return transactions were modified.

The situation of the `NewReadableBlock` and `NewReadableBlocks` is similar.

An important detail is that this function could receive a gateway as a parameter and thus autonomously search the inputs data, without having to always do it before calling `NewReadableTransaction`, which would allow to eliminate repeated code. It was not done that way to maintain the format, since in this file there are no functions that autonomously search for data. However, if desired, the modification can be made.

src/visor/rpc.go
```diff
@@ -20,7 +21,7 @@ func NewTransactionResult(tx *Transaction) (*TransactionResult, error) {
 		return nil, nil
 	}
 
-	rbTx, err := NewReadableTransaction(tx)
+	rbTx, err := NewReadableTransaction(tx, make([]*historydb.UxOut, 0, 0))
```
Senyoret1: The objective of this pr was not to fix the rpc api. However, due to the changes made in readable.go, it was necessary to send an empty array when calling `NewReadableTransaction`. This is the reason of the “Important” note in the initial pr message.

Senyoret1:
The changes made the /rpc and /gui tests stop working. Should I fix them in this pr or in a new one?

gz-c:
@Senyoret1 the tests must pass for the pr to be merged

Senyoret1:
The last commit, in addition to the changes to make the tests pass, modified the the rpc api, now all the endpoints returns the inputs data. It also fixes some problems in the Readme.md of the rpc api.

src/api/webrpc/gateway.go
```diff
@@ -12,6 +12,11 @@ import (
 
 // Gatewayer provides interfaces for getting skycoin related info.
 type Gatewayer interface {
+	GetInputData(in cipher.SHA256) (*historydb.UxOut, error)
+	GetTransactionInputsData(tx *coin.Transaction) ([]*historydb.UxOut, error)
+	GetBlockInputsData(block *coin.Block) ([][]*historydb.UxOut, error)
+	GetSignedBlockInputsData(block *coin.SignedBlock) ([][]*historydb.UxOut, error)
+	GetSignedBlocksInputsData(blocks []coin.SignedBlock) ([][][]*historydb.UxOut, error)
```
iketheadore: Keep interface as small as possible, have GetUxoutByID() in `Gatewayer`, move the `GetTransactionInputsData` `GetBlockInputsData` etc, to the place where you want to call.

src/api/webrpc/gatewayer_mock_test.go
```diff
@@ -26,6 +26,31 @@ func NewGatewayerMock() *GatewayerMock {
...
+// GetSignedBlockInputsData returns the inputs data of a signed block
+func (m *GatewayerMock) GetSignedBlockInputsData(block *coin.SignedBlock) ([][]*historydb.UxOut, error) {
+	return nil, nil
+}
+
+// GetSignedBlocksInputsData returns the inputs data of a signed blocks slice
+func (m *GatewayerMock) GetSignedBlocksInputsData(blocks []coin.SignedBlock) ([][][]*historydb.UxOut, error) {
+	return nil, nil
+}
+
```
iketheadore: The `gatewayer_mock_test.go` is generated by `goautomock`, do not add code there,  otherwise the code will be wiped after running the goautomock.  run `go install`, then `go generate` to automaticlly generate mocker for new methods in Gatewayer interface. check the code in gui/transaction_test.go to see how to use the generated mocker.

src/api/webrpc/gateway.go
Senyoret1: You would like this change to be made only in api/webrpc/gateway.go or also in gui/gateway.go.

Senyoret1:
Would you like `GetTransactionInputsData`, `GetBlockInputsData`, etc removed only form api/webrpc/gateway.go or also from gui/gateway.go.

gui/gateway.go has not yet been updated with `goautomock`

iketheadore:
Remove from gui/gateway.go too, please

Senyoret1:
The methods were removed from gui/gateway.go. Also, gui/gatewayer_mock_test.go was updated with `goautomock`

Senyoret1:
Now the methods are private. Also, I saw that the methods in the `gui` package are mostly private, so I made the change there as well, to maintain consistency.

About removing `if inData`, as you say, if `inData` is `nil` then `len(inData)` returns is 0, so `inInfo` is initialized as a slice of size 0. The problem is that the automatic tests send `inData` as `nil` and expect `inInfo` to be `nil` as well, not a slice of size 0. It seems to me that what the tests expect is adequate, because it is consistent that if the inputs slice is `nil`, the inputs data slice should also de `nil`. However, if you wish, I can remove the `if ` and modify the tests so that they expect a slice of size 0 instead of `nil`.

iketheadore:
Remove the `inInfo = make([]ReadableTransactionInput, len(inData))` and use `append` to add new item, in this way, if the `inData` slice is nil, the returned `inInfo` would be nil.

iketheadore:
Also move `getInputData`, `getSignedBlocksInputsData`, etc., functions to the mehods of `Gateway` struct, and keep them private.

Senyoret1:
I fixed the readable.go code. In fact, it works correctly simply by removing the `if`, as you said, so I returned the code to how it was before the commit. I probably put the if for a behavior that was present only before updating the branch with the changes of "upstream/develop" (there were changes in the tests).

About the second change, I’m sorry, but it is not clear to me how would you like the change to be made. The `Gateway` structure is in the `daemon` package, so if I put the functions there, the `gui` package will not be abble to access them any more. The only way would be to make them public. Additionally, for them to be part of the structure, they would also have to be added to the interface, since the `gui` functions do not directly receive `daemon.Gateway`, but rather an implementation of the `gui.Gatewayer` interface.

iketheadore:
Sorry for the unclear info, I pulled and checked the code again, and I think the input data should be filled in gatewayer.GetBlock, gatewayer.GetTransaction, etc., instead of calling separate functions outside.

Cause each calling of gateway's method triggers a daemon event loop, we'd better get all things done in one event loop. And if calling outside, you have to define `getTransactionsInputData` in gui package, and again in webrpc package.  

In order to fill the input data in gateway.GetBlock and gateway.GetTransaction, etc., we need to change the return type of gateway.GetTransaction from coin.Transaction to ReadableTransaction, so as the gateway.GetBlock*, etc.  And have functions to convert the Readable* functions to coin.Transaction and coin.SignedBlock, incase they are required.

Senyoret1:
I changed the code so that the input data is filled inside `gatewayer.GetBlock`, `gatewayer.GetTransaction`, etc. I also made the necessary modifications to the tests and re-created the two “gatewayer_mock_test.go” files with `goautomock`...

There were only problems with `GetTransaction`, since in the line 230 of “gui/transaction.go” the function was required to continue returning the data as before. To avoid doing a double conversion (to `RedeableTransaction` and then again to `Transaction`), I modified `GetTransaction` to return the data as `TransactionResult` and `Transaction`.

Additionally, I eliminated three tests that are no longer applicable, because they were used to check errors when converting Gateway's responses to readable structures (The Gateway's responses  already have this conversion).

Senyoret1:
Solved a conflict with the latest commits. Apparently, Travis' test failed due to a new integration test:

`make: execvp: ./ci-scripts/integration-test-stable.sh: Permission denied`

gz-c:
Your merge made those scripts non-executable (changed perms from 0755 to 0644)

src/daemon/gateway.go
```diff
@@ -543,13 +638,36 @@ func (gw *Gateway) GetUnconfirmedTxns(addrs []cipher.Address) []visor.Unconfirme
...
 
 // GetLastTxs returns last confirmed transactions, return nil if empty
-func (gw *Gateway) GetLastTxs() ([]*visor.Transaction, error) {
+func (gw *Gateway) GetLastTxs() ([]*visor.TransactionResult, error) {
 	var txns []*visor.Transaction
+	var ret []*visor.TransactionResult
 	var err error
 	gw.strand("GetLastTxs", func() {
 		txns, err = gw.v.GetLastTxs()
+		if err == nil {
```
iketheadore: Return early

Senyoret1:
Github desktop messed with the permissions during the merge. Thanks for pointing it out.

Senyoret1:
Changed the code to return early.

Senyoret1:
Solved a conflict with the latest commits. The pr is getting a little messy, a could create a new one with all the changes if you wish.

On the other hand, Travis reported an error in a test that previously worked correctly, that was not affected by any of the previous changes and that works for me. In fact, Travis says that "Invalid address length" was returned instead of "Invalid base58 character", something that, when looking at the code, should never have happened (in fact I did several local tests and it never happened). Is there any known problem with this?

therealssj:
You could cherry pick and force update the changes into this branch only?

gz-c:
This won't be merged until after v22 is released.  You can start the PR over after that if you have too many issues with conflicts

Senyoret1:
If this won't be merged before v22, then I guess it's best to do what you say, because lately there have been many changes and conflicts. I'll wait until version 22 to make all the changes without so many commits.

Senyoret1:
@gz-c Is it now a good time to go back to work on this?

therealssj:
@Senyoret1 I am working on it :)

Senyoret1:
@therealssj Good :-)

gz-c:
Fixing these merge conflicts is not worth it, we will add this for v24

PR #924, opened 2018-01-30
CSRF Implementation

therealssj:
1. Add CSRF check to all POST request
2. Add `/csrf` endpoint to fetch the current csrf token or refresh it if expired
3. Add `-disable-csrf` flag 
4. Update existing tests

**Tasks**
- [x] Backend
- [x] Frontend
- [ ] Tests

Resolves: #896

PR #937, opened 2018-02-03
CSRF, Origin & Referer checks

### LINKED ISSUE #896: Add CSRF token to the API
Code Changes:

* Make sure all endpoints that modify state are POST requests
* Require all POST requests to have an `X-CSRF-Token` header
* Create an endpoint handler wrapper which will check for the header if the request method is POST
* If the token is invalid or expired return `403 Forbidden` with an error message indicating the failure
* Update gui/README.md with documentation about the CSRF token
* Add a `-disable-csrf` option to the CLI arguments which will disable CSRF checking in the API

CSRF Token:

* Create a `CSRFToken` object:
```go
type CSRFToken struct {
    Value []byte
    ExpiresAt time.Time
}
```
* Create a `CSRFStore` which holds a *single* `CSRFToken` protected by a mutex lock
* `CSRFToken.Value` is 64 CSPRNG bytes
* Return `CSRFToken.Value` as a base64 url-safe encoded string
* Tokens expire in 30 seconds
* There should be only one instance of `CSRFStore`

Client:

* Include `X-CSRF-Token` in all POST requests
* It's okay to make a call to `/csrf` prior to each request, otherwise retain the token
* If the API rejects the token, request a new one
* Possibly, abort if there are too many sequential calls to `/csrf` (this indicates a problem in the backend and the client may enter a busy loop otherwise)

### LINKED ISSUE #931: Check Origin and Referer header
Similar to #891 

https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Verifying_Same_Origin_with_Standard_Headers

If the Origin header is present, check that value
else if the Referer header is present, check that value,
if neither are present, reject

gz-c:
Fixes #896 
Fixes #931 

Extension of #924 

Note:

CSRF should never be disabled when using the wallet UI.  In case this happens, the frontend should detect any failure to /csrf and show a large warning to the user to close their wallet and enable CSRF.


src/gui/wallet_test.go
```diff
@@ -1276,6 +1266,7 @@ func TestWalletNewSeed(t *testing.T) {
 	}
 }
 
+// this is already tested in TestCSRFWrapper
 func TestWalletNewAddressesHandler(t *testing.T) {
```
therealssj: the test was changed so need to add tests to this

gz-c:
@samuelvisscher check that the csrf functionality is correct in `develop` branch and add the warning/abort panel if `/csrf` is failing in the wallet UI.

PR #954, opened 2018-02-07
Add bloom filter

stgleb:


gz-c:
What issue is this related to?

stgleb:
@gz-c This is  implementation of todo for adding index for block and blockchain. Bloom filter may help to check whether tx belongs to block or not faster in O(1).

gz-c:
There is no reason to add this, most blocks have 1 transaction, coin package shouldn't be modified either, we are not adding optimizations yet either when the db interface is wrong (see #622)

PR #976, opened 2018-02-10
Refactor http api integration tests

therealssj:
Follow some more patterns as in the standard library
- [x] Add update flag
- [x] Move test files to test-fixtures folder


src/gui/integration/integration_test.go
```diff
@@ -215,7 +235,11 @@ func TestStableOutputs(t *testing.T) {
 			require.NoError(t, err)
 
 			var expected visor.ReadableOutputSet
-			loadJSON(t, tc.golden, &expected)
+			goldenFile := filepath.Join("test-fixtures", tc.golden)
```
gz-c: This can't be done automatically in `loadJSON`?

therealssj: using the filename in `updateGoldenFile` as well

PR #987, opened 2018-02-12
Fix #981 update docs of wallet.NewWallet and add empty seed checking in wallet.GenerateAddresses

### LINKED ISSUE #981: docs and implementation of NewWallet and Wallet.GenerateAddresses need a review
While trying to use skycoin/services#6, with the seed, generated by skycoin/services#12, I have found following issues:

Docs for `wallet.NewWallet` claims that wallet generates a random seed if seed is "". Also it does not mention which encoding is expected.

However:

1. `wallet.NewWallet` returns error if seed is "" - https://github.com/skycoin/skycoin/blob/feaab10d4f951b8ffea53ef2d37a4cf5e3450b17/src/wallet/wallet.go#L80-L82

2.  `Wallet.GenerateAddresses` does not check for empty seed and on the first pass does not generate new one. Also, on the first pass it converts seed directly to byte representation, while the rest of the code assumes that seed is hex-encoded. 
https://github.com/skycoin/skycoin/blob/feaab10d4f951b8ffea53ef2d37a4cf5e3450b17/src/wallet/wallet.go#L212-L225

3. wallet.NewService uses bip39 to generate wallet's seed and does not encode it to hex.
https://github.com/skycoin/skycoin/blob/feaab10d4f951b8ffea53ef2d37a4cf5e3450b17/src/wallet/service.go#L52-L61

iketheadore:
* Updated wallet.NewWallet doc
* Add empty seed checking in wallet.GenerateAddresses

Fixes #981

### DIFF
--- a/src/wallet/wallet.go
+++ b/src/wallet/wallet.go
@@ -75,8 +75,7 @@ type Options struct {
 
 // NewWallet generates Deterministic Wallet
 func NewWallet(wltName string, opts Options) (*Wallet, error) {
-	seed := opts.Seed
-	if seed == "" {
+	if opts.Seed == "" {
 		return nil, errors.New("seed required")
 	}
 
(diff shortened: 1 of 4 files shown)


src/wallet/wallet.go
```diff
@@ -208,14 +211,14 @@ func (w *Wallet) GenerateAddresses(num uint64) ([]cipher.Address, error) {
 		return nil, nil
 	}
 
-	if w.getLastSeed() == "" {
+	if w.getSeed() == "" {
```
gz-c: also check getLastSeed

PR #1002, opened 2018-02-24
[util] fixes #966 closes #1036 - Expand data-dir paths relative to working directory

### LINKED ISSUE #966: -data-directory does not support relative directories
The -data-directory is always placed relative to $HOME, not the current directory.

olemis:
refs #966 #1036


iketheadore:
This will change the default DataDirectory of `~/.skycoin` to working directory + ./skycoin.

gz-c:
i think we need a separate method for defaulting to HOME/.skycoin and from handling relative paths

olemis:
@iketheadore @gz-c I'll write tests for ensuring the default data-dir location is handled appropriately . This will consist in ensuring that paths starting with tilde `~` e.g. `~/.skycoin` will still be expanded from user home .

olemis:
@iketheadore & @gz-c I added the changes for initial config of `--data-dir` config option . Tests included as well .

olemis:
sorry , I'll check for that . I am curious . Does the test I added recently in 7cc785216 fail for you ?

olemis:
@iketheadore tests did not fail for me bacause my `$GOPATH` was set to a subfolder of my home folder . Sorry about that ...

/me fixing

olemis:
@iketheadore @gz-c after giving this a lot of thought , I think that the fastest, most elegant, readable and intuitive solution seems to be to expand env vars . The changes in e60bf5c7c44 implement this . The (improved) test now passes and checks for an exact match in default path .

@iketheadore could you please review . Thnx in advance

olemis:
In Windows there is no `$HOME` env var , but [envWithHome](https://github.com/skycoin/skycoin/pull/1002/files#diff-c75af6da00e0aeb348fb282704f6ec39R71) makes sure that `UserHome()` is returned instead in that case. It *should* work on Windows , but I have no Windows machine to test it .

iketheadore:
It works on Windows 10.

gz-c:
Is there a point to doing shell expansion?  The shell would already expand the var if provided on the command line, so we are only substituting the `$HOME` token in the default string with `UserHome()`.

Would it better to detect when the user did not configure the data directory, and if they did not, then use `UserHome()/.skycoin`?

olemis:
@gz-c I'll submit another branch with another approach for you to review .

olemis:
@gz-c @iketheadore pls review #1036

PR #1025, opened 2018-03-03
Test gateway wallet methods return error when wallet api disabled

### LINKED ISSUE #887: Test gateway wallet methods return error when wallet api disabled
#877 Adds -disable-wallet-api CLI option.

This configures the daemon Gateway to return ErrWalletApiDisabled.

However, the tests only cover the HTTP endpoints, and mock the response from Gatewayer, so the Gateway methods themselves are not tested to block the wallet API when tested.  That is, the tests check that the HTTP handlers will return a 403 when it receives ErrWalletApiDisabled from the daemon Gateway, but they do not test that the Gateway returns ErrWalletApiDisabled when it is disabled in the config.

Vavilen:
- gw.GetWalletDir
- gw.NewAddresses
- gw.UpdateWalletLabel
- gw.GetWallet
- gw.GetWallets
- gw.GetWalletUnconfirmedTxns
- gw.ReloadWallets
- gw.Spend
- gw.CreateWallet
- gw.ScanAheadWalletAddresses
- gw.GetWalletBalance

Fix #887


src/daemon/gateway.go
```diff
@@ -592,14 +592,14 @@ func (gw *Gateway) GetWalletBalance(wltID string) (wallet.BalancePair, error) {
...
-
-		coins1, hours1, err := gw.v.AddressBalance(auxs)
+		var coins1, hours1 uint64
+		coins1, hours1, err = gw.v.AddressBalance(auxs)
 		if err != nil {
 			err = fmt.Errorf("Computing confirmed address balance failed: %v", err)
 			return
 		}
-
-		coins2, hours2, err := gw.v.AddressBalance(auxs.Sub(spendUxs).Add(recvUxs))
```
Vavilen: Here is a err shadowing.

src/daemon/gateway.go
```diff
@@ -592,14 +592,14 @@ func (gw *Gateway) GetWalletBalance(wltID string) (wallet.BalancePair, error) {
 			err = fmt.Errorf("get unconfirmed receiving failed when when checking wallet balance: %v", err)
 			return
 		}
-
-		coins1, hours1, err := gw.v.AddressBalance(auxs)
```
Vavilen: Here is a err shadowing.

gz-c:
Why are all the interface changes needed? Can't you configure the daemon with walletapidisabled and check that it returns the right error if true?

Vavilen:
@gz-c ,
I decided to cover all test cases for wallet related gateway methods.
Do you think that is incorrect?
Should I remove all other cases and leave only ErrWalletApiDisabled related cases?

gz-c:
@Vavilen test only ErrWalletApiDisabled cases for now, the interface changes are not ready to be used

PR #1036, opened 2018-03-05
[util] [cmd] fixes #966 closes #1002 - Expand -data-dir relative to $PWD . OS env vars not supported

### LINKED ISSUE #966: -data-directory does not support relative directories

olemis:
refs #966 #1002 
Fix #966


src/util/file/file_test.go
```diff
@@ -92,6 +92,15 @@ func TestUserHome(t *testing.T) {
...
 }
 
+func TestBuildDataDirDefault(t *testing.T) {
+	home := UserHome()
+	defaultDir := filepath.Join(home, ".skycoin")
+	dir, err := buildDataDir(defaultDir)
+	require.Nil(t, err)
+	expectedPath := filepath.Join(home, ".skycoin")
+	require.Equal(t, dir, expectedPath)
+}
```
gz-c: Are there tests for relative and absolute directory inputs too?

olemis: This one could work for absolute path . I'll double check that we have test cases for relative paths .

gz-c: I would add an explicit absolute path test to be clear:

* One test using prefixed `UserHome()/.skycoin`
* One test using some path like `/opt/.skycoin`
* One test using some path like `../config/.skycoin`

olemis: Some test cases have to be modified

olemis: - Test for default path `UserHome()/.skycoin` implemented before in this issue
- Test for `/opt/.skycoin` added in 09f5832 
- Test for dot paths (e.g. `../config/.skycoin` ) was already there

PR #1040, opened 2018-03-06
GUI integration test: /explorer/address

gz-c:


ci-scripts/integration-test-live.sh
```diff
@@ -12,11 +12,13 @@ HOST="http://127.0.0.1:$PORT"
 RPC_ADDR="127.0.0.1:$RPC_PORT"
 MODE="live"
 TEST=""
+UPDATE=""
 
 usage () {
   echo "Usage: $SCRIPT"
   echo "Optional command line arguments"
   echo "-t <string>  -- Test to run, gui or cli; empty runs both tests"
+  echo "-u <boolean> -- Update stable testdata"
```
therealssj: Why do we need to update stable data when running live tests?

gz-c: Some of the live tests use stable data where possible

Issue #627, opened 2017-11-22
Connection pool empties when running without network for a while


gz-c:
```
[skycoin.daemon:DEBUG] Failed to connect to 116.196.111.122:6000 with error: dial tcp 116.196.111.122:6000: getsockopt: connection refused
[skycoin.pex:DEBUG] Increase retry times of 116.196.111.122:6000: 4
[skycoin.daemon:DEBUG] Failed to connect to 174.31.25.197:6000 with error: dial tcp 174.31.25.197:6000: getsockopt: connection refused
[skycoin.pex:DEBUG] Increase retry times of 174.31.25.197:6000: 3
[skycoin.daemon:DEBUG] Failed to connect to 91.105.75.60:6000 with error: dial tcp 91.105.75.60:6000: getsockopt: network is unreachable
[skycoin.pex:DEBUG] Increase retry times of 91.105.75.60:6000: 2
[skycoin.visor:DEBUG] visor.Visor.RefreshUnconfirmed
[skycoin.visor:DEBUG] UnconfirmedTxnPool.RefreshUnconfirmed
[skycoin.daemon:DEBUG] Failed to connect to 80.133.128.176:6000 with error: dial tcp 80.133.128.176:6000: i/o timeout
[skycoin.pex:DEBUG] Increase retry times of 80.133.128.176:6000: 1
[skycoin.gnet:ERROR] daemon.gnet.ConnectionPool.BroadcastMessage error: Connection pool is empty
[skycoin.gnet:ERROR] req.Func daemon.gnet.ConnectionPool.BroadcastMessage failed: Connection pool is empty
[skycoin.daemon:ERROR] daemon.Visor.AnnounceBlocks error: Connection pool is empty
[skycoin.daemon:ERROR] Visor request func failed: Connection pool is empty
[skycoin.daemon:DEBUG] Broadcast AnnounceBlocksMessage failed: Connection pool is empty
[skycoin.daemon:DEBUG] Trying to connect to 63.142.253.76:6000
[skycoin.gnet:DEBUG] Making TCP Connection to 63.142.253.76:6000
[skycoin.daemon:DEBUG] Trying to connect to 76.164.192.236:6000
[skycoin.daemon:DEBUG] Trying to connect to 188.226.135.69:6000
[skycoin.gnet:DEBUG] Making TCP Connection to 188.226.135.69:6000
[...]

PR #1050, opened 2018-03-06
Fix #1044 missing trusted connection error

### LINKED ISSUE #1044: /network/connections/trust should not remove trusted peers even they are down
If peers are removed from trusted peers when they are down,  eventually, the users may get no default connections.

iketheadore:
Fix #1044


gz-c:
I think this will fix https://github.com/skycoin/skycoin/issues/627 too

PR #1053, opened 2018-03-06
Do not create new wallet or read wallet if the wallet API is disabled

### LINKED ISSUE #1046: Do not create new wallet or read wallet if the wallet API is disabled

Vavilen:
Fix #1046

### DIFF
--- a/cmd/skycoin/skycoin.go
+++ b/cmd/skycoin/skycoin.go
@@ -537,6 +537,7 @@ func configureDaemon(c *Config) daemon.Config {
 	dc.Visor.Config.DBPath = c.DBPath
 	dc.Visor.Config.DBReadOnly = c.DBReadOnly
 	dc.Visor.Config.Arbitrating = c.Arbitrating
+	dc.Visor.Config.DisabledWalletAPI = c.DisableWalletApi
 	dc.Visor.Config.WalletDirectory = c.WalletDirectory
 	dc.Visor.Config.BuildInfo = visor.BuildInfo{
 		Version: Version,
--- a/cmd/skycoin/skycoin.go
+++ b/cmd/skycoin/skycoin.go
@@ -537,7 +537,7 @@ func configureDaemon(c *Config) daemon.Config {
 	dc.Visor.Config.DBPath = c.DBPath
 	dc.Visor.Config.DBReadOnly = c.DBReadOnly
 	dc.Visor.Config.Arbitrating = c.Arbitrating
-	dc.Visor.Config.DisabledWalletAPI = c.DisableWalletApi
+	dc.Visor.Config.DisableWalletAPI = c.DisableWalletApi
 	dc.Visor.Config.WalletDirectory = c.WalletDirectory
 	dc.Visor.Config.BuildInfo = visor.BuildInfo{
 		Version: Version,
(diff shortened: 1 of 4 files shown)


cmd/skycoin/skycoin.go
```diff
@@ -537,6 +537,7 @@ func configureDaemon(c *Config) daemon.Config {
 	dc.Visor.Config.DBPath = c.DBPath
 	dc.Visor.Config.DBReadOnly = c.DBReadOnly
 	dc.Visor.Config.Arbitrating = c.Arbitrating
+	dc.Visor.Config.DisabledWalletAPI = c.DisableWalletApi
```
gz-c: Should be `DisableWalletAPI` not `DisabledWalletAPI`

src/wallet/service_test.go
```diff
@@ -422,7 +422,7 @@ func (dug dummyUnspentGetter) Get(uxid cipher.SHA256) (coin.UxOut, bool) {
 func TestServiceCreateAndSignTx(t *testing.T) {
 	dir := prepareWltDir()
 
-	s, err := NewService(dir)
+	s, err := NewService(dir, false)
```
gz-c: Need test with disableWalletAPI=true

Vavilen: @gz-c ,
Added tests with disableWalletAPI=true.
Skipped TestServiceCreateAndSignTx test, because logic changes with DisableWalletAPI=true.
Added return with ErrWalletApiDisabled for CreateWallet, because got panic without DisableWalletAPI=true param check.
I can add this check for all wallet service methods if you don't mind.

src/wallet/service_test.go
```diff
@@ -292,9 +344,27 @@ func TestServiceCreateAndScanWallet(t *testing.T) {
...
 }
 
+func TestServiceNewAddressDisabledWalletAPI(t *testing.T) {
+	dir := prepareWltDir()
+	s, err := NewService(dir, true)
+	require.NoError(t, err)
+	dirIsEmpty(t, dir)
+
+	// get the default wallet id
+	var id string
```
gz-c: remove this loop, does nothing, just check s.wallets is empty

src/wallet/service.go
```diff
@@ -24,16 +24,20 @@ type BalanceGetter interface {
 // Service wallet service struct
 type Service struct {
 	sync.RWMutex
-	wallets        Wallets
-	firstAddrIDMap map[string]string // key: first address in wallet, value: wallet id
-
-	WalletDirectory string
+	wallets          Wallets
+	firstAddrIDMap   map[string]string // key: first address in wallet, value: wallet id
+	DisableWalletAPI bool
```
iketheadore: If `DisableWalletAPI ` won't be used outside, we should make it private.

src/wallet/service_test.go
Vavilen: Fixed.

src/wallet/service.go
Vavilen: Fixed.

gz-c:
@iketheadore please check again

iketheadore:
Do not open the browser if the `-disable-wallet-api` is specified.

gz-c:
@iketheadore we can make that a separate issue

PR #1059, opened 2018-03-08
Fix gui live integration test

### LINKED ISSUE #1052: GUI integration test TestLiveAddressTransactions failed
The `height` in `status` is not stable, when blockchain grows, the height changes.

image

iketheadore:
* Fix #1052, recalculate transaction height in status field when do addressTransactions
* /richlist endpoint is not stable in live test, the number of returned items is not equal to 20, I comment out the check to pass the testing.


gz-c:
@iketheadore why is richlist not stable in live? when no n is specified, the default number of items is 20. should always be 20

src/gui/integration/integration_test.go
```diff
@@ -1632,6 +1632,15 @@ func TestLiveAddressTransactions(t *testing.T) {
 
 func testAddressTransactions(t *testing.T, cases []addressTransactionsTestCase) {
 	c := gui.NewClient(nodeAddress())
+	isLiveTest := doLive(t)
```
gz-c: make two separate tests, dont put too much conditions

iketheadore:
The /richlist returns 5k+ items when do live test with no n specified. I will remove the comment if it should only return 20 items.

gz-c:
Oh yes it returns all when not provided, that was a mistake i changed in stable test

gz-c:
Just check `require.NotEmpty` then

gz-c:
richlist-default.golden will need to be regenerated

src/gui/integration/integration_test.go
```diff
@@ -1627,10 +1627,10 @@ func TestLiveAddressTransactions(t *testing.T) {
 		},
 	}
 
-	testAddressTransactions(t, cases)
+	testLiveAddressTransactions(t, cases)
 }
 
-func testAddressTransactions(t *testing.T, cases []addressTransactionsTestCase) {
+func testStableAddressTransactions(t *testing.T, cases []addressTransactionsTestCase) {
```
gz-c: you can move these methods back into the main test method, doesn't need to be separate methods anymore

iketheadore: The `height` transaction `status` in live test is not stable.

gz-c: i mean put `testStableAddressTransactions` in `TestStableAddressTransactions` and `testLiveAddressTransactions` in `TestLiveAddressTransactions`

iketheadore: oh, ok

PR #1060, opened 2018-03-09
writing home for electron tests since compiling

stdevHan:
writing start for electron tests from the compiled, configuring test.js for tests with spectron + mocha + chai

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1,0 +1,45 @@
+const Application = require("spectron").Application;
+const path = require('path');
+const chai = require('chai');
+const should = chai.should();
+const chaiAsPromised = require('chai-as-promised');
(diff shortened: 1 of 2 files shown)


samuelvisscher:
Hi Han, thank you for your pull request, but I'm not entirely sure what I'm supposed to do with this. Could you give us some more information?

stdevHan:
Hello visscherio, is to be able to have a forced foot, to be able to write the tests with Spectron, example cases # 1011.

samuelvisscher:
Could you add a real example test (for example asserting that the gui has been loaded) and clean up the code?

olemis:
@samuelvisscher we'll try to look into this in the next few days . Sorry for the delay to reply .

Issue #939, opened 2018-02-04
CLI integration tests


gz-c:
Need integration tests for the CLI tool.

The integration tests should use a running skycoin node and use the cli tool against it.  Output should be verified for basic operations.

#610 is similar but is for unit testing the createRawTx cli interface

This can be broken down into multiple issues/PRs:

* Basic infrastructure/setup of the CLI integration test suite
* Tests for each CLI command

It does not have to be 100% automated in the start.  It can require some setup to be performed by the operator (such as running a skycoin node and providing a wallet file).

Priority commands for testing:

- `generateAddresses`
- `send`
- `send -m` (send-to-many)
- `broadcastTransaction`
- `createRawTransaction`
- `getWalletBalance`
- `transaction`
- `status`

CLI command checklist:

- [x] `addressBalance`
- [x] `addressGen`
- [x] `addressOutputs`
- [x] `blocks`
- [x] `broadcastTransaction`
- [x] `createRawTransaction`
- [x] `decodeRawTransaction`
- [x] `generateAddresses`
- [x] `generateWallet`
- [x] `lastBlocks`
- [x] `listAddresses`
- [x] `listWallets`
- [x] `send`
- [x] `status`
- [x] `transaction`
- [x] `version`
- [x] `walletBalance`
- [x] `walletDir`
- [x] `walletHistory`
- [x] `walletOutputs`
- [x] `checkdb`
- [x] `verifyAddress`

gz-c:
#974 started this

PR #1061, opened 2018-03-09
Add CLI send integration test

### LINKED ISSUE #982: CLI integration test should confirm coinhour cap behavior
The behavior was added by #955

iketheadore:
* Add  `send` cli live integration test
* Add `uxouts` command
* Fix #982 , verify the coinhours
* Ref #939 

Tested list:

- [x] Spend with change
- [x] Spend without change
- [x] Spend 1 output
- [x] Spend multiple outputs
- [x] Spend decimal coin with coin hour cap
- [x] Spend more than 1 coin
- [x] Spend 0.001 coin
- [x] Spend invalid small decimal coin, like 0.000999
- [x] Spend coin with 0 coin hour


src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1357,236 @@ func TestLiveWalletDir(t *testing.T) {
...
+	output, err := exec.Command(binaryPath, "walletBalance", walletPath).CombinedOutput()
+	require.NoError(t, err)
+	var bal cli.BalanceResult
+	err = json.NewDecoder(bytes.NewReader(output)).Decode(&bal)
+	require.NoError(t, err)
+
+	coins, err := droplet.FromString(bal.Confirmed.Coins)
+	require.NoError(t, err)
+
+	hours, err := droplet.FromString(bal.Confirmed.Hours)
```
therealssj: hours are not stored as droplets

src/api/webrpc/gateway.go
```diff
@@ -19,5 +19,6 @@ type Gatewayer interface {
 	GetTransaction(txid cipher.SHA256) (*visor.Transaction, error)
 	InjectBroadcastTransaction(tx coin.Transaction) error
 	GetAddrUxOuts(addr cipher.Address) ([]*historydb.UxOutJSON, error)
+	GetUxOutByID(id cipher.SHA256) (*historydb.UxOut, error)
```
therealssj: needs to be implemented in all the mock interfaces?

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1357,236 @@ func TestLiveWalletDir(t *testing.T) {
...
+					checkCoinsAndCoinhours(t, tx, w.Entries[1].Address.String(), "0.500000", 1)
+					// Confirms the third address has 1 coin and 1 coinhour
+					checkCoinsAndCoinhours(t, tx, w.Entries[2].Address.String(), "0.500000", 1)
+				}
+
+				// If the total input coin hours is <= 1, the two recipient addresses should receive no coinhours.
+				if hours <= 1 {
+					// Confirms the second address has 0.5 coin and 1 coinhour
+					checkCoinsAndCoinhours(t, tx, w.Entries[1].Address.String(), "0.500000", 0)
+					// Confirms the third address has 1 coin and 1 coinhour
```
gz-c: comment wrong? 0 coinhour?

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1357,236 @@ func TestLiveWalletDir(t *testing.T) {
...
+				if hours >= 4 {
+					// Confirms the second address has 0.5 coin and 1 coinhour
+					checkCoinsAndCoinhours(t, tx, w.Entries[1].Address.String(), "0.500000", 1)
+					// Confirms the third address has 1 coin and 1 coinhour
+					checkCoinsAndCoinhours(t, tx, w.Entries[2].Address.String(), "0.500000", 1)
+				}
+
+				// If the total input coin hours is <= 1, the two recipient addresses should receive no coinhours.
+				if hours <= 1 {
+					// Confirms the second address has 0.5 coin and 1 coinhour
```
gz-c: comment wrong? 0 coinhour?

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1357,236 @@ func TestLiveWalletDir(t *testing.T) {
...
+				// Only when hours >= 4, the two recipient addresses could all receive 1 coinhours
+				if hours >= 4 {
+					// Confirms the second address has 0.5 coin and 1 coinhour
+					checkCoinsAndCoinhours(t, tx, w.Entries[1].Address.String(), "0.500000", 1)
+					// Confirms the third address has 1 coin and 1 coinhour
+					checkCoinsAndCoinhours(t, tx, w.Entries[2].Address.String(), "0.500000", 1)
+				}
+
+				// If the total input coin hours is <= 1, the two recipient addresses should receive no coinhours.
+				if hours <= 1 {
```
gz-c: I think the test should always require a certain number of coinhours, so that the test does not have different behavior depending on the number of coinhours.  We can create spends to control the number of coinhours in another address to test what happens with 1 coinhour and 0 coinhours.

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1357,236 @@ func TestLiveWalletDir(t *testing.T) {
...
+				return []string{"send", "-c", w.Entries[2].Address.String(), w.Entries[0].Address.String(), "1"}
+			},
+			func(t *testing.T, txid string) {},
+		},
+		{
+			"send to one address",
+			func() []string {
+				return []string{"send", w.Entries[0].Address.String(), "2"}
+			},
+			func(t *testing.T, txid string) {},
```
gz-c: This test doesn't check tx?

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1357,236 @@ func TestLiveWalletDir(t *testing.T) {
...
+				}
+			},
+		},
+		{
+			"send with -c(change address) flag",
+			func() []string {
+				// Set the third address as change address
+				return []string{"send", "-c", w.Entries[2].Address.String(), w.Entries[0].Address.String(), "1"}
+			},
+			func(t *testing.T, txid string) {},
```
gz-c: This test doesn't check tx?

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1357,236 @@ func TestLiveWalletDir(t *testing.T) {
...
+			v := bytes.Split(output, []byte(":"))
+			require.Len(t, v, 2)
+			txid := string(v[1])
+			fmt.Println("txid:", txid)
+			_, err = cipher.SHA256FromHex(txid)
+			require.NoError(t, err)
+
+			// Wait untill transaction is confirmed.
+			for {
+				time.Sleep(time.Second)
```
gz-c: Should use a for select loop with a ticker for checking confirmation and a ticker for timeout (maybe 30 seconds timeout?)

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1357,236 @@ func TestLiveWalletDir(t *testing.T) {
...
+		_, err = w.GenerateAddresses(uint64(3 - len(w.Entries)))
+		require.NoError(t, err)
+	}
+
+	coins, hours := getWalletBalance(t, walletPath)
+	if coins < 2 {
+		t.Fatal("Wallet must have at least 2 coins")
+	}
+
+	miniCoinHours := uint64(1 << uint(testCasesNum))
```
gz-c: This depends on `BurnFactor = 2` which will adjust later and is different for other coins

src/api/cli/uxouts.go
```diff
@@ -0,0 +1,43 @@
+package cli
+
+import (
+	"fmt"
+
+	"github.com/skycoin/skycoin/src/cipher"
+	gcli "github.com/urfave/cli"
+)
+
+func uxoutsCmd() gcli.Command {
```
gz-c: Add this command to CHANGELOG.md (but don't mention RPC, it's still deprecated)

src/api/cli/integration/integration_test.go
iketheadore: Oh, yes, this is a mistake.

src/api/cli/integration/integration_test.go
iketheadore: Yes, comment wrong.

therealssj:
Instead of calculating the `uxouts` why don't we just calculate the address balance difference?
compare the difference of address balance before and after the transaction to see if it matches the required amount.

iketheadore:
@therealssj there's no `uxouts` calculation anymore. What's the mean of comparing the address balance difference?

therealssj:
oh sorry my bad.
balance difference as in 
lets say we start with:
```
addr1: 10 coins
addr2: 5 coins
```

we send `0.5` coins to `addr2` from `addr1` so to confirm the test we make sure that `addr2` balance = 5.5 or we can check the balance of both addresses as well

gz-c:
Is it possible to have the code setup the wallet to the correct initial conditions instead of requiring the user to provide wallet that satisfies it already?

We will want to add many different style of spend tests so we cannot force the wallet to have a certain number of outputs.  The code should create the outputs it needs to work with.  Each test can generate the empty addresses it needs, and send the coins to them in a certain way to create the outputs they want to work with.

iketheadore:
@gz-c yes, we can setup the wallet, the user only need to provide a wallet that has total coins and coin hours meet the requirements.

gz-c:
Some spend test cases off of my head:

Should pass:
* Spend without change
* Spend with change
* Spend 1 output
* Spend multiple outputs [should repeat this to discover any intermittent bug like before]
* Spend 0.001 or more coins

Should fail:
* Spend 0 coin hours
* Spend 0.000999 or less coins

One example test would be to take the wallet as-is,

* check that it has enough coin hours (minimum is 16)
* combine all inputs to 1 output
* split 1 output to N outputs for M address [e.g. 5 outputs to 3 different addresses], of various coin amounts
* send some but not all outputs to an M+1 address, repeatedly
* send all from M+1 address to second address, create multiple outputs, all have 0 coin hours except 1
* send all from second address to first address, now it has 0 coin hours
* try to send from first address to second address, should fail, it has 0 coin hours

This takes a minimum of 5 steps.  More, if "send to M+1, repeatedly" repeats.  5 steps requires 16 coin hours.  The repeat step can repeat as long as it has at least 4 coin hours.  Max repeat should be between 6 or 10. 

Before all this, can test the invalid decimals that are too small.  During this, make sure some outputs have amounts like 0.001, 0.01, 0.1, 1, 1.1, 1.01, 1.001.

iketheadore:
For now, we have tested:
* Spend with change
* Spend without change
* Spend 1 output
* Spend multiple outputs
* Spend decimal coin with coin hour cap
* Spend more than 1 coin

The `TestLiveSend` function cases:

* The first test case spend from 1 output to 3 outputs, one output as change, and two to recipient addresses. In this case, we tested the `Spend with change`, `Spend 1 output`, `Spend decimal coin with coin hour cap`. 

* The second test case tested `-c` `-a` options of `send` command, which also tested the `Spend with change`.

* The third test case put all inputs to 1 output, which tested `Spend multiple outputs`, `Spend without change` and `Spend more than 1 coin`.

TODO:

-[ ] `Do the put all inputs to 1 output test first, so that user only need to prepare wallet with sufficient coins and hours`
-[ ] `Test 0.001 coin`
-[ ] `Test too small decimal coin, like 0.000999`
-[ ] `Test 0 coin hour`

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1339,311 @@ func TestLiveWalletDir(t *testing.T) {
...
+			nil,
+			func(t *testing.T, txid string) {
+				tx := getTransaction(t, txid)
+				// Confirms the second address receives 0.5 coin and 1 coinhour in this transaction
+				checkCoinsAndCoinhours(t, tx, w.Entries[1].Address.String(), 500000, 1)
+				// Confirms the third address receives 0.5 coin and 1 coinhour in this transaction
+				checkCoinsAndCoinhours(t, tx, w.Entries[2].Address.String(), 500000, 1)
+				// Confirms the first address has at least 1 coin left.
+				coins, _ := getAddressBalance(t, w.Entries[0].Address.String())
+				require.True(t, coins >= 1000000)
```
gz-c: 1e6 notation is better

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1339,311 @@ func TestLiveWalletDir(t *testing.T) {
...
+
+				return []string{"send", "-m", string(v)}
+			},
+			nil,
+			func(t *testing.T, txid string) {
+				tx := getTransaction(t, txid)
+				// Confirms the second address receives 0.5 coin and 1 coinhour in this transaction
+				checkCoinsAndCoinhours(t, tx, w.Entries[1].Address.String(), 500000, 1)
+				// Confirms the third address receives 0.5 coin and 1 coinhour in this transaction
+				checkCoinsAndCoinhours(t, tx, w.Entries[2].Address.String(), 500000, 1)
```
gz-c: 5e5 notation is better

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1339,311 @@ func TestLiveWalletDir(t *testing.T) {
...
+					"-a", w.Entries[2].Address.String(), w.Entries[1].Address.String(), "0.001"}
+			},
+			nil,
+			func(t *testing.T, txid string) {
+				tx := getTransaction(t, txid)
+				// Confirms the second address receives 0.5 coin and 0 coinhour in this transaction
+				checkCoinsAndCoinhours(t, tx, w.Entries[1].Address.String(), 500000, 0)
+				// Confirms the second address have 1 coin and 1 coin hour
+				coins, hours := getAddressBalance(t, w.Entries[1].Address.String())
+				require.Equal(t, uint64(1000000), coins)
```
therealssj: 1e6 notation and I don't think you need to explicitly wrap it around uint64 here

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1339,311 @@ func TestLiveWalletDir(t *testing.T) {
...
+			},
+			[]byte("ERROR: Transaction has zero coinhour fee. See 'skycoin-cli send --help'"),
+			func(t *testing.T, txid string) {},
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			output, err := exec.Command(binaryPath, tc.args()...).CombinedOutput()
+			if err != nil {
```
therealssj: shouldn't we also print the error here?

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1339,311 @@ func TestLiveWalletDir(t *testing.T) {
...
+				fmt.Println(string(output))
+				return
```
therealssj: In case of an error  it will just `return` and never go to `require.NoError`

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1339,311 @@ func TestLiveWalletDir(t *testing.T) {
...
+			v := bytes.Split(output, []byte(":"))
+			require.Len(t, v, 2)
+			txid := string(v[1])
+			fmt.Println("txid:", txid)
+			_, err = cipher.SHA256FromHex(txid)
+			require.NoError(t, err)
+
+			// Wait untill transaction is confirmed.
+			tk := time.NewTicker(time.Second)
+		loop:
```
therealssj: what is this?

src/api/cli/integration/integration_test.go
```diff
@@ -1341,3 +1339,311 @@ func TestLiveWalletDir(t *testing.T) {
...
+
+	if len(w.Entries) < 3 {
+		// Generates addresses
+		_, err = w.GenerateAddresses(uint64(3 - len(w.Entries)))
+		if err != nil {
+			t.Fatalf("Wallet generateAddress failed: %v", err)
+		}
+	}
+
+	outputs := getWalletOutputs(t, walletPath)
```
therealssj: we should move this above `len(w.Entries)` check

src/api/cli/integration/integration_test.go
iketheadore: Yes, for now, the error message are in the output, if any error, we should not continue the following test code. But instead of just print the output, I should run t.Fatalf() to fail the test.

therealssj: yes, sorry that's what I meant to say that this should fail the test somewhere

src/api/cli/integration/integration_test.go
iketheadore: No, if the wallet doesn't have any address entries, we won't get any outputs from the wallet. We do `len(w.Entries)` check first, and if the wallet has less than 3 address, we will generate more addresses.

therealssj: The test checks if the wallet has no coins, if the wallet has no coins creating more addresses will not generate coins.

Issue #948, opened 2018-02-06
API integration tests


gz-c:
Need integration tests for the HTTP API.

The blockchain APIs will be easier to test than the wallet APIs which modify data.

For the blockchain API integration tests, we could automate this in travis.  The skycoin node would need an option (if it does not have one) to disable the p2p networking.  We would include a default blockchain data.db test file, from which the blockchain APIs could read from.

API method checklist:

- [x] `/version`
- [x] `/outputs`
- [x] `/outputs?addrs=`
- [x] `/outputs?hashes=`
- [x] `/coinSupply`
- [x] `/block?hash=`
- [x] `/block?seq=`
- [x] `/blocks`
- [x] `/last_blocks`
- [x] `/blockchain/metadata`
- [x] `/blockchain/progress`
- [x] `/balance?addrs=`
- [x] `/uxout?uxid=`
- [x] `/address_uxouts`
- [x] `/wallet`
- [x] `/wallets`
- [x] `/wallet/create`
- [x] `/wallet/newAddress`
- [x] `/wallet/balance`
- [x] `/wallet/transactions`
- [x] `/wallet/update`
- [x] `/wallet/folderName`
- [x] `/wallet/spend`
- [x] `/wallet/newSeed`
- [x] `/network/connection`
- [x] `/network/connections`
- [x] `/network/defaultConnections`
- [x] `/network/connections/trust`
- [x] `/network/connections/exchange`
- [x] `/pendingTxs`
- [x] `/transaction`
- [x] `/transactions`
- [x] `/transactions?confirmed=true`
- [x] `/transactions?confirmed=false`
- [x] `/injectTransaction`
- [x] `/resendUnconfirmedTxns`
- [x] `/rawtx`
- [x] `/explorer/address`
- [x] `/richlist`
- [x] `/addresscount`

gz-c:
Important tests are:

* `/coinSupply`
* `/block`
* `/blockchain/progress`
* `/balance?addrs=`

gz-c:
`/coinSupply` is tested now (#967)

Test those others based upon the initial work in `src/gui/integration/`

iketheadore:
I will write the others.

gz-c:
Added tests for the above endpoint in #971

gz-c:
Continue testing all endpoints.

Spend tests will need to run separately, they cannot be run against the stable blockchain and cannot be automated in travis.  Test all of the other endpoints first

Vavilen:
[...]

PR #1070, opened 2018-03-11
GUI live integration test: /wallet/spend

iketheadore:
* Setup gui live integration test for wallet
* Spend all coins to the first address

#948

### DIFF
--- a/ci-scripts/integration-test-live.sh
+++ b/ci-scripts/integration-test-live.sh
@@ -15,22 +15,26 @@ TEST=""
 UPDATE=""
 # run go test with -v flag
 VERBOSE=""
+# run go test with -run flag
+RUN_TESTS=""
 
 usage () {
   echo "Usage: $SCRIPT"
   echo "Optional command line arguments"
   echo "-t <string>  -- Test to run, gui or cli; empty runs both tests"
+  echo "-r <string>  -- Test to run with -run flag"
   echo "-u <boolean> -- Update stable testdata"
   echo "-v <boolean> -- Run test with -v flag"
   exit 1
 }
 
-while getopts "h?t:uv" args; do
+while getopts "h?t:r:uv" args; do
 case $args in
     h|\?)
         usage;
         exit;;
     t ) TEST=${OPTARG};;
+    r ) RUN_TESTS=${OPTARG};;
     u ) UPDATE="--update";;
     v ) VERBOSE="-v";;
   esac
@@ -49,12 +53,12 @@ fi
 
 if [[ -z $TEST || $TEST = "gui" ]]; then
 
-SKYCOIN_INTEGRATION_TESTS=1 SKYCOIN_INTEGRATION_TEST_MODE=$MODE SKYCOIN_NODE_HOST=$HOST go test ./src/gui/integration/... $UPDATE -timeout=3m $VERBOSE
+SKYCOIN_INTEGRATION_TESTS=1 SKYCOIN_INTEGRATION_TEST_MODE=$MODE SKYCOIN_NODE_HOST=$HOST go test ./src/gui/integration/... $UPDATE -timeout=3m $VERBOSE -run=$RUN_TESTS
 
 fi
 
 if [[ -z $TEST || $TEST = "cli" ]]; then
 
-SKYCOIN_INTEGRATION_TESTS=1 SKYCOIN_INTEGRATION_TEST_MODE=$MODE RPC_ADDR=$RPC_ADDR go test ./src/api/cli/integration/... $UPDATE -timeout=3m $VERBOSE
+SKYCOIN_INTEGRATION_TESTS=1 SKYCOIN_INTEGRATION_TEST_MODE=$MODE RPC_ADDR=$RPC_ADDR go test ./src/api/cli/integration/... $UPDATE -timeout=3m $VERBOSE -run=$RUN_TESTS
 
 fi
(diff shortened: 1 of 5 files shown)


src/gui/client.go
```diff
@@ -410,7 +410,8 @@ func (c *Client) Spend(id, dst string, coins uint64) (*SpendResult, error) {
 	v.Add("coins", fmt.Sprint(coins))
 
 	var r SpendResult
-	if err := c.Post("/wallet/spend", strings.NewReader(v.Encode()), &r); err != nil {
+	endpoint := "/wallet/spend?" + v.Encode()
```
gz-c: Why is it in the query string for a POST request?

src/gui/integration/integration_test.go
```diff
@@ -1796,3 +1797,150 @@ func TestLivePendingTransactions(t *testing.T) {
 	_, err := c.PendingTransactions()
 	require.NoError(t, err)
 }
+
+func TestLiveWalletSpend(t *testing.T) {
+	if !doLive(t) {
+		return
+	}
+
+	w, totalCoins, _ := prepareAndCheckWallet(t, 2, 1)
```
therealssj: `2e6`

src/gui/integration/integration_test.go
```diff
@@ -1796,3 +1797,150 @@ func TestLivePendingTransactions(t *testing.T) {
...
+				// Confirms the address balance are equal to the totoalCoins
+				coins, _ = getAddressBalance(t, w.Entries[0].Address.String())
+				require.Equal(t, totalCoins, coins)
+			},
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			c := gui.NewClient(nodeAddress())
```
therealssj: why don't we declare this out of the `for` scope and then use that single instance at other places?

src/gui/integration/integration_test.go
```diff
@@ -1796,3 +1797,150 @@ func TestLivePendingTransactions(t *testing.T) {
...
+	wp, err := c.WalletBalance(walletName)
+	if err != nil {
+		t.Fatalf("Get wallet balance of %v failed: %v", walletName, err)
+	}
+
+	return wp.Confirmed.Coins, wp.Confirmed.Hours
+}
+
+func getTransaction(t *testing.T, txid string) *visor.TransactionResult {
+	c := gui.NewClient(nodeAddress())
```
therealssj: pass this function the `gui client` instance?

src/gui/integration/integration_test.go
```diff
@@ -1796,3 +1797,150 @@ func TestLivePendingTransactions(t *testing.T) {
...
+		t.Fatalf("%v", err)
+	}
+
+	return tx
+}
+
+// getAddressBalance gets balance of given address.
+// Returns coins and coin hours.
+func getAddressBalance(t *testing.T, addr string) (uint64, uint64) {
+	c := gui.NewClient(nodeAddress())
```
therealssj: same as above

src/gui/client.go
iketheadore: The spend handler was not a POST request.

gz-c: https://github.com/skycoin/skycoin/blob/develop/src/gui/wallet.go#L72-L77 ?

iketheadore: Yes, the POST check was added later, but didn't change the parameters. Not sure if changing the parameter would break the exchange platform.

gz-c: It uses r.FormValue which will prefer the PostForm data but fallback on the query string.  So the gui.Client should work with POST form data

iketheadore: Yes, change request's Content-Type: application/x-www-form-urlencoded, and run r.ParseForm() in the handler works.

src/gui/client.go
```diff
@@ -410,8 +410,9 @@ func (c *Client) Spend(id, dst string, coins uint64) (*SpendResult, error) {
 	v.Add("coins", fmt.Sprint(coins))
 
 	var r SpendResult
-	endpoint := "/wallet/spend?" + v.Encode()
-	if err := c.Post(endpoint, nil, &r); err != nil {
+	// endpoint := "/wallet/spend?" + v.Encode()
```
gz-c: delete dead line

src/gui/wallet.go
```diff
@@ -76,6 +76,11 @@ func walletSpendHandler(gateway Gatewayer) http.HandlerFunc {
 			return
 		}
 
+		if err := r.ParseForm(); err != nil {
```
gz-c: Isn't ParseForm called automatically?

iketheadore: Oh, it's called automatically, don't know why it didn't work yesterday, maybe I forgot to restart the node. I will remove this.

ci-scripts/integration-test-live.sh
```diff
@@ -49,12 +53,12 @@ fi
 
 if [[ -z $TEST || $TEST = "gui" ]]; then
 
-SKYCOIN_INTEGRATION_TESTS=1 SKYCOIN_INTEGRATION_TEST_MODE=$MODE SKYCOIN_NODE_HOST=$HOST go test ./src/gui/integration/... $UPDATE -timeout=3m $VERBOSE
+SKYCOIN_INTEGRATION_TESTS=1 SKYCOIN_INTEGRATION_TEST_MODE=$MODE SKYCOIN_NODE_HOST=$HOST go test ./src/gui/integration/... $UPDATE -timeout=3m $VERBOSE -run=$RUN_TESTS
```
gz-c: does this still work when RUN_TESTS is empty?

iketheadore: Yes,  it will run all tests when it's empty.

gz-c:
Update the checklist in #948 if necessary

PR #1074, opened 2018-03-12
Update/fix visor tests

### LINKED ISSUE #401: Update/fix visor tests
Some of the visor test files cannot compile so are marked `// +build ignore`.  These tests should be amended or removed as necessary.

Vavilen:
Done: visor/readable_test.go
Fixes #401

### DIFF
--- a/src/visor/readable_test.go
+++ b/src/visor/readable_test.go
@@ -1,52 +1,76 @@
-// +build ignore
-
 package visor
 
 import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"reflect"
 	"testing"
 	"time"
 
+	"github.com/boltdb/bolt"
+	"github.com/skycoin/skycoin/src/util/droplet"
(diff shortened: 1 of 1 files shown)


src/visor/readable_test.go
```diff
@@ -1,343 +1,274 @@
...
+func setupVisor(t *testing.T) (v *Visor, close func()) {
+	db, close := testutil.PrepareDB(t)
+	vc := setupVisorConfig(t)
+	v, err := NewVisor(vc, db)
+	require.NoError(t, err)
+	return
+}
+
+func cleanupVisor(dir string) {
+	err := os.RemoveAll(dir)
```
gz-c: Don't do this, in case there is a bug and we call os.RemoveAll on the wrong directory (this happened before)

Just leave the temp folders in temp

Vavilen: @gz-c , fixed.
Do I need remove wallet files?

gz-c:
@iketheadore please review

PR #1081, opened 2018-03-15
refactor /wallet/transactions response

therealssj:
skycoin/skycoin/issues/1080


therealssj:
oh okay

gz-c:
Can you show me what the API response is like now, so I can compare the change?

therealssj:
screen shot 2018-03-16 at 6 04 14 pm
@gz-c here is a screenshot of how it looks now

gz-c:
The gui.Client doesn't need any change?

therealssj:
@gz-c done

iketheadore:
The `/wallet/transactions` is not being used, maybe we can deprecate it.

PR #1083, opened 2018-03-15
Add instrumentation/timing to API calls

### LINKED ISSUE #537: Add instrumentation/timing to API calls

Vavilen:
Fix #537


gz-c:
merge conflict

src/gui/http.go
```diff
@@ -126,6 +130,17 @@ func (s *Server) Shutdown() {
 	<-s.done
 }
 
+func ElapseHandler(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+		handler.ServeHTTP(w, r)
+		since := time.Since(start)
+		if since > elapsedThreshold {
+			log.Printf("%s %s %v", r.Method, r.URL.Path, time.Since(start))
```
gz-c: use the package level `logger` instance

src/gui/http.go
```diff
@@ -126,6 +130,17 @@ func (s *Server) Shutdown() {
 	<-s.done
 }
 
+func ElapseHandler(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+		handler.ServeHTTP(w, r)
+		since := time.Since(start)
+		if since > elapsedThreshold {
```
gz-c: Always log, not just when time threshold is reached

src/gui/http.go
```diff
@@ -177,8 +199,10 @@ func newServerMux(c muxConfig, gateway Gatewayer, csrfStore *CSRFStore) *http.Se
 	}
 
 	webHandler := func(endpoint string, handler http.Handler) {
+
 		handler = CSRFCheck(csrfStore, handler)
 		handler = headerCheck(c.host, handler)
+		handler = ElapseHandler(handler)
```
gz-c: Apply the handler before the others so that it is always called

PR #1100, opened 2018-03-19
Stop serving wallet static files when -disable-wallet-api is true

### LINKED ISSUE #1031: Feature --disable-wallet-api should also block / from serving index.html
When wallet api's are disabled we should not be showing a gui either

iketheadore:
Fixes #1031 

No


gz-c:
merge conflict

PR #1101, opened 2018-03-19
remove SendOr404 and SendOr500 due to typed nils

### LINKED ISSUE #1021: Remove SendOr404 and SendOr500
`httphelper.SendOr404` and `httphelper.SendOr500` need to be removed. They don't work properly due to typed nils.

See: https://play.golang.org/p/sCLTQjpaR3I

The check for nil will have to be in the caller.  So all `wh.SendOr404(w, m)` calls become:

```go
if m == nil {
    wh.Error404(w, m)
    return
}

wh.SendJSONOr500(logger, w, m)
```

therealssj:
Fixes #1021 

yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- #1021, remove `SendOr404` and `SendOr500` as they do not work properly due to typed nils
 - Support absolute and relative paths for `-data-dir` option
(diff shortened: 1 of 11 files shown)


gz-c:
This can go into the "Fixed" section of the changelog

gz-c:
@therealssj none of these needed to be checked for nil?

therealssj:
so most of the functions were already previously checking for failures and that `SendOr404` was definitely not needed there as they would have failed beforehand anyways.
At other places it was an array so just returned an empty array

PR #1111, opened 2018-03-19
Add unit tests for iputil

stgleb:
Add unit tests for iputil

### DIFF
--- a/src/util/iputil/iputil_test.go
+++ b/src/util/iputil/iputil_test.go
@@ -1,0 +1,90 @@
+package iputil
+
+import "testing"
+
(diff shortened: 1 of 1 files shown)


src/util/iputil/iputil_test.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+}
+
+func TestIsLocalhost(t *testing.T) {
+	testData := []struct {
+		input string
+		host  string
+		port  uint16
+		err   bool
```
gz-c: Use `err error` and compare the exact error in the test

stgleb:
@gz-c Done

PR #1149, opened 2018-03-26
[util] fixes #858 - Replace op-logging with logrus

### LINKED ISSUE #858: Replace github.com/op/go-logging with logrus
https://godoc.org/github.com/sirupsen/logrus

This is used in teller and skywire

olemis:
refs #858 

Changes:
- Replace op-logging with logrus
- Op-logging removed
- Logrus used for logging
- New logging aliases `Critical` and `Notice` for high priority errors and info messages (respectively)
- Color highlighting for critical entries while running on TTY 
- Go deps added


src/util/logging/hooks.go
```diff
@@ -0,0 +1,86 @@
...
+		Logger: logger,
+		excludeTypes: map[reflect.Type]int{
+			reflect.TypeOf(ReplayHook{}):    1,
+			reflect.TypeOf(ModuleLogHook{}): 1,
+		},
+	}
+}
+
+// Do not replay hooks of given exclude types
+func ExclusiveReplayHook(logger *Logger, exclude []reflect.Type) (h ReplayHook) {
```
gz-c: What does the replay hook do?

olemis: It makes sure that hooks installed in the (default) top level logger are still executed for module-specific loggers . The later are just exact clones of the former also adding module-specific context. The separation is for implementing the same (`ExtendedFieldLogger`) interface for module loggers.

Hooks (if any) may be configured once for standard logger . By doing so all other module-specific loggers are immediate aware of the new configuration . This is expected behavior since the logging API does not provide access to the module-specific loggers , and it should not .

Type exclusion rules are needed so as to not to replay several module hooks (leading to log entries with wrong module tags) or another replay hook (e.g. leading to issues like invoking the hook chain twice, infinite loops , ...)

gz-c: `ExclusiveReplayHook` is not used anywhere?

Probably better to initialize with `[]interface{}` and do the reflection during construction, but if this isn't being used at all I'd remove it.

src/util/logging/hooks.go
```diff
@@ -0,0 +1,84 @@
...
+	h = NewReplayHook(logger)
+	h.excludeTypes = nil
+	for _, _type := range exclude {
+		h.excludeTypes[_type] = 1
```
gz-c: This will segfault because excludeTypes is nil, must be allocated to a `map[reflect.Type]int`

src/util/logging/hooks.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	return
+}
+
+// Hook for replaying hooks bound to another logger
+type ReplayHook struct {
+	Logger *Logger
+	// FIXME : Go lang sets?
+	excludeTypes map[reflect.Type]int
```
gz-c: The convention for sets is to map to the empty struct, e.g.

`excludeTypes map[reflect.Type]struct{}`

To add an element, you would do `excludeTypes[t] = struct{}{}`

src/util/logging/hooks.go
olemis: It's fine for slices to be nil https://tour.golang.org/moretypes/12

olemis: oh this is a map ... u r right

src/util/logging/logger.go
```diff
@@ -0,0 +1,153 @@
...
+func (l *Logger) Critical(args ...interface{}) {
+	l.WithField("priority", "CRITICAL").Error(args...)
+}
+
+func (l *Logger) Criticalln(args ...interface{}) {
+	l.WithField("priority", "CRITICAL").Error(args...)
+}
+
+func (l *Logger) Noticef(format string, args ...interface{}) {
+	l.WithField("priority", "CRITICAL").Info(args...)
```
iketheadore: Notice still use "CRITICAL" ?

iketheadore:
`e.logger.Warning("%s elapsed %s", *e.name, elapsed)` should use `e.logger.Warningf`

image

The logger in util/elapse/elapser.go need to be updated.

src/util/logging/logger.go
olemis: Yes . `Notice` is `Info` entry with `CRITICAL` priority . `Critical` and `Notice` are not logging levels , but helpers for custom priority field . I could set priority value to either `HIGH` or `HIGHEST` instead so as to avoid any confusion .

olemis:
@iketheadore I'd not recommend to remove `module` and `priority` fields out off entries with critical priority . Even if they are duplicated in tty output we'll definitely need them for machine processing of logs . Possible to suppress them in formatter output though .

olemis:
Elapser logging fixed in a899953

iketheadore:
`op/go-logging` is still being used in `util/droplet/droplet.go`.

src/util/logging/logger.go
gz-c: Can you add a constant for `"CRITICAL"` in the logger util library?

olemis:
@iketheadore fixed that and removed op/go-logging from vendor . Thanks

olemis:
anything else ?

gz-c:
We can clean up things like `Errorf("%v", err)` later

PR #1162, opened 2018-03-30
document coinhours spending

### LINKED ISSUE #1157: Document coinhour distribution defaults for CLI/HTTP API
Add a section about "Coin Hours" in here: 

https://github.com/skycoin/skycoin/blob/develop/INTEGRATION.md#sending-coins

And describe the difference in coinhour distribution between the CLI and the HTTP API

therealssj:
Fixes #1157 

Changes:
- Document coinhours spending when making a transaction through `REST API` vs `CLI`

No


gz-c:
Can you mention that a transaction must have coin hours to spend, and that they regenerate at a rate of 1 coinsecond per second, which is converted to hours?

PR #1180, opened 2018-04-04
Handle newline characters in seed input

### LINKED ISSUE #725: Handle trailing newline in create wallet's seed text box
If pasting a seed copied from a text file, it may have a newline appended.

The create wallet's seed input box is a textarea and not an `<input>`.  Pasting into an `<input>` would trim/ignore the trailing whitespace, but pasting into a textarea does not trim it.

Decide how to handle this, options include:

* Show a warning to the user if they have a trailing newline
* Always trim 1 trailing newline from the seed in the textarea
* Use an `<input>` box

This should be handled client-side, not in the backend.

mahansky:
Fixes #725 

Changes:
- newline characters are removed from seed input (also works for seeds copied as multiline text)
- starting/trailing whitespaces are ignored for convenience (when validating), but trimmed on submit

### DIFF
--- a/src/gui/static/src/app/components/pages/wallets/create-wallet/create-wallet.component.ts
+++ b/src/gui/static/src/app/components/pages/wallets/create-wallet/create-wallet.component.ts
@@ -28,12 +28,12 @@ export class CreateWalletComponent implements OnInit {
   }
 
   createWallet() {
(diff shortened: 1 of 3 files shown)


gz-c:
Don't transform data returned by the API

samuelvisscher:
Nice!

PR #1185, opened 2018-04-05
QR Codes

### LINKED ISSUE #1167: Generate QR code
URI specification: https://github.com/skycoin/skycoin#uri-specification

mahansky:
Fixes #1167 

Changes:
- after clicking the qrcode icon next to the addresses in wallets, modal shows up with generated qrcode image

The same thing happens on the "History" page, but only if there is just one address.


src/gui/static/src/app/components/pages/transaction-list/transaction-list.component.ts
```diff
@@ -38,4 +39,14 @@ export class TransactionListComponent implements OnInit, OnDestroy {
     config.data = transaction;
     this.dialog.open(TransactionDetailComponent, config).afterClosed().subscribe();
   }
+
+  showQrCode($event: any, address: string[]) {
```
samuelvisscher: Are you sure that you can just always show the first address in this case?

src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.ts
```diff
@@ -54,6 +55,11 @@ export class WalletDetailComponent {
     setTimeout(function () {
       address.copying = false;
     }, 500);
+  }
 
+  showQrCode(address: string) {
+    const config = new MatDialogConfig();
+    config.data = { address };
```
samuelvisscher: Don't you have to add a key for address? Does this compile or work?

mahansky: I send it there as a string, not an object
https://github.com/skycoin/skycoin/pull/1185/files/10b612704520814a8212effa9cd6abc83f332d02#diff-16f295f5e79f9cbf7bb4baf688e19cdcR5

samuelvisscher: Why not remove brackets?

src/gui/static/src/app/components/pages/transaction-list/transaction-list.component.ts
mahansky: There can be multiple addresses in a transaction, so I decided to add the ability to show a QR code only when there is exactly one.

I could transform multiple addresses into list (instead of a paragraph) and have the qrcode icon next to each other (that might take a lot of space and look weird)

src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.ts
mahansky: QRCode component expects an object (it might receive more parameters in the future)

gz-c:
@samuelvisscher is this fine now?

merge conflicts

mahansky:
There is still that question - what should happen when there are multiple addresses in a transaction (history page) and you click the QR code image. 

> There can be multiple addresses in a transaction, so I decided to add the ability to show a QR code only when there is exactly one.
> I could transform multiple addresses into list (instead of a paragraph) and have the qrcode icon next to each other (that might take a lot of space and look weird)

Senyoret1:
Pressing the qr code button in the transaction history page stops the propagation of the event. The same should be done on the wallet screen, to prevent the address from being copied at the same time as the qr code is displayed.

With this pr the QR codes now use BIP21, so the addresses start with "skycoin:". I guess there's no problem with that. @gz-c , does it seem good to you?

Regarding what to do when a transaction has more than one address, I do not see any problem with placing the addresses as a list, each one on a news line, it would just need an adequate margin between lines, to make it look good. In any case, leaving it as a paragraph it is also possible to have so many transactions that a single line is not enough. Going a little further, it is also possible to convert the qr buttons to inline elements and continue using the paragraph style, but I think the list style would probably look better.

gz-c:
We could leave out QR codes on the history page

mahansky:
screen shot 2018-04-27 at 14 05 58

PR #1187, opened 2018-04-05
Generate seed with different entropy

### LINKED ISSUE #800: Option to generate seeds with 24 words

mahansky:
Fixes #800 

Changes:
- added option to generate seed with 128 and 256 bit entropy to create wallet UI

Does this change need to be mentioned in CHANGELOG.md?
Yes


gz-c:
Update the CHANGELOG.md file if it needs updating

mahansky:
screen shot 2018-04-26 at 12 10 01

That's how it looks. Any ideas on how to improve it? I wanted to go with icons, but I couldn't find any relevant (I don't want to just put two same icons there)

Senyoret1:
Works correctly.

As for the appearance, what do you think of this?

seed_icons1

Probably just a few people will know what the difference between 128 and 256 bits is and there may be confusion, so I think using common words would be better.

The icons in that image are custom bitmaps. However, it is posible to create something similar without the need for custom bitmaps, using the following icons:

- Key: https://fontawesome.com/icons/key?style=solid
- Lock: https://fontawesome.com/icons/lock?style=solid

The end result is this:

seed_icons2

The thing is that Skycoin's graphic style uses delineated icons, not solid icons, so the custom bitmaps would be a more consistent option. If the first option is selected, I can send you the icons. If you want something more elaborate, I can try to do something else.

Additionally, in my opinion, there should be some kind of effect when the mouse pointer is over any of the options, as a visual feedback indicating that it is possible to click the element. Also, I think that these options in the wizard should be white instead of black.

gz-c:
Maybe instead of normal/strong, it can say "12 words" "24 words"

mahansky:
Since Material Icons are already in the project, I tried using them.
If you don't like the result, just send me your custom icons and I'll use those.

screen shot 2018-04-27 at 12 59 29

gz-c:
The different key/lock icons for 12/24 are confusing

Senyoret1:
Here are two more alternatives:
ico1

ico2

However, creating self explanatory icons in such a small space is difficult, so here are two more options:
ico3
ico4

An additional alternative, which would work if complemented with the tooltip, is this:
ico5

Personally, I think I would prefer the fourth option.

mahansky:
Here is one more style
screen shot 2018-04-30 at 12 14 57

I don't like it below the seed input, the dialog is too big as it is.
Also, I couldn't find two same icons (full shield and outlined shield) in Material Icons, so we would need images for the first style.

mahansky:
Ok, current (final) version

screen shot 2018-05-02 at 12 38 30

PR #1211, opened 2018-04-09
Unspent outputs

### LINKED ISSUE #1120: Ability to view unspent outputs for a wallet and addresses
Transactions spend "unspent outputs" which affect how coinhours are burned.  Add a way to view these unspent outputs.  Related to #872

mahansky:
Fixes #1120

Changes:
- "Outputs" page shows outputs in a more categorized way (wallets, addresses)
- Added button to wallets page to filter single address from outputs (similar to "Copy" address, just next to the coin balance)


src/gui/static/src/app/services/wallet.service.ts
```diff
@@ -81,6 +82,22 @@ export class WalletService {
       .flatMap(addresses => this.apiService.get('outputs', {addrs: addresses}));
   }
 
+  outputsWithWallets(): Observable<any> {
+    return Observable.zip(this.all(), this.outputs(), (wallets, outputs) => {
+      wallets = JSON.parse(JSON.stringify(wallets));
```
mahansky: I'm modifying wallets provided by wallet.service; since I don't want to modify the original object, I'm using this "hack" to create a copy of it. Any ideas on how to improve it?

mahansky: I'm modifying wallets provided by wallet.service; since I don't want to modify the original object, I'm using this "hack" to create a copy of it. Any ideas on how to improve it?

Senyoret1:
Just looking at the changes in the code, there seem to be no problems. However, I would like to make a more appropriate revision tomorrow.

gz-c:
@Senyoret1 I would prefer to minimize dependencies; what is the advantage of the other method?

PR #1215, opened 2018-04-10
Drop ngx-datatable

### LINKED ISSUE #860: Remove ngx-datatable from angular front-end
Due to the decreased use of tables in the front-end we should get rid of the ngx-datatable dependency

mahansky:
Fixes #860 

Changes:
- replaces ngx-datatable component with simple layouts

This PR conflicts with #1211 - I will fix it once one of them gets merged.

If you don't like styles I used, I can replace them with styles based on the "Wallets" page (grey rounded centered tables)


gz-c:
#1211 is merged

Senyoret1:
The pending transactions page has a layout problem with the "transaction id" column:

layout

Apart from that, there are some additional details that I would like to ask for opinions:

- The styles used are very similar to the previous ones. I think it would be better to use styles more similar to those on the wallets screen (as mentioned in the initial comment), but I think it is not a priority and could be done in another pr, including some additional UX improvements.

- With these changes, there are no more paging options. I think this should only cause problems in very few cases and could be solved in another pr.

- Strangely, Github shows me all the code of the previous commits while opening the "Merge branch 'develop' into drop_datatables" commit, but the "Files Changed" tab shows me just some files, so I supose there is no problem.

Issue #968, opened 2018-02-09
[lib] Export cipher and cli functions to be reused via libskycoin


olemis:
Cipher and CLI functionality will be needed elsewhere . It should be available for reuse in libskycoin .

Cc @gz-c feel free to highlight functions to be exported this way

gz-c:
It depends on how much you want to export versus how much you want to implement in the code using libskycoin.

At minimum you will want to export the cipher methods.  You could potentially reimplement the CLI methods as needed.

Looking at the teller code, which uses the cli library natively to sign transactions:

`cli.CreateRawTxFromWallet` would be the main cli method to export.

For `cipher`:

`cipher.GenerateKeyPair`
`cipher.DecodeBase58Address`
`cipher.GenerateDeterministicKeyPairs`

But likely most of the public methods in `cipher/address.go`, `cipher/hash.go`, `cipher/crypto.go` should be exported.

samuelvisscher:
This would also be great for the Java, Objective-C and Javascript libraries produced through GopherJS and GoMobile.

olemis:
@samuelvisscher do you need client libs for those languages ? We could build some as well (and test'em).

PR #1225, opened 2018-04-12
[lib] fixes #992 fixes #1182 fixes #1537 - libskycoin v0.24.0-alpha

### LINKED ISSUE #992: [lib] [test]  Test libskycoin cipher and cli functions
... as a continuation of the work done in #968

### LINKED ISSUE #1182: Compile Criterion in make install-deps-libc
Aforementioned make target currently downloads Citerion binaries . Those do not include functions needed by the test suite (e.g. assertions for custom types) . Instead Criterion bleeding branch should be checked out , compiled and installed locally in `./build` .

### LINKED ISSUE #1537: Libskycoin cipher tests
Use the cipher golden files test suite with `libskycoin` C library .

olemis:
Fixes #992 #968 #1537 

Changes:
- Expose `cipher` (address, hash, crypto) and `cli` (create_rawtx) methods.
- API docs generated by [Doxygen](https://github.com/skycoin/skycoin/issues/www.doxygen.org/) + [Moxygen](https://github.com/sourcey/moxygen)
- Tested !

yes


gz-c:
Tests are failing due to missing criterion.h headers

Does this call go code or is the go code converted to C?

Biggest concern would be that the seed generation uses the go lib, and that determinism is unaffected

olemis:
Criterion is the C testing framework

olemis:
@gz-c there is no golang to C translation . The go compiler has several build modes including the ones  that output shared and static libraries . So this is basically the go compiler doing exactly the same thing to obtain a binary file and a C header file with function definitions .

gz-c:
Merge conflict

.travis.yml
```diff
@@ -11,6 +11,11 @@ matrix:
     - os: osx
       # Do not start osx build for PR
       if: type != pull_request
+      osx_image: xcode8
```
gz-c: @iketheadore will this impact our release builds?

include/skystring.h
```diff
@@ -0,0 +1,15 @@
+#ifndef LIBSKY_STRING_H
+#define LIBSKY_STRING_H
+
+#include <stdio.h>
+#include <stdlib.h>
+#include "libskycoin.h"
+
+extern void randBytes(GoSlice *bytes, size_t n);
```
gz-c: Is this method only for tests? 

It its public to everyone it should use the cipher rand methods and not math/rand

lib/cgo/cipher.crypto.go
```diff
@@ -0,0 +1,278 @@
...
+func SKY_cipher_PubKey_ToAddressHash(_pk *C.PubKey, _arg0 *C.Ripemd160) {
+	pk := (*cipher.PubKey)(unsafe.Pointer(_pk))
+	h := pk.ToAddressHash()
+	copyToBuffer(reflect.ValueOf(h[:]), unsafe.Pointer(_arg0), uint(SizeofRipemd160))
+}
+
+//export SKY_cipher_NewSecKey
+func SKY_cipher_NewSecKey(_b []byte, _arg1 *C.SecKey) (errcode uint32) {
+	defer func() {
+		errcode = catchApiPanic(errcode, recover())
```
gz-c: Is there a problem with allowing it to panic?

The reason it panics is because if the condition ever occurs, the code should not be used at all or else there is a risk of fatal cryptographic errors

.travis.yml
olemis: according to travis docs that's needed for `gcc6` which is needed to compile C testing binaries.

include/skystring.h
olemis: Only for testing purposes . Maybe we should have two header include paths : the one for release and a another just for the test suite ... @gz-c makes sense ?

PR #1226, opened 2018-04-12
Improve wallet metadata API format

### LINKED ISSUE #1220: Wallet API returns boolean as string
@iketheadore #1189 

Endpoints `/wallet`, `/wallets`, `/wallets/create` return `encrypted` as string instead of boolean as other endpoints.

Expected output:
```diff
{
    "meta": {
        "coin": "skycoin",
        "cryptoType": "",
-       "encrypted": "false",
+       "encrypted": false,
        "filename": "***",
        "label": "Wallet",
        "lastSeed": "",
        "secrets": "",
        "seed": "",
        "tm": "1523265861",
        "type": "deterministic",
        "version": "0.1"
    },
    "entries": [
        {
            "address": "***",
            "public_key": "***",
            "secret_key": ""
        }
    ]
}
```

iketheadore:
Fixes #1220 

Changes:
- Remove seed, lastSeed, secrets meta fileds and secret in address entries of wallet apis response
- Change the "encrypted" type to boolean
- Rename the "tm" to "timestamp", and change the type from string to integer

Yes.

### DIFF
--- a/src/gui/wallet.go
+++ b/src/gui/wallet.go
@@ -28,26 +28,28 @@ type UnconfirmedTxnsResponse struct {
 	Transactions []visor.ReadableUnconfirmedTxn `json:"transactions"`
 }
 
-type walletEntry struct {
+// WalletEntry the wallet entry struct
+type WalletEntry struct {
 	Address string `json:"address"`
 	Public  string `json:"public_key"`
 }
 
-type walletMeta struct {
+// WalletMeta the wallet meta struct
+type WalletMeta struct {
 	Coin       string `json:"coin"`
 	Filename   string `json:"filename"`
 	Label      string `json:"label"`
 	Type       string `json:"type"`
 	Version    string `json:"version"`
-	CryptoType string `json:"cryptoType"`
+	CryptoType string `json:"crypto_type"`
 	Timestamp  int64  `json:"timestamp"`
 	Encrypted  bool   `json:"encrypted"`
 }
 
 // WalletResponse wallet response struct for http apis
 type WalletResponse struct {
-	Meta    walletMeta    `json:"meta"`
-	Entries []walletEntry `json:"entries"`
+	Meta    WalletMeta    `json:"meta"`
+	Entries []WalletEntry `json:"entries"`
 }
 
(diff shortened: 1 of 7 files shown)


src/gui/wallet.go
```diff
@@ -28,6 +28,98 @@ type UnconfirmedTxnsResponse struct {
 	Transactions []visor.ReadableUnconfirmedTxn `json:"transactions"`
 }
 
+type walletEntry struct {
+	Address string `json:"address"`
+	Public  string `json:"public_key"`
+}
+
+type walletMeta struct {
```
gz-c: these structs shouldn't be private

src/gui/wallet.go
```diff
@@ -28,6 +28,98 @@ type UnconfirmedTxnsResponse struct {
...
+	Coin       string `json:"coin"`
+	Filename   string `json:"filename"`
+	Label      string `json:"label"`
+	Type       string `json:"type"`
+	Version    string `json:"version"`
+	CryptoType string `json:"cryptoType"`
```
gz-c: crypto_type

src/gui/wallet.go
```diff
@@ -28,6 +28,98 @@ type UnconfirmedTxnsResponse struct {
...
+			Address: e.Address.String(),
+			Public:  e.Public.Hex(),
+		})
+	}
+
+	return &wr, nil
+}
+
+// ToWallet converts WalletResponse to wallet.Wallet
+func (wr *WalletResponse) ToWallet() (*wallet.Wallet, error) {
```
gz-c: Do we need this? Can gui.Client just return *WalletResponse?

src/gui/wallet.go
iketheadore: OK, will turn to public

src/gui/wallet.go
iketheadore: Hmm, maybe can let gui.Client return *WalletResponse, and move the convert function into integration_test.go, as the tests are using *wallet.Wallet.

gz-c: ok, do that

PR #1228, opened 2018-04-12
Scan address when create wallet

### LINKED ISSUE #1227: Scan ahead address when creating wallet
Create wallet with encryption by `/wallet/create` api is too slow, as it generates a encrypted wallet first, then decrypt and scan addresses, then encrypt it again. 

We should create unencrypted wallet first, then scan addresses, then encrypt it.

iketheadore:
Fixes #1227 

Changes:
- Scan addresses when creating wallet, only lock wallet once.

No


src/daemon/gateway.go
```diff
@@ -538,30 +538,15 @@ func (gw *Gateway) Spend(wltID string, password []byte, coins uint64, dest ciphe
 }
 
 // CreateWallet creates wallet
-func (gw *Gateway) CreateWallet(wltName string, options wallet.Options) (*wallet.Wallet, error) {
+func (gw *Gateway) CreateWallet(wltName string, options wallet.Options, scanN uint64) (*wallet.Wallet, error) {
```
gz-c: put scanN in options?

src/gui/wallet.go
```diff
@@ -326,12 +326,12 @@ func walletCreate(gateway Gatewayer) http.HandlerFunc {
 			}
 		}
 
-		wlt, err = gateway.ScanAheadWalletAddresses(wlt.Filename(), []byte(password), scanN-1)
-		if err != nil {
-			logger.Errorf("gateway.ScanAheadWalletAddresses failed: %v", err)
-			wh.Error500(w)
-			return
-		}
+		// wlt, err = gateway.ScanAheadWalletAddresses(wlt.Filename(), []byte(password), scanN-1)
```
gz-c: remove

src/gui/wallet_test.go
```diff
@@ -1260,8 +1236,11 @@ func TestWalletCreateHandler(t *testing.T) {
 	for _, tc := range tt {
 		t.Run(tc.name, func(t *testing.T) {
 			gateway := &GatewayerMock{}
-			gateway.On("CreateWallet", "", tc.options).Return(&tc.gatewayCreateWalletResult, tc.gatewayCreateWalletErr)
-			gateway.On("ScanAheadWalletAddresses", tc.wltName, tc.options.Password, tc.scnN-1).Return(&tc.scanWalletAddressesResult, tc.scanWalletAddressesError)
+			if tc.scnN == 0 {
+				tc.scnN = 1
+			}
+			gateway.On("CreateWallet", "", tc.options, tc.scnN).Return(&tc.gatewayCreateWalletResult, tc.gatewayCreateWalletErr)
+			// gateway.On("ScanAheadWalletAddresses", tc.wltName, tc.options.Password, tc.scnN-1).Return(&tc.scanWalletAddressesResult, tc.scanWalletAddressesError)
```
gz-c: remove

src/daemon/gateway.go
gz-c: or separate CreateWalletScanAhead method?

src/wallet/wallet.go
```diff
@@ -178,6 +178,77 @@ func NewWallet(wltName string, opts Options) (*Wallet, error) {
 	return w, nil
 }
 
+// NewWalletWithScan create wallet and scan ahead N address
+func NewWalletWithScan(wltName string, opts Options, scanN uint64, bg BalanceGetter) (*Wallet, error) {
```
gz-c: is this almost the same as `NewWallet`?  If so, can use something like `newWallet()` internal method with an option to scan

src/daemon/gateway.go
iketheadore: The CreateWallet in gateway always scan ahead addresses, so no need to have a separate method.

src/wallet/wallet.go
```diff
@@ -178,6 +194,16 @@ func NewWallet(wltName string, opts Options) (*Wallet, error) {
 	return w, nil
 }
 
+// NewWallet creates wallet without scanning addresses
+func NewWallet(wltName string, opts Options) (*Wallet, error) {
+	return newWallet(wltName, opts, nil)
```
gz-c: error if scanN != 0

PR #1234, opened 2018-04-16
[docker] fixes #1190 - Improvements in Dockerfile for skycoin/skycoin mainnet image

### LINKED ISSUE #1190: Improvements for skycoin Docker containers
Run integration tests ( [Docker Cloud](https://docs.docker.com/docker-cloud/builds/automated-testing/) ? ) prior to publishing containers onto Docker Hub .

stdevMauricio:
Fixes #1190

Changes:
- Upgrades to golang 1.10
- Removes angular build step
- Fixes permissions in volumes and executes the skycoin daemon as a normal user.
- Makes possible to pass arguments to the daemon.
- Updates README.md with changes

yes


docker/images/mainnet/Dockerfile
```diff
@@ -31,20 +31,21 @@ FROM alpine:3.7
 
 ENV COIN="skycoin" \
     RPC_ADDR="0.0.0.0:6430" \
-    DATA_DIR="/root/.$COIN" \
+    DATA_DIR="/data/.$COIN" \
```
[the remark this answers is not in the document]
stdevMauricio: Done

stdevMauricio:
I'd like to point out that on the mainnet Dockerfile I can't use the `docker_launcher` script. I made it to create a regular user for skycoin to run and fix the permissions accordingly, but is not available on the old code. The easiest fix is to document how to build the image using the source code as the context and change the Dockerfile accordingly.
In this case would be like this
```sh
$ SKYCOIN_VERSION="0.22.0"
$ docker build -t skycoin/skycoin:v${SKYCOIN_VERSION} \ 
--build-arg=SKYCOIN_VERSION=${SKYCOIN_VERSION}
```
Please tell me if that's OK and I'll proceed.

olemis:
@stdevMauricio , in my opinion, if you need the launcher since you recommend it's the right way to go looking forward then please do so. 

/me reviewing your changes .

README.md
```diff
@@ -119,14 +121,51 @@ $ docker run -ti --rm \
...
+
+The container will run with some default options, but you can change them
+by just appending flags at the end of the `docker run` command. The following
+example will show you the available options.
+
+```sh
+docker run --rm skycoin/skycoin -help
+```
 
 Access the dashboard: [http://localhost:6420](http://localhost:6420).
```
olemis: In order for this to work , isn't it necessary to bind ports in the host while `docker run`ing the container?

README.md
```diff
@@ -119,14 +121,51 @@ $ docker run -ti --rm \
...
+### Building your own images
+
+There are some folders in the `docker/images/` folder. You can build different
+images with the Dockerfiles they contain for different purposes.
+
+#### Mainnet
+
+When you build this Dockerfile, docker will download the source code from
+github on the specified tag and make an image out of it. The stable image
+should come from this one. The `--build-arg` is optional as there is a default
```
olemis: What's _this one_ in this context ? What's the `--build-arg` comment about ?

README.md
```diff
@@ -119,14 +121,51 @@ $ docker run -ti --rm \
...
+in the Dockerfile.
+
+```sh
+$ cd docker/images/mainnet
+$ docker build -t skycoin:stable --build-arg=SKYCOIN_VERSION=0.22.0
+```
```
olemis: Your comment about `--build-arg` is more clear if included after this .

README.md
```diff
@@ -119,14 +121,51 @@ $ docker run -ti --rm \
...
+
+#### Snapshot
+
+This Dockerfile it's used to build the current state of the source you are
```
olemis: _the source you are working on_ => _your local working copy or develop branch_ .

docker/images/snapshot/Dockerfile
```diff
@@ -1,26 +1,12 @@
 # skycoin build binaries
```
olemis: Is it possible to have a single Dockerfile to build both them images ?

olemis:
@gz-c I edited PR title because changes in here will not add neither testing nor continuous integration of Docker containers . It will be aimed at improving the mainnet `Dockerfile` .

README.md
stdevMauricio: No, that's just like running skycoin (default options) -help, so it displays the help and exits. It's just an example.

docker/images/snapshot/Dockerfile
stdevMauricio: It could be possible, but the user would have to checkout to the intended tag manually. The only difference between the two dockerfiles is where do they get the code from: snapshot uses the working copy and mainnet downloads the source from github. You just tell me what to do.

olemis: The source packages downloaded from github.com are built from tags . Repository check out could work for both cases , for instance , by having release tag name passed in as build / run argument , you'll get same files as if you had downloaded the source package .

gz-c: `https://github.com/skycoin/skycoin/archive/$branch_or_tag.zip`
`https://github.com/skycoin/skycoin/archive/$branch_or_tag.tar.gz`
`git clone https://github.com/skycoin/skycoin.git; git checkout $branch_or_tag`

These all work to get the source for `master`, `develop`, `v0.22.0` etc.  Don't need multiple methods of acquiring the source.  Downloading the zip or tar.gz should be faster and take less space.

olemis: 👍

stdevMauricio: OK, I'll unify them. As the repository will be on the build computer already, I'll just copy it to the compile container, checkout to the specified branch/tag/revision, if passed using `--build-args`, and build it. If nothing is specified, it will build the working copy.

gz-c: We can't assume the repo is on the build computer before calling docker, if they are fetching the image from docker hub

stdevMauricio: If you are fetching the image from docker hub, the image is already built. If you are going to build the image, you need the Dockerfile and it is in the repo. The only case I can see this failing, is somebody downloading just the Dockerfile and trying to build it alone. It seems to me a very edge case. Anyway just tell me what to do and I'll do it.

docker/images/mainnet/Dockerfile
```diff
@@ -1,50 +1,48 @@
...
 
 # volumes
 VOLUME $WALLET_DIR
 VOLUME $DATA_DIR
 
 EXPOSE 6000 6420 6430
 
-CMD ["/usr/bin/skycoin", "--gui-dir=/usr/local/skycoin/static", "--web-interface-addr=0.0.0.0"]
+ENTRYPOINT ["docker_launcher.sh"]
+CMD ["--web-interface-addr=0.0.0.0", "--rpc-interface-addr=0.0.0.0", "--gui-dir=/usr/local/skycoin/src/gui/static"]
```
iketheadore: Can't run with argument, node will crash when arguments are provided. 

Can put all in `ENTRYPOINT` like:
```dockerfile
ENTRYPOINT ["docker_launcher.sh", "--web-interface-addr=0.0.0.0", "--rpc-interface-addr=0.0.0.0", "--gui-dir=/usr/local/skycoin/src/gui/static"]
```

README.md
```diff
@@ -105,7 +105,9 @@ cd $GOPATH/src/github.com/skycoin/skycoin
 make ARGS="--launch-browser=false -data-dir=/custom/path" run
 ```
 
-### Docker image
+### Docker images
```
iketheadore: The anchor in the top of the README is `#docker-image`, which should be `#docker-images`.

docker/images/mainnet/Dockerfile
stdevMauricio: Done

README.md
stdevMauricio: Done

gz-c:
@iketheadore please check again

PR #1238, opened 2018-04-16
Wallet setup wizard

### LINKED ISSUE #1237: Wallet setup wizard
Add wallet setup wizard as in the web wallet.

Default wallet creation needs to be disabled on the backend.

mahansky:
Fixes #1237 

Changes:
- adds a wallet setup wizard

Yes

~~This PR requires additional work on backend to disable default wallet creation.~~ #1236

It depends on #1230 to be merged so encryption step can be enabled.


gz-c:
merge conflicts

src/gui/static/src/app/components/pages/onboarding/onboarding-create-wallet/onboarding-create-wallet.component.ts
```diff
@@ -0,0 +1,104 @@
...
+        });
+      }
+    });
+  }
+
+  loadWallet() {
+    this.button.setLoading();
+
+    this.walletService.create(this.form.value.label, this.form.value.seed, 100, null).subscribe(() => {
+      this.router.navigate(['/wallets']);
```
Senyoret1: When recovering a wallet, there is no screen for encrypting it.

Senyoret1:
In addition to the comments above, there are some details to consider:

- The disclaimer dialog says "mobile wallet", maybe the best thing is simply to put "wallet".
dialog

- When a wallet is recovered with the seed, instead of creating a new one, the encryption options are not shown (indicated in the comments of the code).

- With the addition of the encryption options, the `setLoading()` function of the buttons is being used in various places. This function shows the loading indicator, but does not prevent it from continue dispatching events if the user continues clicking. It would be advisable to prevent events from being dispatched if the loading indicator is being displayed, but I do not know if it should be done in this pr.

- The way in which the wizard was implemented is as follows: in the first form the wallet is created without encrypting and then, if the user enters a password in the second form, the wallet is encrypted. The time the wallet is unencrypted may have some security implications, so I think it is necessary to know what @gz-c thinks about it.

mahansky:
@Senyoret1 thanks for the review, I forgot to remove some stuff. I updated it based on your comments. Disclaimer was taken from skycoin-web (as most of the code), but it makes sense so I changed that as well. 

If I wanted to rework the wallet creation to create it in the final step, what would be the best way to send the parameters from the first step (label, seed)? Would be query parameters sufficient (in terms of security)?

Senyoret1:
Being honest, I don't think passing the parameters by query string is appropriate. Some options would be:

- Doing everything in a single step, just as in the "Create Wallet" popup.

- Saving the data in the `sessionStorage` and deleting it immediately after creating the wallet. The advantage of sessionStorage is that the data is not saved on the URL and is valid only for the current session and only while usign the current window and and the same tab. However, I think the previous option would still seem better, because `sessionStorage` could be store the data in persistent memory. More information in the first answer here: https://security.stackexchange.com/questions/36958/is-it-safe-to-store-password-in-html5-sessionstorage?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

- Using the second form as a simple control. This means that, after passing the first step, the second form would be displayed without changing the page. However, it is important to be careful with the operation of the browser's back button, sometimes it is simple and sometimes it is problematic.

gz-c:
Avoid saving anything sensitive in sessionStorage.  Maybe its best to do a single page setup here

Senyoret1:
I think the same, it's the simplest solution and that's why I put it first.

gz-c:
@mahansky Can you switch this to do the single view setup? You can drop the disclaimer too

Senyoret1:
Merge conflict.

Senyoret1:
Tested, it works well.

Now the wizard looks as to have 2 pages, but in reality the whole process is done in one and the wallet is created encrypted in a single step. The back button does not allow to return from the password page to the seed page, but it does not seem to be a very relevant problem. In my opinion, it would be worth fixing with `pushState` or `replaceState`, but only if it does not involve excessive work unnecessarily complicating the code.

In my opinion, it may be merged.

mahansky:
If it's really needed, I think that adding a "back" button would be the easiest way to do it.

Senyoret1:
I think that would be a good addition, so it would work with Electron as well. A simple button at the bottom, where the "Skip" button used to be, would be more than enough.

Senyoret1:
Sorry if this pr is getting a little more problematic than it should be, but I think there is only one more thing left. The "return" button works correctly if a new wallet is being created, but not if the user is loading an existing one.

gz-c:
@Senyoret1 is it ok now?

Senyoret1:
Yes, ready to merge.

Issue #735, opened 2017-12-10
Send peers to client before disconnecting due to full peers


gz-c:
When a node receives a connection and its connection limit is reached, send more peers to the connection before disconnecting it.

This message must come after the introduction message.

olemis:
@gz-c I sketched a solution to this issue in 0c3f3f7 . It's not tested yet . Nonetheless , could you please review ? I just want to be sure it's on the right track . Thanks in advance .

stdevStark:
@olemis can you check 4e47534 ?

olemis:
Travis failed

olemis:
Please fix [these errors](https://travis-ci.org/simelo/skycoin/jobs/414645614#L5822-L5938) .

olemis:
@gz-c will review #1773 eventually.

PR #1241, opened 2018-04-16
[test] [util] fixes #1110 fixes #913 - Annotated hexdump

### LINKED ISSUE #913: [examples] Automatically generate hex dump of gnet messages
Script input should be a set of message objects , which are meant to be serialized (binary) and printed as an hex dump , similar to the snippet below . Beware of fields excluded from serialization (e.g. message context) .

```
02000000 ........................... Block version: 2

b6ff0b1b1680a2862a30ca44d34
6d9e8
910d334beb48ca0c0000000000000000 ... Hash of previous block's header
9d10aa52ee949386ca9385695f04ede2
70dda20810decd12bc9b048aaab31471 ... Merkle root

24d95a54 ........................... Unix time: 1415239972
30c31b18 ........................... Target: 0x1bc330 * 256**(0x18-3)
fe9f0864 ........................... Nonce
```

### LINKED ISSUE #1110: [test] [util] Annotated hex dump
- [x] Render hex dump of binary buffer considering annotations
  * Helper function for hex dumps in `src/util` having three parametres
  * First parameter should be the object containing the data
  * Second one should be the annotations i.e. a list of `(index, name)` tuples where index is the buffer offset of the annotation and `name` its caption
  * Third parameter should be an instance of `io.Writer`
- [ ] Transform `MessageHexDump` function accordingly

stdevEdu:
Fixes #1110 

Changes:
- Dump P2P network messages in annotated hex format
- Testable examples for all `gnet` messages

no


src/daemon/message_hexdump.go
```diff
@@ -0,0 +1,255 @@
...
+	"github.com/skycoin/skycoin/src/daemon/gnet"
+
+	"reflect"
+
+	"github.com/skycoin/skycoin/src/util"
+)
+
+var registered = false
+
+func getSliceContentsString(sl []string, offset int) string {
```
olemis: @stdevEdu what is this function for ? Pls add one-line comment.

src/daemon/message_hexdump.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+	}
+	for i := 0; i < (16 - counter); i++ {
+		res += "..."
+	}
+	res += "..."
+	return res
+}
+
+func printLHexDumpWithFormat(offset int, name string, buffer []byte) {
```
olemis: @stdevEdu what is this function for ? Pls add one-line comment.

src/daemon/message_hexdump.go
```diff
@@ -0,0 +1,255 @@
...
+	for i := 0; i < len(buffer); i++ {
+		if len(hexBuff[i]) == 1 {
+			hexBuff[i] = "0" + hexBuff[i]
+		}
+	}
+	fmt.Println(getSliceContentsString(hexBuff, offset), name)
+}
+
+// NonGenericHexDump : version of hexdump without Annotations
+func NonGenericHexDump(message gnet.Message) string {
```
olemis: like this 👍

olemis: Please remove this approach ... let's only keep the one relying on iterators .

src/daemon/message_hexdump.go
stdevEdu: This method was also a helper fot NonGenericHexDump, so it was removed too.

src/daemon/message_hexdump.go
stdevEdu: Done. Removed both the non-generic implementation and the IAnnotationsGeneration implementation for gnet.Message

olemis:
needed for #735

gz-c:
@olemis how does this relate to #735?

olemis:
@gz-c I'll need this for the tests, not the feature . Better to see that in the patch . Review will be appreciated . Thanks !

PR #1243, opened 2018-04-16
Do gui static file path resolving only when wallet api is enabled

### LINKED ISSUE #1242: Remove the dependence of gui static when wallet api is disabled
Node should be able to run normally without the gui static files when wallet api is disabled.

iketheadore:
Fixes #1242

Changes:
- Only when node wallet api is enabled, should we do the gui static dir path resolving.

No.

### DIFF
--- a/cmd/skycoin/skycoin.go
+++ b/cmd/skycoin/skycoin.go
@@ -383,8 +383,9 @@ func (c *Config) postProcess() {
 
 	// Don't open browser to load wallets if wallet apis are disabled.
 	if c.EnableWalletAPI {
-		c.LaunchBrowser = true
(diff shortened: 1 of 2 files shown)


cmd/skycoin/skycoin.go
```diff
@@ -382,8 +382,9 @@ func (c *Config) postProcess() {
 	}
 
 	// Don't open browser to load wallets if wallet apis are disabled.
-	if !c.EnableWalletAPI {
-		c.LaunchBrowser = false
+	if c.EnableWalletAPI {
+		c.LaunchBrowser = true
```
gz-c: This forces the browser launch if wallet API is enabled

PR #1261, opened 2018-04-17
[util] Fixes #1235 - Log http error messages

### LINKED ISSUE #1235: Log http response error messages
Modify either the elapsed handler logger to add the response body when the status code is not 200, or inside the `Error400` etc methods, accept a logger argument and log the status and error message

stdevHan:
Fixes #1235 

Changes:
-Define for each error of http the message that corresponds to it, in the log


src/util/http/log.go
```diff
@@ -13,7 +13,30 @@ func ElapsedHandler(logger logrus.FieldLogger, handler http.Handler) http.Handle
 		lrw := newWrappedResponseWriter(w)
 		start := time.Now()
 		handler.ServeHTTP(lrw, r)
-		logger.Infof("%v %s %s %v", lrw.statusCode, r.Method, r.URL.Path, time.Since(start))
+		if lrw.statusCode == 200 || lrw.statusCode == 304 {
```
olemis: Condition should be `lrw.statusCode < 400`

src/util/http/log.go
```diff
@@ -13,7 +13,30 @@ func ElapsedHandler(logger logrus.FieldLogger, handler http.Handler) http.Handle
...
+			logger.Infof("%v %s %s %v", lrw.statusCode, r.Method, r.URL.Path, time.Since(start))
+		} else {
+			if lrw.statusCode == 400 {
```
olemis: A single statement here , please

src/util/http/error.go
```diff
@@ -36,7 +42,8 @@ func Error403Msg(w http.ResponseWriter, msg string) {
 }
 
 // Error404 respond with a 404 error
-func Error404(w http.ResponseWriter) {
+func Error404(w http.ResponseWriter, r *http.Request) {
+	logger.Errorf("%s %s", "404 Not Found", r.URL.Path)
```
gz-c: Log for the methods that pass in an extra message

PR #1277, opened 2018-04-18
Prevent closing modals when working

### LINKED ISSUE #1265: Modal should not be dismissable while operation in progress
Create/load wallet and password modals can be dismissed while the decrypt/encrypt operation is in progress. The modal should not be dismissable while it is working

mahansky:
Fixes #1265 

Changes:
- user won't be able to close modal when it's working (password dialog, create/load wallet)


Senyoret1:
This pr is going to have a lot of conflicts with https://github.com/skycoin/skycoin/pull/1238, so I think it would be better to review it after mixing that pr.

gz-c:
@Senyoret1 #1238 is merged

merge conflicts

PR #1278, opened 2018-04-18
Make info about recovering seeds more visible

### LINKED ISSUE #1273: Improve Backup Wallets page text
There is a warning about encrypted wallets in a question mark bubble.

Most of this text should be in the main text, not hidden behind the question mark.

Explain that only encrypted wallets can access their seed through this page.  If they want to recover their seed from an unencrypted wallet, they should open the wallets folder on their system, open the `.wlt` file(s) in a text editor and recover the seed.

mahansky:
Fixes #1273

Changes:
- info about recovering seeds moved from tooltip to main text


Senyoret1:
i can't proofread it, but looks good to me.

text

PR #1280, opened 2018-04-18
Prevent browser to store passwords from inputs

### LINKED ISSUE #1249: Prevent browser from suggesting to save wallet password
After encrypting a wallet, chrome suggests to save the password with no username.  This should be disabled if possible.

This may solve it:
https://stackoverflow.com/questions/32369/disable-browser-save-password-functionality/37292424#37292424

[This issue does not apply until #1230 is merged]

mahansky:
Fixes #1249

Changes:
- added `autocomplete="new-password"` to password inputs so browser won't ask to store it

#1238 has additional password inputs that will need to be added here after it's merged

### DIFF
--- a/src/gui/static/src/app/app.module.ts
+++ b/src/gui/static/src/app/app.module.ts
@@ -59,6 +59,7 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { HttpClientModule } from '@angular/common/http';
 import { AppService } from './services/app.service';
 import { SeedModalComponent } from './components/pages/settings/backup/seed-modal/seed-modal.component';
+import { DontsavepasswordDirective } from './directives/dontsavepassword.directive';
 
 
 const ROUTES = [
@@ -141,6 +142,7 @@ const ROUTES = [
     ModalComponent,
     PasswordDialogComponent,
     SeedModalComponent,
+    DontsavepasswordDirective,
   ],
   entryComponents: [
     AddDepositAddressComponent,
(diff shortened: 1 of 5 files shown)


Senyoret1:
I did some tests and the problem is still present in the configuration I used, but it seems that there is no simple solution. Both before and after this pr, the problem is not present in the following configuration:

- Windows / Microsoft Edge 38
- Windows / Firefox 59

However, in the following configuration the problem is present both before and after this pr:

- Windows / Google Chrome 66

The tricky thing is that there is no standard/multiplatform way to solve the problem and Chrome tries to be "smart" when it comes to detecting passwords, in a way that makes it very difficult to avoid the problem. In fact, the "tricks" that can be found on the Internet have stopped working with each new update of Chrome. However, it may be a good idea to use this tool if it works, since it is not very complicated and can be useful for some users.

gz-c:
If it's too difficult and there is not a standard solution, a simple solution that works sometimes is ok.  This only affects people running it in the browser (not electron), and its mostly a nuisance.

Senyoret1:
I made a mistake in my last comment, what I meant to say was: "However, it may be a good idea to use this pr if it works for @mahansky , since it is not very complicated and may be useful for some users.". So, if a simple solution that works sometimes is ok then, in my opinion, this pr can be merged as it is.

mahansky:
I could combine it with solution provided by @gz-c in the issue to maximize the chance it won't show the password popup.

Senyoret1:
With this new commit Chrome no longer tries to save the password and everything works correctly in the 3 browsers I used to test. Very good improvement, in fact.

PR #1290, opened 2018-04-19
Return 401 for invalid password

### LINKED ISSUE #1267: Invalid or missing password err should return HTTP 401
Missing password can stay as 400, if that makes more sense.

iketheadore:
Fixes #1267 

Changes:
- Return 401 Unauthorized for invalid password
- Add integration-test-disable-wallet-api and integration-test-disable-seed-api to make check

Yes.


gz-c:
Not doing this since we don't have anything to put in the `WWW-Authenticate` header

gz-c:
@iketheadore can we make up our value for `WWW-Authenticate: $thing`? If so, then its ok, we can add that

PR #1296, opened 2018-04-19
Fixes #1093 Speed up make test-libc command

### LINKED ISSUE #1093: Speed up make test-libc command
While working on libskycoin .c test files, library binaries get rebuilt even if there are no changes in source files (i.e `lib/cgo/*.go`).

stdevHan:
Fixes #1093 

Changes:
-Modify the Makefile file so that it only compiles when you find changes in the /lib/cgo

No


olemis:
#1093 will be fixed in #1395 .

PR #1289, opened 2018-04-19
Use POST for /wallet/seed; run more wallet integration tests by default

gz-c:
Fixes #1282 

Changes:
- Change `/wallet/seed` to use `POST`
- Stable tests run against wallet APIs when possible
- `--test-wallet` flag only applies for tests that require a real wallet

Does this change need to mentioned in CHANGELOG.md?
No

PR #1301, opened 2018-04-20
Require password confirmation when encrypting wallet

### LINKED ISSUE #1299: (Wallet) A field for confirming the password is needed.
If the user has an unencrypted wallet and press the "Encrypt Wallet" button, a popup appears to allow to enter the password, but it only has one field, so the user could make an error.
pass_dialog

It would be good to have an additional field to confirm the password, as in other screens.

mahansky:
Fixes #1299

Changes:
- when encrypting wallet, you are required to enter two passwords (for confirmation)

screen shot 2018-04-20 at 13 27 28

### DIFF
--- a/src/gui/static/src/app/components/layout/password-dialog/password-dialog.component.html
+++ b/src/gui/static/src/app/components/layout/password-dialog/password-dialog.component.html
@@ -5,9 +5,13 @@
       <input formControlName="password" id="password" type="password"
              placeholder="Wallet Password" appDontSavePassword>
     </div>
+    <div class="form-field" *ngIf="data.confirm">
+      <label for="password">Confirm password</label>
+      <input formControlName="confirm_password" id="confirm_password" type="password" appDontSavePassword>
+    </div>
   </div>
   <div class="-buttons">
-    <app-button #button (action)="proceed()" class="primary">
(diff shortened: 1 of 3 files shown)


Senyoret1:
With the changes, the dialog stopped working on the backup page (because `data ` is `null`). Also, thinking about it, it might be good to remove the "wallet password" hint, because it is somewhat redundant and none of the fields in the other dialogs have hints, so it is also a bit inconsistent.

mahansky:
Good points, I added more parameters to configure password dialog. It looks like this now:
screen shot 2018-04-20 at 20 33 19

Senyoret1:
Great. Tested and no more complaints from me.

gz-c:
Merge conflict;

Is the password data cleared from the DOM when using this new modal?

Senyoret1:
It does it in here every time the dialog is removed from the screen: https://github.com/mahansky/skycoin/blob/cb5aa952ea82e2beba3f5b69b5b47a5db4c05896/src/gui/static/src/app/components/layout/password-dialog/password-dialog.component.ts#L65-L68

gz-c:
merge conflict

Senyoret1:
Getting this while trying to recover a seed:

dec_error

mahansky:
#1289 changed the endpoint from GET to POST, updated

PR #1293, opened 2018-04-19
Show different errors when decrypting seed

mahansky:
Fixes #1276

Changes:
- show different errors based on response status code
screen shot 2018-04-19 at 12 34 59
screen shot 2018-04-19 at 12 33 25

PR #1302, opened 2018-04-20
Show error on wallet creation

### LINKED ISSUE #1298: (Wallet) Create 2 wallets with the same seed
Surely this is super low priority, but if the user creates a wallet with a seed and then tries to create or load a second wallet with the same seed, the wallet is not created and the create button shows the loading animation indefinitely , without displaying an error message.

mahansky:
Fixes #1298

Changes:
- if the wallet creation fails, show the error in "Create" button


Senyoret1:
Will review after https://github.com/skycoin/skycoin/pull/1238

src/gui/static/src/app/components/pages/wallets/create-wallet/create-wallet.component.ts
```diff
@@ -32,11 +32,12 @@ export class CreateWalletComponent implements OnInit {
   }
 
   createWallet() {
+    this.button.resetState();
     this.button.setLoading();
 
     const password = this.encrypt ? this.form.value.password : null;
     this.walletService.create(this.form.value.label, this.form.value.seed, this.scan, password)
-      .subscribe(() => this.dialogRef.close());
+      .subscribe(() => this.dialogRef.close(), e => this.button.setError(e));
```
Senyoret1: Would be better if the message is cleaned with `parseResponseMessage` from src/app/utils/index.ts, added in https://github.com/skycoin/skycoin/pull/1293

mahansky: That should happen inside the setError method 
https://github.com/skycoin/skycoin/blob/76a82204b6350a784194976f47673f9e38fc7c47/src/gui/static/src/app/components/layout/button/button.component.ts#L34

src/gui/static/src/app/components/pages/wallets/create-wallet/create-wallet.component.ts
```diff
@@ -34,13 +34,14 @@ export class CreateWalletComponent implements OnInit {
...
   createWallet() {
+    this.createButton.resetState();
     this.createButton.setLoading();
     this.cancelButton.setDisabled();
     this.disableDismiss = true;
 
     const password = this.encrypt ? this.form.value.password : null;
     this.walletService.create(this.form.value.label, this.form.value.seed, this.scan, password)
-      .subscribe(() => this.dialogRef.close());
+      .subscribe(() => this.dialogRef.close(), e => this.createButton.setError(e));
```
Senyoret1: It is necessary to add the following lines in the error handler, for making possible to close the dialog after an error:
```
this.cancelButton.disabled = false;
this.disableDismiss = false;
```

src/gui/static/src/app/components/pages/wallets/create-wallet/create-wallet.component.ts
Senyoret1: You're right, I did not notice it because the pr did not include that change.

src/gui/static/src/app/components/pages/wallets/create-wallet/create-wallet.component.ts
gz-c: @Senyoret1 can you add it? then we can merge it

PR #1304, opened 2018-04-20
User's pending transactions

### LINKED ISSUE #1297: Pending transactions warning
There is a big warning saying "There are some pending transactions. Data you see may not be updated". Currently, it's shown whenever there are some pending transactions. It might be visible all the time as the network will be used more often. 

We could show the warning only if one of the user's addresses is listed as an output of one of the pending transactions.

mahansky:
Fixes #1297

Changes:
- red warning saying "There are some pending transactions.." will now be shown only if your addresses are part of those transactions
- using `/transactions?confirmed=0&addrs=<user's adresses>` instead of `/pendingTxs`

`Settings -> Pending transactions` are still using the default `/pendingTxs` endpoint

### DIFF
--- a/src/gui/static/src/app/services/wallet.service.ts
+++ b/src/gui/static/src/app/services/wallet.service.ts
@@ -32,12 +32,7 @@ export class WalletService {
   }
 
   addressesAsString(): Observable<string> {
-    return this.all().map(wallets => wallets.map(wallet => {
-      return wallet.addresses.reduce((a, b) => {
-        a.push(b.address);
-        return a;
-      }, []).join(',');
-    }).join(','));
+    return this.allAddresses().map(addrs => addrs.map(addr => addr.address)).map(addrs => addrs.join(','));
   }
 
   addAddress(wallet: Wallet, password?: string) {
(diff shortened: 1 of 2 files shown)


Senyoret1:
Tested with real and simulated unconfirmed transactions (modifying the node, not the frontend). Working without problems.

mahansky:
Switched form using `/transactions?confirmed=0` to `/wallet/transactions`

src/gui/static/src/app/services/wallet.service.ts
```diff
@@ -88,13 +83,30 @@ export class WalletService {
...
-      this.pendingTxs.next(txs);
+    this.wallets.first().subscribe(wallets => {
+      Observable.forkJoin(wallets.map(wallet => this.apiService.get('wallet/transactions', { id: wallet.filename })))
+        .subscribe(pending => {
+          this.pendingTxs.next(
+            pending
+              .filter(response => response.transactions.length > 0)
+              .map(response => response.transactions)
+              .reduce((txs, tx) => {
+                if (!txs.find(t => t.transaction.txid === tx.transaction.txid)) {
```
Senyoret1: Throwing "ERROR TypeError: Cannot read property 'txid' of undefined" both with real transactions and with simulated responses.

For initial testing, you can force the node to always return one transaction per wallet by adding this code on the line 520 of src/gui/wallet.go:

```
testTx := visor.UnconfirmedTxn {}
txns = make([]visor.UnconfirmedTxn, 1)
txns[0] = testTx;
```

mahansky:
I tested it with real transactions and it worked for some reason, but shouldn't. Thanks for the spot.

PR #1308, opened 2018-04-20
[cmd] [gui]Fixed #1275 Randomize web interface port

### LINKED ISSUE #1275: Randomize web interface port when running as desktop wallet
If an attacker has access to the machine, they could bind to the standard port while the backend node is off, and in some cases, this could result in the wallet communicating to a malicious backend.  Randomize the web interface port when running as a desktop wallet to reduce the feasibility of this attack.

stdevHan:
Fixes #1275 

Changes:
-Modifying cmd/skycoin.go so that it generates a random port to the api
-Modifying /src/gui/static/src/environments/environment.prod.ts to read through any port, the api

No

### DIFF
--- a/src/gui/static/src/environments/environment.prod.ts
+++ b/src/gui/static/src/environments/environment.prod.ts
@@ -1,5 +1,5 @@
 export const environment = {
-  nodeUrl: 'http://127.0.0.1:6420/',
+  nodeUrl: '../' ,
   production: true,
   tellerUrl: 'https://event.skycoin.net/api/',
 };
--- a/cmd/skycoin/skycoin.go
+++ b/cmd/skycoin/skycoin.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"math/rand"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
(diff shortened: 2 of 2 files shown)


gz-c:
Port randomization should only occur when run via `run.sh` or electron.  The runner should choose their own randomized port and configure the skycoin node's options with this port. Do not change the internal port.

run.sh
```diff
@@ -15,6 +14,7 @@ go run -ldflags "${GOLDFLAGS}" cmd/skycoin/skycoin.go \
     -launch-browser=true \
     -enable-wallet-api=true \
     -rpc-interface=false \
+    -web-interface-port $(($RANDOM%65563))  \
```
gz-c: * Ports below 1000 can only be bound by root
* Port should make sure nothing else is using it
* Should use a CSPRNG-generated number if possible
* Need to do this for electron as well

src/gui/static/src/environments/environment.prod.ts
```diff
@@ -1,5 +1,5 @@
 export const environment = {
-  nodeUrl: 'http://127.0.0.1:6420/',
+  nodeUrl: '../' ,
```
olemis: Does this always work ? Shouldn't it be an absolute URL ?

stdevHan: @olemis  absolute urls are set when you are only going to point to a standard address, but in this case they change the addresses a lot, that is, ip: random (port) is declared in this way so that it reads the api from the address that continues ".. / ".

PR #1341, opened 2018-04-22
Cli with encryption

### LINKED ISSUE #953: Update CLI to support wallet encryption

### LINKED ISSUE #1171: Update CLI to support encrypted wallets
Wallet CLI password handling:

Password arguments can be specified with a cli flag.  
If a password is provided and the wallet is not encrypted, this is an error.
If a password is not provided and the wallet is encrypted, prompt the user for a password.

Prompting for a password:

https://godoc.org/golang.org/x/crypto/ssh/terminal#ReadPassword
    - Make sure this does not need cgo to build (add CGO_ENABLED=0 to makefile test commands?)
    - If it does, maybe try https://github.com/bgentry/speakeasy 

Wallet CLI changes:

`addPrivateKey`
    - Add password argument

`createRawTransaction`
    - Add password argument

`generateAddresses`
    - Add password argument

`generateWallet`
    - Add encrypt (bool) and password arguments

`send`
    - Add password argument

New command:

`showSeed`

Args:
    - wallet file from env vars
    - password, if encrypted

Print the seed for the wallet on the command line.  As opposed to `GET /wallet/seed`, this doesn't care if the wallet is encrypted or not.

New command:

`decryptWallet`

Args:
    - wallet file from env vars
    - password

Return an error if the wallet is not encrypted (exit code 1), otherwise if successful return nothing (exit code 0)

New command:

`encryptWallet`

Args:
    - wallet file from env vars
    - password

Return an error if the wallet is already encrypted.

iketheadore:
Fixes #1171 

Changes:
- Add password argument to `addPrivateKey`, `createRawTransaction`, `generateAddresses`, `generateWallet`, `send`.
- Add `showSeed` command with encryption support
- Add `decryptWallet` command
- Add `decryptWallet` command
- Update partial integration tests

TODO:
- [x] Add CLI integration tests for `showSeed`
- [x] Add CLI integration tests for `decryptWallet`
- [x] Fill CLI integration tests for `encryptWallet`
- [x] Update CHANGELOG.md

Yes

### DIFF
--- a/src/api/cli/add_private_key.go
+++ b/src/api/cli/add_private_key.go
@@ -63,7 +63,7 @@ func addPrivateKeyCmd(cfg Config) gcli.Command {
 				errorWithHelp(c, err)
 				return nil
 			case WalletSaveError:
-				return errors.New("Save wallet failed")
+				return errors.New("save wallet failed")
 			default:
 				return err
 			}
--- a/src/api/cli/add_private_key.go
+++ b/src/api/cli/add_private_key.go
@@ -51,7 +51,7 @@ func addPrivateKeyCmd(cfg Config) gcli.Command {
 				return err
 			}
 
-			err = AddPrivateKeyToFile(w, skStr, c)
+			err = AddPrivateKeyToFile(w, skStr, []byte(c.String("p")))
 
 			switch err.(type) {
 			case nil:
@@ -97,14 +97,12 @@ func AddPrivateKey(wlt *wallet.Wallet, key string) error {
 }
 
 // AddPrivateKeyToFile adds a private key to a wallet based on filename.  Will save the wallet after modifying.
-func AddPrivateKeyToFile(walletFile, key string, c *gcli.Context) error {
+func AddPrivateKeyToFile(walletFile, key string, password []byte) error {
 	wlt, err := wallet.Load(walletFile)
 	if err != nil {
 		return WalletLoadError(err)
 	}
 
-	// read password from -p flag
-	password := []byte(c.String("p"))
 	if !wlt.IsEncrypted() {
 		if len(password) != 0 {
 			return wallet.ErrWalletNotEncrypted
@@ -116,7 +114,7 @@ func AddPrivateKeyToFile(walletFile, key string, c *gcli.Context) error {
 	} else {
 		if len(password) == 0 {
 			var err error
-			password, err = readPasswordFromTerminal(c)
+			password, err = readPasswordFromTerminal()
 			if err != nil {
 				return err
 			}
--- a/src/api/cli/add_private_key.go
+++ b/src/api/cli/add_private_key.go
@@ -103,7 +103,11 @@ func AddPrivateKeyToFile(walletFile, key string, pr PasswordReader) error {
 			return wallet.ErrMissingPassword
 		}
 	case PasswordFromBytes:
-		p, _ := pr.Password()
+		p, err := pr.Password()
+		if err != nil {
+			return err
+		}
+
 		if !wlt.IsEncrypted() && len(p) != 0 {
 			return wallet.ErrWalletNotEncrypted
 		}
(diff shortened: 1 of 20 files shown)


src/api/cli/cli.go
```diff
@@ -305,3 +309,72 @@ func printJSON(obj interface{}) error {
 
 	return nil
 }
+
+// readPasswordFromTerminal promotes user to enter password and read it.
+func readPasswordFromTerminal() ([]byte, error) {
+	// Promotes to enter the wallet password
+	fmt.Fprint(os.Stdout, "enter password:")
+	bp, err := terminal.ReadPassword(0)
```
gz-c: should we use os.Stdin?

src/api/cli/cli.go
```diff
@@ -305,3 +309,72 @@ func printJSON(obj interface{}) error {
...
+func (w WalletLoadError) Error() string {
+	return w.error.Error()
+}
+
+// WalletSaveError is returned if a wallet could not be saved
+type WalletSaveError struct {
+	error
+}
+
+func (w WalletSaveError) Error() string {
```
gz-c: isn't this redundant?

src/api/cli/generate_wallet.go
```diff
@@ -120,11 +131,49 @@ func generateWallet(c *gcli.Context) error {
...
-	wlt, err := GenerateWallet(wltName, label, sd, num)
+
+	cryptoType, err := wallet.CryptoTypeFromString(c.String("x"))
+	if err != nil {
+		return err
+	}
+
+	pr := NewPasswordReader([]byte(c.String("p")))
+	switch pr.(type) {
+	case nil:
```
gz-c: pr won't be nil?

src/api/cli/add_private_key.go
```diff
@@ -90,13 +91,42 @@ func AddPrivateKey(wlt *wallet.Wallet, key string) error {
...
+		return WalletLoadError{err}
+	}
+
+	switch pr.(type) {
+	case nil:
+		if wlt.IsEncrypted() {
+			return wallet.ErrMissingPassword
+		}
+	case PasswordFromBytes:
+		p, _ := pr.Password()
```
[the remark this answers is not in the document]
iketheadore: `PasswordFromBytes` type doesn't have error.

src/api/cli/cli.go
iketheadore: `os.Stdin` is type of `*os.File`, while the `ReadPassword` accepts `int` type as argument. I tried to use `syscall.Stdin`, but it has compilation problem, would have error: 

```
cannot use syscall.Stdin (type syscall.Handle) as type int in argument to terminal.ReadPassword
```

gz-c: Nevermind, its safe to use 0, syscall.Stdin is always 0 anyway

iketheadore: Yes, syscall.Stdin is 0, but not for windows, as the `build` flag in `syscall_unix.go` is
`// +build darwin dragonfly freebsd linux netbsd openbsd solaris`.

gz-c: What is it in windows? Then do we need to do `int(syscall.Stdin)`?

gz-c: I see, `syscall_windows.go:	Stdin  = getStdHandle(STD_INPUT_HANDLE)`

then we probably need to do `int(os.Stdin.Fd())` or `int(syscall.Stdin)`

src/api/cli/add_private_key.go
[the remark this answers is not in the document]
gz-c: should still check in case that changes

src/api/cli/cli.go
iketheadore: I will test `int(syscall.Stdin)` on windows to see if it is able to work

iketheadore: OK,` int(syscall.Stdin)` works, I will use it.

Issue #1342, opened 2018-04-23
Reject the empty address in /injectTransaction, /wallet/spend


gz-c:
A bug in skycoin-web created a transaction sending coins to the empty address,

https://github.com/skycoin/skycoin-web/issues/60

`/injectTransaction` and possibly other APIs (such as `/wallet/spend`) should detect empty addresses and reject requests with them.

`POST /wallet/transaction` already detects this and rejects them.

gz-c:
This could possibly be a soft constraint, instead of only at the API layer

gz-c:
Verification needs to be structured as follows:

- Hard constraint
    Used by transactions in blocks
- Soft + Hard constraint
    Used by transactions received from other users
- User + Soft + Hard constraint
    Used by user-created/injected transactions

Currently we only have Hard and Soft constraints.

This null address check only matters for User constraints.  It should not be treated as a soft constraint, because soft constraint-violating transactions are still retained in the pool under the assumption that they may become valid in the future (and are then propagated).  

A transaction which sends to the null address is valid and not subject to changing rules.  It is only important to prevent the user from creating or injecting such a transaction to avoid client mistakes.

gz-c:
`/injectTransaction` will check this, but does so it in its own handlers, and not in a common verification layer.

Issue #1320, opened 2018-04-21
Logging configuration


gz-c:
Allow log levels to be configurable per module

gz-c:
The ability to disable all logging would be useful.  This should be done first.

stdevStark:
@olemis I can not find why it is not working . Log levels are parsed correctly and used to set the log configuration structs . Anyway does not work .

olemis:
In my opinion this is happening because package loggers are instantiated prior to configuring package-specific log level. It's necessary to keep track of the loggers created for each package so that they can be configured correctly once logging options are parsed via command-line args .

PR #1349, opened 2018-04-23
Created txn verification; fix hours/calculated hours problem in rendering created txn

gz-c:
Changes:
- In `/injectTransaction`, check that the transaction does not spend to the null address.  This catches a possible mistake with transaction generation by other systems (e.g `POST /wallet/transaction`, web wallet, mobile wallet).  In the future, a full fix for #1342 will be added which will handle the verification in the visor layer.
* Fix a bug where the created transaction could not be rendered.  This was due to a validation check using original coin.UxOut hours instead of the calculated hours.  Return and render both the original hours and calculated hours.

Yes


src/wallet/wallet.go
```diff
@@ -1257,18 +1260,121 @@ func (w *Wallet) CreateAndSignTransactionAdvanced(params CreateTransactionParams
...
 		if !ok {
-			return nil, nil, errors.New("Created transaction's input is not in the UxHashSet, this should not occur")
+			return nil, nil, errors.New("Created transaction's input is not in the UxBalanceSet, this should not occur")
 		}
-		inputs[i] = uxOut
+		inputs[i] = uxBalance
+	}
+
+	if err := verifyCreatedTransactionInvariants(params, txn, inputs); err != nil {
+		logger.Criticalf("CreateAndSignTransactionAdvanced created transaction that violates invariants, aborting: %v", err)
```
iketheadore: In genera we should not print error if the error will be returned, as if one day we need to call this function in CLI, the error will be print out unexpectedly.

gz-c: I would have it panic, but don't want to crash node in case this has problem.  This should never print though

gz-c: For that problem with CLI, if we implement logging configuration (#1320) we can disable

PR #1350, opened 2018-04-23
Validate amount in Send form

### LINKED ISSUE #1336: Wallet send page, coins validation issues
The send page does some checking that the coins value entered is not greater than the wallet balance. If the value is not ok, it makes the button grey.

Some inconsistencies with the validation:

1. A string with a very high number of decimals is permitted (likely due to the smallest values being truncated/rounded off).

### Example

`10.0000000000000001`

In this case, the wallet will send 10 coins.

### Solution

The number of digits to the right of an optional decimal place must be no higher than 6. Check this string length.

2. Zero is allowed.

### Example

`0`
`0.000`

3. Nonsense values are allowed

Example:

`asdasdwda`
`.....`
`0..1`

Number 1 is the biggest problem since the frontend transforms the value, changing the user's input unexpectedly. The spend would succeed.

Number 2 and 3 would be caught by the backend, but if the client is validating the value, they should check these cases too.

Note: the client should *not* check the artificial decimal restriction (to 3 decimal places out of 6) because this is something we'll change.  The backend will catch it.

mahansky:
Fixes #1336

Changes:
- validates amount in send form to prevent values like 'asd', '0', '10.0000000001' being submitted

### DIFF
--- a/src/gui/static/src/app/components/pages/send-skycoin/send-skycoin.component.ts
+++ b/src/gui/static/src/app/components/pages/send-skycoin/send-skycoin.component.ts
@@ -1,6 +1,6 @@
 import { Component, OnDestroy, OnInit, ViewChild } from '@angular/core';
 import { WalletService } from '../../../services/wallet.service';
-import { FormBuilder, FormGroup, Validators } from '@angular/forms';
+import { FormBuilder, FormControl, FormGroup, Validators } from '@angular/forms';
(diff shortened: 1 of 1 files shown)


Senyoret1:
Somehow, after sending the coins the form is cleaned correctly, but the send button remains showing the loading animation. This does not happen when I use the develop branch.
busy

PR #1365, opened 2018-04-24
Encryption option enabled by default in wallet setup wizard

### LINKED ISSUE #1361: Encrypt wallet option should be checked by default in wallet setup wizard

gz-c:
Fixes #1361


gz-c:
#1366

PR #1370, opened 2018-04-24
Force refresh pending txs after send

### LINKED ISSUE #1357: Wallet should check for unconfirmed pending txns immediately after spend
The pending txns banner can take a delay to show up because it checks for pending txns on an interval. After a successful spend we can assume there will be pending txn so we should make an extra request at that point to get the banner displayed.

mahansky:
Fixes #1357

Changes:
- after creating transaction in "Send", pending txs are refreshed so the banner is displayed asap
- same thing happens after opening "Pending transactions"


Senyoret1:
It is a minor problem, but if you send a coin and quickly enter to the pending transactions page, the transaction appears in the list and does not disappear even after the alert message is gone, the transaction stays there until the user leaves the page and returns to it.

mahansky:
I temporarily fixed it by showing only user's pending txs (refreshed together with banner warning) and created an issue to add button to toggle between all and user's txs #1375

Issue #1354, opened 2018-04-23
(Wallet) Clean the "Notes" field


Senyoret1:
After sending coins, all fields in the form are cleaned, except the "Notes" field.

PR #1373, opened 2018-04-24
UI fixes

### LINKED ISSUE #1351: Improved usability in button errors
If a button is pressed and the operation fails, it is possible to see the error message by placing the mouse pointer over the button.
error msg
However, if the button is pressed and the operation fails while the mouse pointer is over the button, the error message is not displayed, it is necessary to move the pointer away and put it back on the top of the button. Ideally, the error message should be displayed without the need to do that.

mahansky:
Fixes #1351 #1354

Changes:
- "notes" field is cleared after submitting Send form
- tooltip on button is shown even without moving mouse off the button and over it again

### DIFF
--- a/src/gui/static/src/app/components/layout/button/button.component.html
+++ b/src/gui/static/src/app/components/layout/button/button.component.html
@@ -1,5 +1,16 @@
-<button mat-button type="submit" mat-raised-button color="primary" [disabled]="disabled && !forceEmitEvents"
-        [matTooltip]="error ? error : null" (click)="this.state !== 0 ? onClick() : null" [ngClass]="{enabled: !disabled}">
+<button
+  mat-button
+  type="submit"
+  mat-raised-button
+  color="primary"
+  [disabled]="disabled && !forceEmitEvents"
+  [matTooltip]="error ? error : null"
+  (click)="this.state !== 0 ? onClick() : null"
+  [ngClass]="{ enabled: !disabled }"
+  (mouseenter)="mouseOver = true"
+  (mouseleave)="mouseOver = false"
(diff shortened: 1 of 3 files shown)


Senyoret1:
In the tests I did the last two commits worked without problems. The first one worked erratically (it seems that the message is shown only if the pointer is on certain parts of the button), but it seems to work well in all cases if the error message is shown usign this code: `setTimeout(() => this.tooltip.show(), 32);`.

Senyoret1:
I'm not entirely sure why it does not work for me. I put 32 as delay because it is about 2 frames at 60fps, I ususually use that number while testing race conditions. Anyway. I tested with a 500ms delay and the problem continues. Also tried `npm update` before, just in case.

To understand a little more that is happening I used this code:
```
setError(error: any) {
  this.error = null;
  this.state = 2;
  setTimeout(() => this.error = "msg", 200);
  this.tooltip.show(1000);
}
```
Nomally, if `this.error` is null, the tooltip is not displayed. In this case it should not matter, because a value is set for `this.error` after 200ms and the tooltip should appear only after 1000ms, but the tooltip does not appear, no matter where the mouse is. However, if the last line is replaced with `setTimeout(() => this.tooltip.show(), 1000);` it works. That makes me think that any of this lines is canceling the operation as soon as the null tooltip text is detected: https://github.com/angular/material2/blob/2046b7a3287bbff87a48800cf7d770f1976ed757/src/lib/tooltip/tooltip.ts#L549-L556 . However, I do not know why it works differently in your tests and in mine and I think that checking exactly what happens inside material2 would take a good amount of time.

PR #1374, opened 2018-04-24
Restore log functionality

### LINKED ISSUE #1319: Remove -logbufsize and -logtogui options, and related dead code
Remove those options, as well as related dead code in cmd/skycoin/skycoin.go and in the gui/ package http handlers.

### LINKED ISSUE #1372: Logs are not saved by the client

gz-c:
Fixes #1372 
Fixes #1319 

Changes:
- Refactor logging module into a single master logger that provides a `Critical()` method to tag lines as critical and works with subsequent chained `WithField` and `WithFields` calls
- Add support for multi-destination writing with `WriteHook`; restore `-logtofile` functionality
- Restore `-color-log` option
- Restore `-log-level` option
- Remove unused `-logtofile`, `-logbufsize` options and related code

Yes

### DIFF
--- a/src/util/logging/logger.go
+++ b/src/util/logging/logger.go
@@ -1,170 +1,73 @@
 package logging
 
 import (
-	"io/ioutil"
 	"os"
 
 	"github.com/sirupsen/logrus"
 )
 
-// ExtendedFieldLogger is an enhanced logger supporting critical and important levels
-type ExtendedFieldLogger interface {
-	logrus.FieldLogger
-
-	// Critical level
-	Criticalf(format string, args ...interface{})
-	Critical(args ...interface{})
-	Criticalln(args ...interface{})
-
-	// Notice level
-	Noticef(format string, args ...interface{})
-	Notice(args ...interface{})
-	Noticeln(args ...interface{})
-}
-
-// Logger wraps sirupsen/logrus.Logger to implement ExtendendFieldLogger
-type Logger struct {
+// MasterLogger wraps logrus.Logger and is able to create new package-aware loggers
+type MasterLogger struct {
 	*logrus.Logger
-	formatter         *TextFormatter
-	module            string
-	allModulesEnabled bool
-	moduleLoggers     map[string]*Logger
-	PriorityKey       string
-	CriticalPriority  string
 }
 
-var (
-	// QuietLogger disables all log output
-	QuietLogger = logrus.Logger{
-		Out:       ioutil.Discard,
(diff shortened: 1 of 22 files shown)


src/util/logging/logger.go
```diff
@@ -1,172 +1,71 @@
...
-		} else {
-			newLogger = &Logger{
-				Logger:            &QuietLogger,
-				allModulesEnabled: logger.allModulesEnabled,
-				moduleLoggers:     logger.moduleLoggers,
-			}
-		}
+// PackageLogger instantiates a package-aware logger
+func (logger *MasterLogger) PackageLogger(moduleName string) *Logger {
+	return &Logger{
```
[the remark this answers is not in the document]
gz-c: They're all instantiated once globally per package, I don't think it would be a problem anyway, especially compared to WithField etc (which copies a map each time) or the formatter

src/util/logging/logger.go
```diff
@@ -1,172 +1,71 @@
...
+		FieldLogger: logger.WithField(logModuleKey, moduleName),
```
olemis: This is something I considered doing when submitting the patch for #858 . I noticed that there is a difference between logger and entry instances , even if they both implement `FieldLogger` interface. Instances of the former are meant to be stand-alone objects whereas the later are stored in a pool and eventually released back to the pool.

Since [`WithField` method returns an entry](https://github.com/sirupsen/logrus/blob/master/logger.go#L93-L105) , it was not clear to me whether this object interacted with the pool somehow . If that was the case then it might be put back into the logger pool . It was not clear to me either whether reused entries could be polluted from one call to the next one . Another scenario that came into my mind was a change applied upon the `Data` of the persistent entry referenced by our `PackaLogger` instances. This would be definitely feasible and would change significantly logger behavior , thus leading to obscure errors very hard to detect .

That's why I didn't follow that path and stick to logger instances instead of keeping a reference to entries beyond the life-time of the `FieldLogger` method calls. In the previous implementation entries were ephemeral , and I'd recommend to keep it that way so as to avoid headaches . Nonetheless , if this is no

gz-c: It turns out it is unsafe to modify entry.Data ourselves as was done in the ModuleHook (it paniced on a concurrent map write once I condensed everything into MasterLogger, I checked previous github issues and the author mentioned Hooks are not meant to write to Data), hence the prefix handling is moved to the formatter.

The entry pooling shouldn't have any effect on this

olemis: 👍

Issue #1145, opened 2018-03-24
Preview/confirm transaction view in UI 


gz-c:
Before sending, display a confirmation dialogue which shows all details of the transaction, including coinhours to be sent.

This requires a `/createRawTransaction` endpoint.

If we rely on something such as `/spend?preview=true`, then we must make sure that `/spend` is deterministic. This would be complicated and require additional parameters.

Instead, the API calls will need to be like:

`/createRawTransaction` - client displays this to the user
`/injectTransaction` - client provides the transaction to spend it

samuelvisscher:
It would be really great if the underlying method could be used to create raw transactions in the web cipher and the mobile cipher. This would, for example, be possible if we would allow the underlying method to receive a list of inputs, destination address, destination amount, change address, or alternatively list of inputs and list of outputs.

gz-c:
That should be handled by #872

PR #1379, opened 2018-04-25
Submit forms by pressing enter

### LINKED ISSUE #1376: Submit forms with enter

mahansky:
Fixes #1376

Changes:
- forms can be submitted by pressing enter instead of clicking buttons

### DIFF
--- a/src/gui/static/src/app/components/pages/wallets/change-name/change-name.component.html
+++ b/src/gui/static/src/app/components/pages/wallets/change-name/change-name.component.html
@@ -9,7 +9,7 @@
     <app-button (action)="closePopup()">
       Cancel
     </app-button>
-    <app-button (action)="rename()" class="primary" [disabled]="!form.valid">
+    <app-button (action)="rename()" class="primary" [disabled]="!form.valid" #button>
       Rename
     </app-button>
   </div>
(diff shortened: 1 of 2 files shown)


Senyoret1:
There are some issues:

- In the components in which the functionality was added, it is not possible to press the button repeatedly to execute the action several times, because the button stops sending click events as soon as it enters in the loading state. However, no safety measures of this type were placed when pressing enter, so it is possible to execute the action several times if you press enter several times fast. I think this could be especially bad on the send coins page.

- Pressing enter skips the validations in the wizard's create wallet screen and nothing happens in the load wallet screen. However, it might be best to leave these pages unchanged, since the wallet would be created by pressing enter while the focus is on a textarea, which would be against the way textareas are expected to work.

- It would probably be good idea to leave the send coins page unchanged, because any user error on that screen could be very bad, and also the last field is a textarea. In fact, it may be good to place a confirmation dialog, but in another pr.

- In the create wallet dialog also should not have the option to proceed by pressing enter while the focus is in a textarea. In this case it could even affect users using seeds with more than one line.

- It probably would be good to define if the pressing enter should work only if the focus is on the last field of the form (the option that seems best to me) or on any of them (as long as it is not a textarea) and check if all the forms work the same way.

mahansky:
Valid points. 
- I removed listeners from textareas, added them only to last inputs in the forms
- Fixed double submit
- Removed functionality from wizard create wallet (encrypt step still works)

Send form will get a "Preview" #1145

Senyoret1:
It is still possible to send multiple times by pressing enter in the change name dialog. In fact, it is also possible to do it by pressing the button repeatedly.

PR #1381, opened 2018-04-25
[pex] Fixes #712 Connection Pool Three Peer List

### LINKED ISSUE #712: Connection Pool: Should be Three peer lists
1> First list is peers that are configured by hand by node owner. Should always to connect to all peers on list. Peers should not count towards the connection pool and should not disconnect peers if over connection pool limit.

2> Second list of peers is default peers. These are peers the client will use to connect to network.

stdevEdu:


olemis:
@stdevEdu is this PR finished ? It needs a description based on [PR template](https://github.com/skycoin/skycoin/blob/develop/.github/pull_request_template.md)

PR #1385, opened 2018-04-26
Always report error msg on 500; remove duplicate log lines

gz-c:
Changes:
- Always include an error message for HTTP 500 errors
- Remove duplicate logging of errors, all messages are logged automatically now

No

### DIFF
--- a/src/gui/blockchain.go
+++ b/src/gui/blockchain.go
@@ -17,8 +17,8 @@ func blockchainHandler(gateway Gatewayer) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		metadata, err := gateway.GetBlockchainMetadata()
 		if err != nil {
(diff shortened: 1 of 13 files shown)


iketheadore:
How about use optional arguments as Error msgs, like:

```go
func Error500(w http.ResponseWriter, msgs ...interface{}) {
    msg := msgsToString(msgs)
    errorXXXMsg(w, http.StatusInternalServerError, msg)
}

func msgsToString(msgs ...interface{}) string {
        ss := []string{}
        for _, msg := range msgs {
                ss = append(ss, fmt.Sprint(msg))
        }
        return strings.Join(ss, " ")
}
```

In this way, we can write `Error500(w)`, instead of `Error500(w, "")`, also if we have error, we can write `Error500(w, err)`, and don't have to write `Error500(w, err.Error())`.

gz-c:
I don't like to use ... or interface{} since it breaks type system

PR #1393, opened 2018-04-26
Fix a bug in /blockchain/progress api

### LINKED ISSUE #1392: Bad blockchain progress in wallet banner
Some one encountered the following error, which is caused by a bug in `blockchain/progress` api.

image

iketheadore:
Fixes #1392

Changes:
- Fix a bug in `/blockchain/progress` api

No


gz-c:
If ourLen > maxLen, shouldn't we use ourLen?

gz-c:
Does this fix that bug? Their UI shows 22233/9, isn't 9 the estimated blockchain length? So maxLen should have been ourLen

iketheadore:
Yes, I think this should be able to fix, the 9 is the estimated blockchain length, and the previous code, if we only have two peers in the pool and all have 9 as highest seq, then the estimated blockchain length would be 9.

PR #1399, opened 2018-04-27
Use /wallet/transaction endpoint 

### LINKED ISSUE #1335: Switch frontend spend to use POST /wallet/transaction API
Switch from using `POST /wallet/spend` to using `POST /wallet/transaction`

https://github.com/skycoin/skycoin/blob/develop/src/gui/README.md#create-transaction

For `"hours_selection"`, use 

```json
{
  "type": "auto",
  "mode": "share",
  "share_factor": "0.5"
}
```

This API returns an encoded transaction under `"encoded_transaction"` in the response object.  Use this value to a subsequent request to `POST /injectTransaction`

mahansky:
Fixes #1335

Changes:
- use `/wallet/transaction` instead of `/wallet/spend`


gz-c:
Note that injectTransaction can fail and if it does, it should look the same to the user as if createTransaction fails

gz-c:
@Senyoret1 how does it burn more coin hours?

mahansky:
I noticed that as well. I'm not sure what exactly causes that (many unspent outputs or sending everything).
For example, having 1 coin and 53 coin hours, when I want to send it to same address (1 coin), it burns 40 coin hours, leaving me with 13 (75% burned). 
Sending 0.1 burns just 7 coin hours. (from one output that has 14, which makes it 50%)

Senyoret1:
Because of this (from the REST API Documentation):

> Note that if there are remaining coin hours as change, but no coins are available as change from the wallet, these remaining coin hours will be burned as an additional fee.

Using `/wallet/transaction`, if all the coins of the outputs are used, 75% of the hours is burned. `/wallet/spend` sends the remaining hours to the recipient, so the burn rate is always 50%.

therealssj:
@Senyoret1 good point there. I think the answer to that lies here https://github.com/skycoin/skycoin/blob/develop/src/wallet/wallet.go#L1235

there is no else part of this condition where in we could try to split the coinhours among the output addresses to make it behave like `/wallet/spend`
But I m unable to recollect why we decided to go that way

Senyoret1:
I did not investigate so deeply, hehehe... I hope it is solvable.

PR #1408, opened 2018-04-30
Send - transaction verification

### LINKED ISSUE #1145: Preview/confirm transaction view in UI

### LINKED ISSUE #1446: (Wallet) Select style in the send coins page
The select control on the send coins page does not show the arrow on the right side and does not show a text asking the user to choose an option. I think it would be better to remove `-webkit-appearance: none;` from it and put an initial text like the one in the web wallet:

drop

mahansky:
Fixes #1145 
Fixes #1446 

Changes:
- added extra step to "Send" page - transaction verification
  - it shows inputs and outputs of the transaction
  - fee
  - option to submit or go back and edit
- changed transaction detail modal in transaction history to look the same as transaction verification

Yes

~Depends on #1399~
~Should wait for #1407 to update balance right after tx is submitted~

screen shot 2018-04-30 at 15 21 20


src/gui/static/src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
```diff
@@ -0,0 +1,44 @@
...
+  @Input() transaction: any;
+  @Input() encodedTransaction: string;
+  @Output() onBack = new EventEmitter<boolean>();
+
+  constructor(
+    private walletService: WalletService,
+  ) {}
+
+  send() {
+    if (this.sendButton.isLoading()) {
```
Senyoret1: This is only necessary if `send()` can be called in some other way than pressing the button.

Senyoret1:
I think this is a very good improvement. In fact, except for the conflict, this pr would be ready to be mixed. Only a few things that could be worth discussing:

- Before, when a transaction failed the error was shown in a snackbar, now it's only shown on the button. I understand that it was redundant to show the error in both parts, but taking into account that this is one of the most used pages of the wallet, would not it be better to show the error in the snackbar, even if it is not shown in the button? It's matter of opinions, I'm not affirming that the snackbar is the right control to use.

- In the modal windows of the wallet, the cancel button is always on the left and the confirm button is on the right. It would be good to do the same in this case, to maintain consistency, but I think it would only look good if the buttons are centered or aligned to the right. If this is done and the alignment of the buttons is changed, it would also be necessary to change the alignment of the "Verify" button in the send coins form.

Senyoret1:
Additionally, I like the information shown, but maybe for a common user it may be confusing, especially as the box with the details of the transaction does not have a title. What do you think about placing a text on the top saying something like "sending x coins to j address". That text could be separated by a light gray line, such as those separating the addresses in the wallets page. That would probably help to improve the usability for less experienced users.

gz-c:
I don't know if the snackbar is the best way to show the error, but I feel that the button error is not enough to make it clear to the user that it failed.

Button order should be consistent in all views.

There should be a summary of the transaction above the inputs/outputs that shows the important data for the user like coins sent, coin hours sent, coin hours burned.

In electrum, there is a single "transaction" view that is used for previews and for completed transactions. We could possibly have the same thing. Metadata at the top, then inputs/outputs, then control buttons.

mahansky:
I need some inputs from you @Senyoret1 @gz-c 

1) current layout + info about transactions you mentioned
screen shot 2018-05-07 at 13 57 41

2) has just basic info about transaction + there would be a little toggle `[+] Show Inputs & Outputs` that would expand the info to be the same as 1)
screen shot 2018-05-07 at 13 58 00

3) use @gz-c 's idea to have the same component for existing transactions and for preview
This is the current component for existing tx
screen shot 2018-05-07 at 13 59 37
This is how it could look when previewing tx you are about to send
screen shot 2018-05-07 at 14 26 15

Also, what's the correct naming convention?
Skycoin, SKY, Coin
Hours, Coin hours, SKY Hours

Senyoret1:
Personally, I prefer option 2. Keeping the inputs and outputs hidden can help to make the UI more user-friendly for inexperienced users, and it would be more likely to be able to press "Send" without having to scroll first. Also, in my opinion, the modal is a bit more difficult to read, more cluttered.

However, it would be good to hear other opinions.

gz-c:
Option 2 is fine, but you can include the txid in the view too.  If the API is missing the txid, let me know (could be labelled "hash").  The window may need a title too ("Transaction Preview", "Confirm Transaction" e.g.)

gz-c:
We could change the existing transaction modal to look more like the data in option 2.  I think they should be consistent, but we don't need to stick with the current design.  Hours sent, hours burned, inputs and outputs are useful information for confirmed transactions as well.

mahansky:
+ Transaction detail in transaction history looks the same as transaction verification step.

`/explorer/address` endpoint doesn't have `fee` property, so I can't show "burned" hours in tx history detail

This is how I calculate `sent hours`:
Find outputs with addresses which are not input addresses, add their hours together.
If the result is 0 and there is just one output, use it's hours (when you send everything to yourself)

gz-c:
#1479 to add `fee` to the `/explorer/address`

If its complicated to compute fee for historical transactions, you can omit it for now

Senyoret1:
👍  I filed a separate issue: https://github.com/skycoin/skycoin/issues/1494

gz-c:
@Senyoret1 good?

Senyoret1:
This is almost ready. Only one detail remains: the amount of coins sent is incorrect. Here is an example:

hours1

In this example, the return address is fAaN, so only 376 hours are being sent to U2TF. However, the control says that 752 hours are being sent. Additionally, when fixing this problem, it is necessary to consider the following special case:

hours2

In that case, the coins are being sent from the second address of the wallet and the remaining coins are being returned to the first address of the wallet, so the approach that was prepared in the code (discard the hours sent to addresses that are in the input list) would fail.

Apart from this, everything seems to be pretty good. @gz-c If you want to merge this pr now, you may do it and a new issue could be opened about the sent hours calculation.

PR #1411, opened 2018-05-01
Wallets page styles

### LINKED ISSUE #1250: Improve wallet operation button visibility
The wallet operation buttons [New address, hide empty, encrypt wallet, edit wallet] are placed below the wallet's addresses so are easy to miss when there are a lot of addresses.  Possibly this should be the first row.  Furthermore the button text is very grey suggesting they can't be clicked, and there's no onhover change to suggest otherwise.

mahansky:
Fixes #1250 

Changes:
- hover effect on wallet action buttons (add address, rename..)
- opened wallet has lighter background 
- coins/hours are aligned to the right
- link to outputs is an icon instead of text 

Opened wallet:
screen shot 2018-05-01 at 15 05 37

Hover effect:
screen shot 2018-05-01 at 15 05 58


Senyoret1:
That sounds good. The option to copy the address would be less accessible, but that should not be a big problem, hardly the users need to copy the addresses as fast as possible, and it will allow to easily add more options in the future, if necessary. Also, with that change it would be good to make the mouse pointer stay with the standard icon when it is over the rows, because now it seems to indicate that it is possible to click anywhere.

mahansky:
- lightened borders of opened wallet 
- removed `cursor:pointer` from address row
- fixed the alignment problem
- added small dropdown to wallet detail
- changed style of dropdowns (skycoin font + border radius)

src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.html
```diff
@@ -3,22 +3,25 @@
     <div class="id-column">{{ i + 1 }}</div>
     <div class="address-column" (click)="copyAddress(address)">
       <img src="../../../../../assets/img/qr-code-black.png" (click)="showQrCode($event, address.address)">
-      <p class="click-to-copy" [ngClass]="{ copying: address.copying }">
-        {{ address.address }} <span data-label="Copied" class="copy-label">Copy</span>
-      </p>
+      <p class="click-to-copy" [ngClass]="{ copying: address.copying }">{{ address.address }}</p>
```
Senyoret1: `click-to-copy` and `copying` were removed from the css file.

src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.html
```diff
@@ -3,22 +3,25 @@
...
-    <div class="coins-column">
-      {{ (address.coins ? address.coins : 0) | number:'1.0-6' }}
-    </div>
+    <div class="coins-column">{{ (address.coins ? address.coins : 0) | number:'1.0-6' }}</div>
     <div class="outputs-column">
-      <mat-icon class="outputs-label"
-                routerLink="/settings/outputs"
-                [queryParams]="{ addr: address.address }"
-      >chevron_right</mat-icon>
+      <mat-icon class="outputs-label" [matMenuTriggerFor]="optionsMenu">more_vert</mat-icon>
```
Senyoret1: This is no longer a label, so the name of the css class should change.

PR #1417, opened 2018-05-02
Fixes #1388 Fusion of GetConnections () with GetBlockchainProgress ()

### LINKED ISSUE #1388: Include peer's reported block height in /connections responses
After this is done, then the client can display the value in the connections panel.

This data is already in /blockchain/progress but should be added to /connections data

stdevHan:
Fixes #1388 

Changes:
-Modifying /daemon/gateway.go in GetConnections () and crafting a new struct for the output of the api

No


olemis:
Wrong approach , see #1447 .

PR #1425, opened 2018-05-03
Fixed #420 API Cleanup

### LINKED ISSUE #420: API cleanup
The API methods have inconsistent path and argument naming, and their response json formats also have inconsistent names and values.  Known defects collected here:

- [ ] `/explorer/getEffectiveOutputs` should be `/explorer/coinSupply`
- [ ] `/outputs` value for "hours" is an int, but "coins" are a string.  The value for "hours" should be a string

stdevHan:
Fixes #420 

Changes:
-Modified the `struct ReadableOutput` so that `hours` returns a `string` in the endpoint

No


gz-c:
#420 is not ready to be worked on and we cannot change the API response data until we add versioning

PR #1432, opened 2018-05-05
Fixed #485 Assigned prefix `/v1` to API endpoints

### LINKED ISSUE #485: Version the REST API
API endpoints need to be prefixed with something like `/v1/` so we can handle breaking changes.
Need to figure out plan for how to do this correctly

stdevHan:
Fixes #485

Changes:
-Modify the file `src/gui/client.go`, to assign the prefix
-Create a new function in `src/gui/http.go` called` webHandlerAPI`, which assigns the prefix to the urls declared above, separating them if they are stables, assigns the prefix `/v1`, but `/v2`
-Change the value of `nodeUrl` in `src/gui/static/src/environments /environment.prod.ts` so that the `desktopwallet` sees the API

No

### DIFF
--- a/src/gui/client.go
+++ b/src/gui/client.go
@@ -54,7 +54,7 @@ func NewClient(addr string) *Client {
 		Transport: transport,
 		Timeout:   httpClientTimeout,
 	}
-	addr = strings.TrimRight(addr, "/")
+	addr = strings.TrimRight(addr, "/v1")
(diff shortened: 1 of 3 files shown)


gz-c:
We can simply register all existing endpoints a second time, with the /v1/ prefix prepended, and update all documentation to refer to the /v1/ endpoint instead.

PR #1447, opened 2018-05-08
[daemon] fixes #1388 - Add PeerHeight field in RPC Connection struct

### LINKED ISSUE #1388: Include peer's reported block height in /connections responses

olemis:
Visor is used to put together a table to lookup blockchain height by peer address.
This is done once per request

Fixes #1388

Changes:
- Add `PeerHeight` field in `./src/daemon/rpc.go:Connection` struct
- Compute (peer) address to block height lookup table
- Build aforementioned index once, per API request

yes


olemis:
Wrong approach

Issue #1426, opened 2018-05-03
Add Content-Security-Policy header to HTTP responses


gz-c:
`Content-Security-Policy: script-src 'self' <host>`

Also add a cli flag, `-disable-csp`, which will disable this header.

mahansky:
`ng serve` uses `eval()` which violates this CSP

I can think of two solutions
1. add header on server side, but only when the app is running in "production" mode (after build)
2. add meta tag to index.html using javascript

Also, going a bit further, if the first option is selected, CSP policy could be improved to contain hashes of js files

gz-c:
What is `ng serve` used for? development?

gz-c:
If it's only a development thing, we can add a cli flag to disable it

mahansky:
Yes, it's just for development, cli flag is ok

PR #1448, opened 2018-05-08
Remove CSP tag

mahansky:
This is a temporary solution to this problem https://github.com/skycoin/skycoin/issues/1426#issuecomment-387392305

Changes
- remove CSP meta tag because `ng serve` uses `eval()`


gz-c:
If we use a CSP header, will that be sufficient? Or is there additional protection given by having it in the index.html

mahansky:
CSP Header is preferred to meta tag, so it will be okay.

PR #1463, opened 2018-05-09
Simplify seed trimming

### LINKED ISSUE #1459: Instead of cleaning up the seed input value, warn the user if the seed needs cleaning
Instead of transforming the seed input (replacing multiple spaces, trimming), display a warning to the user if their seed would be different after transformation.

The user should still be allowed to use these values as seeds.

mahansky:
Fixes #1459

Changes:
- only last newline character is removed from seed (allowing multiline input, leading/trailing whitespaces)

### DIFF
--- a/src/gui/static/src/app/services/wallet.service.ts
+++ b/src/gui/static/src/app/services/wallet.service.ts
@@ -51,7 +51,7 @@ export class WalletService {
   }
 
   create(label, seed, scan, password) {
-    seed = seed.replace(/\r?\n|\r/g, ' ').replace(/ +/g, ' ').trim();
+    seed = seed.replace(/\r?\n|\r$/, '');
 
     return this.apiService.postWalletCreate(label ? label : 'undefined', seed, scan ? scan : 100, password)
       .do(wallet => {
--- a/src/gui/static/src/app/services/wallet.service.ts
+++ b/src/gui/static/src/app/services/wallet.service.ts
@@ -51,7 +51,7 @@ export class WalletService {
   }
 
   create(label, seed, scan, password) {
-    seed = seed.replace(/\r?\n|\r$/, '');
+    seed = seed.replace(/(\n|\r\n)$/, '');
 
     return this.apiService.postWalletCreate(label ? label : 'undefined', seed, scan ? scan : 100, password)
       .do(wallet => {


Senyoret1:
In my tests, one new line is deleted only if it is at the beginning of the text, not at the end.

mahansky:
It wasn't modified in any way before that change.
This PR just removes last newline character https://github.com/skycoin/skycoin/issues/1459#issuecomment-387781246
Trim would be probably the safest way to go https://github.com/skycoin/skycoin/issues/725#issuecomment-367034613

So it's trim (newlines/whitespaces, can be left or right or both) vs last newline

Senyoret1:
With the latest commit this pr works correctly: deletes a new line character (only one) at the end of the seed, if it exists.

Now, I think it is very unlikely to be necessary, but just in case someone has a problem recovering a wallet, this is the story of how the processing of the seeds has changed:

- When the development of the wallet began, the wallet sent the seeds to the node as the user wrote them, regardless of whether they started or ended in white space, new line or any other character.

- On April 4 of this year 2 commits were created that eliminated all the white spaces at the beginning and the end of the seeds, without modifying the white spaces in the middle. However, this change was valid only until April 9, so it probably did not affect any user. (https://github.com/skycoin/skycoin/commit/a5446fb1ed040371f53dab65c949591dd3bf3f1a#diff-efbcae38996159431d5abd4a192ee0c1R31 and https://github.com/skycoin/skycoin/commit/d4f7bc0d2b913e4957b64797cd3f885b0f575306#diff-0e10fb135d474b6bcf07555f87955782R33)

- On April 9 of this year, a commit was created that invalidated the changes from April 4. From that moment all the white spaces were eliminated at the beginning and at the end of the seed, the sequences of more than one white space were replaced by a single white space and all the new line characters were eliminated. (https://github.com/skycoin/skycoin/commit/970276e9b407afc0a5f853d6a69a25b3c219349d#diff-6800b6315629d7fd3311e30dabd17e1c)

- From this pr, the seeds are sent to the node practically unprocessed, only one new line character is eliminated at the end of the seed (if there is at least one). So if a user intentionally entered a seed with characters that previous commits might have deleted and have problems when recovering the wallet, the user should try writing the seed using the filtering rules of the commits of April 4 and 9.

PR #1492, opened 2018-05-13
Set time value in /transaction api response & clean up unused disable wallet api settings for CLI integration test

iketheadore:
Changes:
- Clear unused disable wallet api settings for CLI integration tests, as CLI handles wallet files directly
- Set `time` in `/transaction` API endpoint

No


gz-c:
Needs changelog update since the API changed

iketheadore:
The api is not changed, we just forget to assign the time field value, if you test the `/transaction?txid=` api, you will find that the `time` field is always `0`.

PR #1501, opened 2018-05-15
change gui to api in logger and test flags

therealssj:
Changes:
- change `gui` to `api` in logger and integration test falgs

no


therealssj:
@gz-c done

PR #1281, opened 2018-04-18
[libc] fixes #1191 fixes #1628 - Automatic generation of go functions to export to C

stdevEclipse:
Fixes #1191 
- New generated code is compilable with make build-libc.
Changes:
- Wrapper for Go functions that can exported to c library. All functions are automatically generated by cgogen. Avoided functions that use function types, interfaces, map or external types, unless specified that these types are replaced by handles or other custome types.
- Code that was hand-written was kept intact. 
- Some minor bugs in generated code was fixed by hand. Due to array sizes using constants or complex expressions.
- Checks relying on secp256k1 for cipher golden test suite

PR #1502, opened 2018-05-15
Cipher output test suite

### LINKED ISSUE #1436: Cipher output test suite
Create a test suite that compares the output of cipher operations against "golden" files, i.e. data generated once with a current version.  This is to ensure the output of cipher does not change.

Use a set of 10 random seeds.  5 of them should be generated with `bip39.NewDefaultMnemomic()` and 5 should be generated with `cipher.SumSHA256(cipher.RandByte(1024))`.
Generate 10 deterministic <pub, sec, addr> sets from each of these seeds.  Check that the same values are always generated.

Use a set of 10 different random SHA256 hashes (generate 32 random byte and convert to `cipher.SHA256`). Add the null hash (all 0s), and a hash of all 1s to this list.
For each hash and secret key, generate a signature.  Check that the signature matches the known signatures.

Structure the golden file data such that it can be reused for other library implementations to test their output. e.g. libskycoin could use this data to verify that its cipher library is correct.

Additional specific tests, add a golden file for each:

- Using a seed of length 1, generate 10 addresses
- Using a random seed, generate 1000 addresses. Since this file would be large, don't add signatures, this one is only for checking address generation.

Vavilen:
Fixes #1436 

Changes:
- golden files generator
- testdata
- test suite tool

@gz-c , generated 10 seeds instead of 100 for now.


olemis:
Needed for merging #1225 #1281 .

PR #1503, opened 2018-05-15
Set indexed height in unspent pool addr index

### LINKED ISSUE #1499: Record last indexed block for unspent addr index
Record last indexed block for unspent addr index and check it during initialization for building the index, instead of checking if the index is empty or not.

Otherwise, if you switch to an older version of skycoin, the database can skip indexing for some blocks.

gz-c:
Fixes #1499 

Changes:
- Record the last indexed height when indexing unspent outputs, and rebuild if the height does not match the current block height

No


iketheadore:
Resolve conflicts

gz-c:
@iketheadore resolved

PR #1504, opened 2018-05-15
Limit the connections to default peers

### LINKED ISSUE #1367: Don't connect to all default peers
From the list of default peers, don't try to connect to all of them, only stay connected to a max of 1

iketheadore:
Fixes #1367 

Changes:
- Don't  connect to all default peers, add `-max-default-outgoing-connections` flag to config the maximum allowed connections to default peers.

No

### DIFF
--- a/src/daemon/messages.go
+++ b/src/daemon/messages.go
@@ -279,6 +279,10 @@ func (intro *IntroductionMessage) Handle(mc *gnet.MessageContext, daemon interfa
 			return ErrDisconnectOtherError
 		}
 
+		// Checks if the introduction message is from outgoing connection.
+		// It's outgoing connection if port == intro.Port, as the incoming
+		// connection's port is a random port, it's different from the port
+		// in introduction message.
 		if port == intro.Port {
 			if d.Pool.Pool.IsDefaultConnection(mc.Addr) {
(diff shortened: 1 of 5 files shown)


src/daemon/messages.go
```diff
@@ -280,6 +280,20 @@ func (intro *IntroductionMessage) Handle(mc *gnet.MessageContext, daemon interfa
 		}
 
 		if port == intro.Port {
+			if d.Pool.Pool.IsDefaultConnection(mc.Addr) {
```
gz-c: This will prevent default connections from exceeding 1 even if the default peer is incoming.  This is ok but the config value is named `MaxDefaultOutgoingConnections` so maybe change it to `MaxDefaultPeerConnections`

gz-c: This will prevent default connections from exceeding 1 even if the default peer is incoming.  This is ok but the config value is named `MaxDefaultOutgoingConnections` so maybe change it to `MaxDefaultPeerConnections`

iketheadore: No, it's actually only the outgoing connection, the `if port == intro.Port` here filters out incoming connections, as the incoming connection's port is a random port, it can't be the same as the port in introduction message.

PR #1508, opened 2018-05-15
Frontend cleanup

### LINKED ISSUE #1445: Some wallet subscriptions should be cleaned
The wallet uses a lot of observables, subjects and subscriptions. Many of those subscriptions are automatically cleaned. For example, after finishing an HTTP connection to the backend, the subscription is terminated (the `complete` callback is called). However, others are not cleaned properly and remain active throughout the session. This is not a critical problem, because the wallet has been working without many problems, but it would be good to fix it, because they cause memory leaks and in some cases cause multiple requests to be unnecessarily send repeatedly to the node throughout the session. In addition, this will be more important in the light clients, which could use code from this repository as reference.

There are 3 main cases in which this is happening.

The first is in subscriptions to subjects, as in https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/settings/pending-transactions/pending-transactions.component.ts#L21. In that line, the subscription is made indirectly to a subject, so the subscription never ends. A good example in which this was taken into account and the subscription was cleaned is in https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/layout/header/header.component.ts#L84

The second is in subscriptions to events of custom controls. An example is in https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.ts#L40, where the `complete` callback is never called. There are some examples of control events that were not created specifically for the wallet in which this problem does not happen, as in https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.ts#L108, where the `complete` callback is called.

The third is verey similar to the second, is in events of form elements, like in https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/layout/password-dialog/password-dialog.component.ts#L50, where the `complete` callback is never called.

I did some revisions and the problem seems to be in the following lines (although there could be more):

https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/settings/outputs/outputs.component.ts#L24

https://github.com/skycoin/skycoin/blob/develop/src/gui/stsatic/src/app/components/pages/settings/backup/backup.component.ts#L25

https://github.com/skycoin/skycoin/blob/develop/src/gui/stsatic/src/app/components/pages/settings/backup/backup.component.ts#L25

https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/settings/backup/backup.component.ts#L35

https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/settings/pending-transactions/pending-transactions.component.ts#L21

https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/send-skycoin/send-skycoin.component.ts#L47

https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/send-skycoin/send-skycoin.component.ts#L92

https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.ts#L40

https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.ts#L72

https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/buy/buy.component.ts#L56

https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/pages/buy/buy.component.ts#L77

https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/layout/header/header.component.ts#L58

https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/components/layout/password-dialog/password-dialog.component.ts#L50

mahansky:
Fixes #1445 

Changes:
- unsubscribed from active subscriptions
- added more rules to tslint
- removed dead code

### DIFF
--- a/src/gui/static/src/app/components/pages/send-skycoin/send-form/send-form.component.ts
+++ b/src/gui/static/src/app/components/pages/send-skycoin/send-form/send-form.component.ts
@@ -1,4 +1,4 @@
-import { Component, EventEmitter, Input, OnInit, Output, ViewChild } from '@angular/core';
+import { Component, EventEmitter, Input, OnDestroy, OnInit, Output, ViewChild } from '@angular/core';
 import { WalletService } from '../../../../services/wallet.service';
 import { FormBuilder, FormControl, FormGroup, Validators } from '@angular/forms';
 import 'rxjs/add/operator/delay';
@@ -7,19 +7,22 @@ import { ButtonComponent } from '../../../layout/button/button.component';
 import { PasswordDialogComponent } from '../../../layout/password-dialog/password-dialog.component';
 import { MatDialog, MatSnackBar, MatSnackBarConfig } from '@angular/material';
 import { parseResponseMessage } from '../../../../utils/errors';
+import { ISubscription } from 'rxjs/Subscription';
 
 @Component({
   selector: 'app-send-form',
   templateUrl: './send-form.component.html',
   styleUrls: ['./send-form.component.scss'],
 })
-export class SendFormComponent implements OnInit {
+export class SendFormComponent implements OnInit, OnDestroy {
   @ViewChild('button') button: ButtonComponent;
   @Input() formData: any;
   @Output() onFormSubmitted = new EventEmitter<any>();
 
   form: FormGroup;
   transactions = [];
+
+  private subscription: ISubscription;
 
   constructor(
     public formBuilder: FormBuilder,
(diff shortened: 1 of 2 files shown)


mahansky:
I somehow deleted it when resolving merge conflicts, it's back.

PR #1512, opened 2018-05-16
CLI should have non-zero exit code if command not found

### LINKED ISSUE #1314: CLI should have non-zero exit code if command not found

Vavilen:
Fixes #1314 

Changes:
- exit code 1 for cli if command is not found


src/cli/cli.go
```diff
@@ -278,6 +278,7 @@ func NewApp(cfg Config) (*App, error) {
 	app.CommandNotFound = func(ctx *gcli.Context, command string) {
 		tmp := fmt.Sprintf("{{.HelpName}}: '%s' is not a {{.HelpName}} command. See '{{.HelpName}} --help'.\n", command)
 		gcli.HelpPrinter(app.Writer, tmp, app)
+		gcli.OsExiter(1)
```
gz-c: Doesn't this make it always exit with 1?

iketheadore: It's the same as `os.Exit(1)` here if the `OsExiter` is not set.

PR #1520, opened 2018-05-18
Cipher test suite

### LINKED ISSUE #1436: Cipher output test suite

gz-c:
Fixes #1436 

Changes:
- `cmd/cipher-testdata/` to generate testdata for the cipher test suite
- create `cipher/testsuite` with tests and testdata

Yes


cmd/cipher-testdata/cipher-testdata.go
```diff
@@ -0,0 +1,214 @@
...
+		fmt.Fprintf(os.Stderr, "%s\n\nUsage of %s:\n", help, os.Args[0])
+		flag.PrintDefaults()
+	}
+}
+
+func main() {
+	var j job
+
+	var seedsCount int
+	flag.IntVar(&seedsCount, "seeds", 10, "number of seeds to generate")
```
iketheadore: Can be `seedsCount := flag.Int("seeds"...` like the following flags.

src/cipher/testsuite/testsuite.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil, err
+	}
+
+	p, err := cipher.PubKeyFromHex(d.Public)
+	if err != nil {
+		return nil, err
+	}
+
+	var sigs []cipher.Sig
+	if d.Signatures != nil {
```
iketheadore: Seems no need to check if `d.Signatures` is nil, `range` would take care of it.

gz-c: I want it to be exact copy, so that if d.Signatures is nil, new sigs will be nil, but if d.Signatures is non-nil empty array, new sigs will be non-nil empty array

PR #1521, opened 2018-05-18
Add ngx-translate

### LINKED ISSUE #1263: Use string translation object for frontend
Use the string translation object thing that i18n uses so we can support multiple languages and also make it easier to modify the frontend for other coins (e.g. so they can replace SKY with their coin's label with one line)

mahansky:
Fixes #1263 

Changes:
- added ngx-translate to handle translation
- added "en" translation 

Some strings are duplicated on multiple places to support more customization "per component".


Senyoret1:
Please also add the texts of this pr: https://github.com/skycoin/skycoin/pull/1522

mahansky:
I originally skipped translating errors because backend returns english errors anyways.

What do you think about strings like these?
https://github.com/mahansky/skycoin/blob/66753ba3a617321e8efe4af582cb01d433d95684/src/gui/static/src/assets/i18n/en.json#L41
There is one more way to do it - by rephrasing the sentence to have the html part at the end. So "Wallet update v0.23.0 available" would be "Wallet update available: v.0.23.0"

PR #1522, opened 2018-05-18
Current rate warning

### LINKED ISSUE #1494: Tell the users that the history page shows amounts in USD at the current rate
The history page in the wallet shows the amount of coins that have been sent or received, as well as their value in USD, calculated according to the current exchange rate. It would be good to tell the user this last fact, that the value in USD is calculated according to the current exchange rate, not the exchange rate at the moment when the transaction was made, to avoid confusion.

This change should be made after https://github.com/skycoin/skycoin/pull/1408

mahansky:
Fixes #1494

Changes:
- shows tooltip when viewing sky->usd conversion on Transaction History page


Senyoret1:
It works correctly. However, after merging https://github.com/skycoin/skycoin/pull/1521 it will be necessary to use ngx-translate for the texts, so it might be better to wait and for making that change in this pr.

gz-c:
I'll merge this first, ngx-translate PR can rebase/merge and add this to it

PR #1524, opened 2018-05-19
Refactor skycoin

therealssj:
Changes:
- create skycoin package
- run skycoin node cli using skycoin package
 

No


gz-c:
conflicts

Gopkg.lock
```diff
@@ -112,6 +112,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "454aacd6dc7c6819166ecc75a75483f4bcebc23ab61c4b2b32e694ee67020d52"
+  inputs-digest = "33d8d0c2675c0851febc914bfcadef2a117d2cc81db1cfea6ca00da527543f18"
```
gz-c: why did this change?

cmd/skycoin/skycoin.go
```diff
@@ -295,7 +101,7 @@ var devConfig = Config{
 
 	LaunchBrowser: false,
 	// Data directory holds app data -- defaults to ~/.skycoin
-	DataDirectory: filepath.Join(home, ".skycoin"),
+	DataDirectory: "~/.skycoin",
```
gz-c: I think we should use `$HOME` and manually substitute that

src/skycoin/config.go
```diff
@@ -0,0 +1,278 @@
...
+
+// BuildConfig records build info
+type BuildConfig struct {
+	Version string // version number
+	Commit  string // git commit id
+	Branch  string // git branch name
+}
+
+// BlockchainConfig records genesis data
+type BlockchainConfig struct {
```
gz-c: All of these can be merged into a single Config struct.  If we don't want to expose any as CLI parameters, we just won't register them with `flag`

Gopkg.lock
therealssj: oh thats becoz I had introduced some dependencies in an earlier commit but removed them now

cmd/skycoin/skycoin.go
therealssj: yea right now I am substituting `~` with the home path.

src/skycoin/testdata/test.fiber.toml
```diff
@@ -0,0 +1,26 @@
...
+	"172.104.85.6:6000",
+	"139.162.7.132:6000",
+]
+launch_browser = true
+peer_list_url = "https://downloads.skycoin.net/blockchain/peers.txt"
+
+[build]
+version = "0.23.1-rc2"
+commit = "0aab9bf7730827d6fd11beb0d02096b40cea1872"
+branch = "test-branch"
```
gz-c: make sure newline at end of files

fiber.toml
```diff
@@ -0,0 +1,73 @@
...
+#db_path = ""
+#db_read_only = false
+#arbitrating = false
+#log_to_file = false
+#version = false
+
+[build]
+version = "0.23.1-rc2"
+#commit = ""
+#branch = ""
```
gz-c: make sure newline at end of files

Gopkg.lock
gz-c: still need to undo the deps change

gz-c:
merge conflict

therealssj:
@gz-c done

gz-c:
Need to update vendored dependencies, according to vendorcheck

gz-c:
build still failing for the same reason

PR #1540, opened 2018-05-22
Force share_factor 1.0 if coin hours would be otherwise lost

### LINKED ISSUE #1421: POST /wallet/transaction, redistribute hours if extra change output can't be made
If extra change cannot be created to preserve coin hours, redistribute the leftover hours to the destination addresses instead of burning more coin hours.

gz-c:
Fixes #1421 

Changes:
- In `POST /wallet/transaction`, if there are extra coin hours but change cannot be made, and the type is `auto` and the mode is `share`, then force the `share_factor` to be 100% so that no extra coin hours are burned.

Yes


gz-c:
In the future, if there is a need to handle the lost change hours differently in different use cases, we can implement the following:

```go
	// ExtraChangeHoursPolicyError will return an error if extra change hours would be lost
	ExtraChangeHoursPolicyError = "error"
	// ExtraChangeHoursPolicyRedistribute will redistribute extra change hours to the outputs
	ExtraChangeHoursPolicyRedistribute = "redistribute"
	// ExtraChangeHoursPolicyBurn will burn extra change hours as fee
	ExtraChangeHoursPolicyBurn = "burn"

	// DefaultExtraChangeHoursPolicy is the default ExtraChangeHoursPolicy if the value is not set
	DefaultExtraChangeHoursPolicy = ExtraChangeHoursPolicyRedistribute
```

And allow this to be controlled in the API.

For now, this would complicate the code and there is no definite use case for it.  Although ignoring the share_factor is not good from an API standpoint, returning an error is the least sensible action from most users' perspectives.  The only way for a user to correct the situation would be to change the hours they are sending to 1.0 anyway, reduce the number of coins they are sending, or obtain another unspent output. 

Burning makes sense if you don't care about the hours the recipients get, and if you would rather have a higher prioritized transaction.  At this point in the network, it is nowhere near transaction capacity, so priority is irrelevant.

PR #1547, opened 2018-05-22
Randomize desktop port

### LINKED ISSUE #1275: Randomize web interface port when running as desktop wallet

gz-c:
Fixes #1275 

Changes:
- Randomize the web interface port when running as a desktop client, by specifying port 0.
- This mitigates an attack where the attacker manages to close the backend node and run its own service on the stable port 6420, which the UI then connects to instead.

Yes


cmd/skycoin/skycoin.go
```diff
@@ -676,11 +667,22 @@ func Run(c *Config) {
...
 		}
+
+		scheme := "http"
+		if c.WebInterfaceHTTPS {
+			scheme = "https"
+		}
+		fullAddress = fmt.Sprintf("%s://%s", scheme, webInterface.Addr())
+		logger.Critical().Infof("Full address: %s", fullAddress)
+		if c.PrintWebInterfaceAddress {
+			fmt.Println(fullAddress)
```
iketheadore: Seems we should exit if `PrintWebInterfaceAddress` is true, as the comment says 
```go
// If true, print the configured client web interface address and exit
PrintWebInterfaceAddress bool
```

PR #1562, opened 2018-05-24
refs #1554 Fixed issue when removing wallet.

### LINKED ISSUE #1554: Error in skycoin when creating and deleting a wallet with a specific seed and then creating with the same seed.
Creating a wallet with a given specific seed, then deleting and unloading the wallet should allow creating a new wallet with the same seed.
There is a test in libc that is failing because of this. The test passes the first time but fails when executing again. First test is made with static library and second test with shared library. The node is not restarted between the two executions.

stdevEclipse:
Remove reference to wallet first address in field firstAddrIDMap.

Fixes #1554 

Changes:
- When deleting a wallet update map field firstAddrIDMap. This allows creating a wallet with the same seed.

No

### DIFF
--- a/src/wallet/service.go
+++ b/src/wallet/service.go
@@ -411,8 +411,18 @@ func (serv *Service) Remove(wltID string) error {
 	if !serv.enableWalletAPI {
 		return ErrWalletAPIDisabled
 	}
-
-	serv.wallets.remove(wltID)
+	
+	wlt, found := serv.wallets.get(wltID)
+	if found {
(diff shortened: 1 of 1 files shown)


gz-c:
/wallet/unload is only added to help with integration tests

iketheadore:
Yes, I will add comment about it.

olemis:
@stdevEclipse , travis build is failing .

gz-c:
This will overwrite existing wallet files on disk. There is no problem with `/wallet/unload`

PR #1581, opened 2018-05-27
Support omitempty for final field in encoder

### LINKED ISSUE #1576: Support omitempty flag for encoder struct tag

gz-c:
Fixes #1576

Changes:
- Support `omitempty` tag for enc struct. Only valid for the final field in a struct and only if the field is string, slice or map.  Intended usage is for a final `[]byte` slice field on structs to support transitional protocol upgrades.

No

### DIFF
--- a/src/cipher/encoder/encoder.go
+++ b/src/cipher/encoder/encoder.go
@@ -13,6 +13,15 @@
 // Varints are a method of encoding integers using one or more bytes;
 // numbers with smaller absolute value take a smaller number of bytes.
 // For a specification, see http://code.google.com/apis/protocolbuffers/docs/encoding.html.
+//
+// Fields can be ignored with the struct tag `enc:"-"`
+//
+// Fields can be skipped if empty with the struct tag `enc:",omitempty"`
+// Note the comma, which follows package json's conventions.
+// Only Slice, Map and String types recognize the omitempty tag.
+// When omitempty is set, the no data will be written if the value is empty.
+// If the value is empty and omitempty is not set, then a length prefix with value 0 would be written.
+// omitempty can only be used for the last field in the struct
 package encoder
 
 import (
@@ -22,6 +31,7 @@ import (
 	"log"
 	"math"
 	"reflect"
+	"strings"
 )
 
 /*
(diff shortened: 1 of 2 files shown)


src/cipher/encoder/encoder.go
```diff
@@ -382,20 +391,35 @@ func Size(v interface{}) int {
...
+	t := v.Type()
+	switch t.Kind() {
+	case reflect.String:
+		return v.Len() == 0
+	case reflect.Map:
+		return v.IsNil() || v.Len() == 0
+	case reflect.Slice:
+		return v.IsNil() || v.Len() == 0
+	default:
+		return false
```
iketheadore: Can check if empty for other types in this way:

```go
return reflect.DeepEqual(v.Interface(), reflect.Zero(t).Interface())
```

src/cipher/encoder/encoder.go
```diff
@@ -466,6 +502,18 @@ func datasizeWrite(v reflect.Value) (int, error) {
 	}
 }
 
+func parseTag(tag string) (string, bool) {
+	tagSplit := strings.Split(tag, ",")
+	name := tagSplit[0]
+
+	omitempty := false
+	if len(tagSplit) > 1 && tagSplit[1] == "omitempty" {
```
iketheadore: Trim space and use `tagSplit[len(tagSplit)-1]` to get the last tag might be better.

src/cipher/encoder/encoder.go
gz-c: Do we have any reason to?

src/cipher/encoder/encoder.go
gz-c: The tag format follows json package

src/cipher/encoder/encoder.go
gz-c: Really only meant for []byte, this omitempty tag is a hack to allow us to append a field temporarily

olemis: I know this is already merged , but I'd be in favor of having this added for all other types . Support limited to slices leads to [re-implementing encoder behavior](https://github.com/iketheadore/skycoin/blob/da40449/src/daemon/messages.go#L275-L289) .

PR #1585, opened 2018-05-28
newcoin creator tool

### LINKED ISSUE #1485: coin template creator tool
This is a part of https://github.com/skycoin/skycoin/issues/1474

The goal is to create a copy of https://github.com/skycoin/skycoin/tree/develop/cmd/skycoin/skycoin.go automatically with provided inputs

The values to be changed are(add more if any):
- `GenesisAddressStr`
- `BlockchainPubkeyStr`
- `GenesisCoinVolume`
- `GenesisTimestamp` (exact timestamp value is not important)
- `DataDirectory`
- `GenesisSignatureStr`
- `DefaultConnections` 

And the uxid here
https://github.com/skycoin/skycoin/blob/e915716d4decb8aa163898b93a841232159d414c/cmd/skycoin/skycoin.go#L776

But its not really necessary as the coin distribution will be handled by the tool https://github.com/skycoin/skycoin/issues/1468

---

- Create a `template` folder in the repo root and create a file `coin.template` inside it
- Copy over the contents of https://github.com/skycoin/skycoin/tree/develop/cmd/skycoin/skycoin.go
   into `coin.template` and put placeholders like `GenesisSignatureStr = "<GENESISSIGNATURESTR>"` etc
- Create a cli tool which will read a configuration file using https://github.com/skycoin/skycoin/issues/1484 and replace the placeholders in `coin.template` and create a new file `cmd/<coin>/<coin>.go` with the content of `coin.template`

therealssj:
Fixes #1485 

Changes:
- add new coin creator 
- move parameterized values from `config.go` to `parameters.go` in `skycoin` package
- add `VisorParameters` to `skycoin.Parameters`
- add visor parameters to`fiber.toml`
- move genesis constants from `distribution.go` to `parameters.go`

Yes


fiber.toml
```diff
@@ -65,9 +65,120 @@ peer_list_url = "https://downloads.skycoin.net/blockchain/peers.txt"
...
+   	"rQpAs1LVQdphyj9ipEAuukAoj9kNpSP8cM",
+   	"6NSJKsPxmqipGAfFFhUKbkopjrvEESTX3j",
+   	"cuC68ycVXmD2EBzYFNYQ6akhKGrh3FGjSf",
+   	"bw4wtYU8toepomrhWP2p8UFYfHBbvEV425",
+   	"HvgNmDz5jD39Gwmi9VfDY1iYMhZUpZ8GKz",
+   	"SbApuZAYquWP3Q6iD51BcMBQjuApYEkRVf",
+   	"2Ugii5yxJgLzC59jV1vF8GK7UBZdvxwobeJ",
+   	"21N2iJ1qnQRiJWcEqNRxXwfNp8QcmiyhtPy",
+   	"9TC4RGs6AtFUsbcVWnSoCdoCpSfM66ALAc",
+   	"oQzn55UWG4iMcY9bTNb27aTnRdfiGHAwbD",
```
therealssj: dont know why these are showing up like this.
looks fine on my machine.

gz-c: github renders tabs wider

src/skycoin/config.go
```diff
@@ -134,7 +137,7 @@ type NodeConfig struct {
 	DBReadOnly  bool   `mapstructure:"db_read_only"`
 	Arbitrating bool   `mapstructure:"arbitrating"`
 	LogToFile   bool   `mapstructure:"log_to_file"`
-	Version     bool   `mapstructure:"version"` // show node version
+	ShowVersion bool   `mapstructure:"show_version"` // show node version
```
gz-c: This shouldn't be changed. -version is a standard way to print an application's version.  The version wouldn't be configurable from the command line

gz-c: And this would always be false by default

src/visor/parameters.go
```diff
@@ -0,0 +1,130 @@
+package visor
+
+/*
+* CODE GENERATED AUTOMATICALLY WITH FIBER COIN CREATOR
+* AVOID EDITING THIS MANUALLY
+ */
+
+const (
```
gz-c: Should have some more sanity checks in an init() method in visor somewhere to make sure these values makes sense

template/coin.template
```diff
@@ -0,0 +1,179 @@
...
+var devConfig = skycoin.NodeConfig{
+	GenesisSignatureStr: GenesisSignatureStr,
+	GenesisAddressStr:   GenesisAddressStr,
+	GenesisCoinVolume:   GenesisCoinVolume,
+	GenesisTimestamp:    GenesisTimestamp,
+	BlockchainPubkeyStr: BlockchainPubkeyStr,
+	BlockchainSeckeyStr: BlockchainSeckeyStr,
+	DefaultConnections:  DefaultConnections,
+	// Disable peer exchange
+	DisablePEX: {{.DisablePEX}},
```
gz-c: Most of these options aren't meant to be configurable by the template.

They have the same defaults for all fiber coins because they control the basic app behavior.

The template needs to parse the fiber parameters that matter, merge them into a struct set with defaults then write the full struct out

cmd/newcoin/newcoin.go
```diff
@@ -0,0 +1,220 @@
...
+				Name:  "config-file, cf",
+				Usage: "config file path",
+				Value: "fiber.toml",
+			},
+		},
+		Action: func(c *cli.Context) error {
+			// -- parse flags -- //
+
+			templateDir := c.String("template-dir")
+			templateDir, err := homedir.Expand(templateDir) // expand ~
```
gz-c: Isn't `~` expanded by the shell before passed as an argument to the executable?

cmd/newcoin/newcoin.go
```diff
@@ -0,0 +1,220 @@
...
+			// -- parse template and create new coin.go and visor parameters.go -- //
+
+			config, err := skycoin.NewParameters(configFile, configDir)
+			if err != nil {
+				log.Errorf("failed to create new fiber coin config")
+				return err
+			}
+
+			coinDir := fmt.Sprintf("./cmd/%s", coinName)
+			if _, err := os.Stat(coinDir); os.IsNotExist(err) {
```
gz-c: The err from `os.Stat` needs to be returned if `!os.IsNotExist(err)`

src/skycoin/parameters.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return params, nil
+}
+
+func setDefaults() {
+	// node defaults
+	viper.SetDefault("node.genesis_coin_volume", 100e12)
+	viper.SetDefault("node.port", 6000)
+	viper.SetDefault("node.web_interface_port", 6420)
+	viper.SetDefault("node.data_directory", "$HOME/.skycoin")
```
gz-c: I think we can default these based on the coin name and not make them part of the configurable options

template/coin.template
```diff
@@ -0,0 +1,183 @@
...
+
+	// DefaultConnections the default trust node addresses
+	DefaultConnections = []string{
+    {{- range $index, $address := .DefaultConnections}}
+        "{{$address -}}",
+    {{- end}}
+	}
+)
+
+var devConfig = skycoin.NodeConfig{
```
gz-c: devConfig should not be in the template file

cmd/newcoin/newcoin.go
therealssj: looks correct to me. why would I negate an already negative condition?

PR #1589, opened 2018-05-28
Fixes #XXXX [encoder] [docs] Document encoder public functions

stdevEdu:
Added documentation to each public function in cipher/encoder package

### DIFF
--- a/src/cipher/encoder/encoder.go
+++ b/src/cipher/encoder/encoder.go
@@ -186,7 +186,12 @@ func DeserializeRaw(in []byte, data interface{}) error {
 	return d1.value(v)
 }
 
-// Deserialize takes reader and number of bytes to read
+// Deserialize reads `dsize` bytes from `r` and deserializes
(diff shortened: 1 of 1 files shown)


src/cipher/encoder/encoder.go
```diff
@@ -32,7 +32,10 @@ Todo:
 
 // TODO: constant length byte arrays must not be prefixed
 
-// EncodeInt encodes int
+// EncodeInt encodes an Integer type contained in `data`
+// into buffer `b`. If `data` is not an Integer type,
+// error message `log.Panic("PushAtomic, case not handled")`
```
therealssj: I dont think we should put the error message in the comments like that.
Its just unnecessary extra maintenance work if the message were to be changed.
I would trust the code more than a comment to see what error is returned. 
The comment would be just to get a better idea of what the function is doing.

Can just say that it panics instead of putting `log.Panic("PushAtomic, case not handled")` in the comments

PR #1590, opened 2018-05-29
Advanced spend frontend

### LINKED ISSUE #1423: Advanced spend UI
Create a new UI view for "advanced spend" which allows the user to control the parameters of the `POST /wallet/transaction` API.

We need to design the UI for this first.

Parameters:

- Which addresses to spend from (must be from the same wallet)
- Multiple destination addresses with coins.  Hours can be optionally specified.  If hours are specified for one they must be specified for all.
- Change address
- "Automatic" hours allocation.  Allow control of share factor percentage.

In the future, we'll let them choose which outputs to spend, but the API doesn't support this yet.

mahansky:
Fixes #1423

Changes:
- adds advanced spend form
  - source addresses from single wallet
  - change address
  - multiple destinations with coins and hours
  - coin hours allocation settings
- adds switch to change simple/advanced spend form to the nav bar
- added hammerjs dependency ([angular mat-slider needs it to work properly](https://material.angular.io/guide/getting-started#step-5-gesture-support))

Yes


src/gui/static/src/app/components/pages/send-skycoin/send-form-advanced/send-form-advanced.component.html
```diff
@@ -0,0 +1,102 @@
+<div [formGroup]="form">
+  <div class="form-field">
+    <label>{{ 'send.wallet-label' | translate }}</label>
+    <div class="-select">
+      <select formControlName="wallet">
+        <option *ngFor="let wallet of walletService.all() | async" [disabled]="wallet.balance <= 0" [ngValue]="wallet">
```
Senyoret1: The option does not get disabled if the wallet does not have coins.

gz-c:
@Senyoret1 at least when the form validation fails (e.g. letting you send 0 coins), the API returns an error and it is displayed to the user?

gz-c:
Also regarding a coinhour minimum, yes enforce a minimum of 1 in the UI

Senyoret1:
@gz-c yes, the api returns errors, and the errors are displayed to the user. When trying to send 0 coin the error is 
> To[0].coins must not be zero

And when trying to send 1.1 hours the error is

> Invalid hours value: strconv.ParseUint: parsing "1.1": invalid syntax

gz-c:
I'm not sure the slider should reset after hiding it, that could be unexpected.  I'm imaging if i opened the slider, changed it, hid it, then reopened it, the slider value would remain where i left it

gz-c:
The slider minimum should be at least 1% (0.01).  With #1604 it will also ensure at least 1 coin hour is sent to outputs when possible automatically.

Senyoret1:
Reseting the slider would be good if the user changes the slider, hides it, and sends the transaction, without showing it again. In that case, with the current code, the value of the slider would still be taken into account, but the expected behavior would surely be to send 50%, because it is most likely that the user hid the slider to cancel the change in the value, as there is no "reset" button and it may be uncomfortable to re-select 50% with the mouse.

To take into account both cases the code should use the value of the slider if it is visible, and send 50% if it is not.

therealssj:
can we add a `?` symbol somewhere which explains how the slider is going to work?
A simple tooltip explaining how it works.

Senyoret1:
A `?` symbol is a good idea. I think it is not only needed in the "value" field, but that field really is the most complicated to understand.

mahansky:
@Senyoret1 https://github.com/skycoin/skycoin/pull/1590/files#diff-902a7e85dac1f989066981e3a90d80a0R280 this is where the value is being assigned to share_factor 
I have no problem with the functionality. I can choose whether I want to send more or burn more.

I agree with @gz-c when it comes to resetting the value. 
Another solution to this problem would be to remove the "options" toggle entirely and leave the form visible all the time.

Senyoret1:
Wow, I have no idea why the slider did not work for me in the previous test, but it works correctly (I did several tests to try to reproduce the error and I could not). In the last commit the slider was limited to 0.01 and the help texts were added. The form works correctly, so I would say that this pr is practically ready for merging, except for the conflicts and the problem with the tests.

Issue #1517, opened 2018-05-18
Design two-stages upgrade to fix the mix port issue


iketheadore:
Steps to fix mix port issue:

- [ ] v24:Remove message size check so that we can accept message with extra fields. If genesis hash is present in message, parse it and reject if not matching. Requires #1576

- [ ] v25: Add genesis hash in INTR message,  disconnect the connection if receiving INTR message with different genesis hash. Accept the INTR message without genesis hash. 

- [ ] v26: Disconnect if receiving INTR message without genesis hash; disconnect if the genesis hash in INTR message doesn’t match.

Then
<=v23:  can connect to <=v24 [will reject incoming >=v25 because of extra field]
v24: can connect to <=v25 [would accept >=v26, but >=v26 will disconnect because of missing field]
v25: can connect to v24, v25 [would accept <=v23, but <=v23 will disconnect because of extra field]
v26: can connect to >=v25 [will reject <=v24 because of missing field]

gz-c:
@iketheadore is the first one done?

iketheadore:
Yes, the first one is done, except the genesis hash checking as we can't parse out the genesis hash without adding new fields to INTR message, while the `encoder` doesn't support `omitempty` tag, adding new field would close the connections v24 to <=23.

gz-c:
#1576 for omitempty struct tag

PR #1595, opened 2018-05-30
Check blockchain pubkey in INTR message if extra bytes exist

iketheadore:
Fixes: #1517 

Changes:
- Add `ExtraData []byte` field with `omitempty` tag in Introduction message
- Add the message data size checking back
- If `ExtraData` is not empty, then parse it as blockchain pubkey and compare it with local node's blockchain pubkey, disconnect if not match.

No

### DIFF
--- a/src/daemon/daemon.go
+++ b/src/daemon/daemon.go
@@ -53,6 +53,11 @@ var (
 	ErrDisconnectMaxDefaultConnectionReached                       = errors.New("Maximum number of default connections was reached")
 	// ErrDisconnectMaxOutgoingConnectionsReached is returned when connection pool size is greater than the maximum allowed
 	ErrDisconnectMaxOutgoingConnectionsReached gnet.DisconnectReason = errors.New("Maximum outgoing connections was reached")
+	// ErrDisconnectBlockchainPubkeyNotMatched is returned when the blockchain pubkey in introduction does not match
+	ErrDisconnectBlockchainPubkeyNotMatched gnet.DisconnectReason = errors.New("Blockchain pubkey in Introduction message is not matched ")
+	// ErrDisconnectInvalidExtraData is returned when extra filed can't be parsed as specific data type.
+	// e.g. ExtraData length in IntroductionMessage is not the same as cipher.PubKey
+	ErrDisconnectInvalidExtraData gnet.DisconnectReason = errors.New("Invalid extra data")
 
 	logger = logging.MustGetLogger("daemon")
 )
@@ -126,6 +131,8 @@ type DaemonConfig struct { // nolint: golint
 	Version int32
 	// IP Address to serve on. Leave empty for automatic assignment
 	Address string
+	// BlockchainPubkey blockchain pubkey string
+	BlockchainPubkey cipher.PubKey
 	// TCP/UDP port for connections
 	Port int
 	// Directory where application data is stored
@@ -887,7 +894,8 @@ func (dm *Daemon) onConnect(e ConnectEvent) {
 
 	dm.expectingIntroductions.Add(a, utc.Now())
 	logger.Debugf("Sending introduction message to %s, mirror:%d", a, dm.Messages.Mirror)
-	m := NewIntroductionMessage(dm.Messages.Mirror, dm.Config.Version, dm.Pool.Pool.Config.Port)
+	// TODO: replace the last paramenter of nil with dm.Config.BlockchainPubkey in v25
+	m := NewIntroductionMessage(dm.Messages.Mirror, dm.Config.Version, dm.Pool.Pool.Config.Port, nil)
 	if err := dm.Pool.Pool.SendMessage(a, m); err != nil {
 		logger.Errorf("Send IntroductionMessage to %s failed: %v", a, err)
 	}
--- a/src/daemon/daemon.go
+++ b/src/daemon/daemon.go
@@ -55,7 +55,7 @@ var (
 	ErrDisconnectMaxOutgoingConnectionsReached gnet.DisconnectReason = errors.New("Maximum outgoing connections was reached")
 	// ErrDisconnectBlockchainPubkeyNotMatched is returned when the blockchain pubkey in introduction does not match
 	ErrDisconnectBlockchainPubkeyNotMatched gnet.DisconnectReason = errors.New("Blockchain pubkey in Introduction message is not matched ")
-	// ErrDisconnectInvalidExtraData is returned when extra filed can't be parsed as specific data type.
+	// ErrDisconnectInvalidExtraData is returned when extra field can't be parsed as specific data type.
 	// e.g. ExtraData length in IntroductionMessage is not the same as cipher.PubKey
 	ErrDisconnectInvalidExtraData gnet.DisconnectReason = errors.New("Invalid extra data")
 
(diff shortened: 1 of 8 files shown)


olemis:
Just a question ... given this particular case of backwards-compatible upgrades of network packets . Wouldn't it be good practice to standardize this and add a reserved 32-bit field at the end of all packets in such a way that in light of a future upgrade the field is already there to hook to the next ?

The transition would be from a struct like ...

```
type SomeMessageStruct struct {
  // ... fields ...
  []byte reserved
}
```

could be upgraded to become e.g.

```
type SomeOtherStruct {
  // other fields
}

type SomeMessageStruct struct {
  // ... fields ...
  int32 extra_size // at least sizeof(SomeOtherStruct) + sizeof(int32)
  SomeOtherStruct OtherField
  []byte reserved
}
```

olemis:
In this particular case, and assuming we had a struct like this.

```
type IntroductionMessage struct {
	// Mirror is a random value generated on client startup that is used
	// to identify self-connections
	Mirror uint32
	// Port is the port that this client is listening on
	Port uint16
	// Our client version
	Version int32
	c       *gnet.MessageContext `enc:"-"`
	// We validate the message in Handle() and cache the result for Process()
	valid bool `enc:"-"` // skip it during encoding
	// Reserved for future use
	Reserved []byte `enc:",omitempty"`
}
```

and the upgrade would look like this .

```
type BlockchainCredentials {
  // Struct size
  int32 Size
  PubKey BlockchainPubKey
  // Reserved for future use
  byte[] Reserved
}

type IntroductionMessage struct {
	// Mirror is a random value generated on client startup that is used
	// to identify self-connections
	Mirror uint32
	// Port is the port that this client is listening on
	Port uint16
	// Our client version
	Version int32
	c       *gnet.MessageContext `enc:"-"`
	// We validate the message in Handle() and cache the result for Process()
	valid bool `enc:"-"` // skip it during encoding
	// Reserved for future use
	BlockchainIntroData BlockchainCredentials `enc:",omitempty"`
}
```

... and the same procedure could be repeated again and again .

src/daemon/messages.go
```diff
@@ -269,6 +271,23 @@ func (intro *IntroductionMessage) Handle(mc *gnet.MessageContext, daemon interfa
 
 		logger.Infof("%s verified for version %d", mc.Addr, intro.Version)
 
+		// Checks the genesis hash if not empty
+		if len(intro.Extra) > 0 {
+			var bcPubKey cipher.PubKey
+			if len(intro.Extra) != len(bcPubKey) {
```
olemis: IMO it would be nice to have a typed struct instead of an opaque `Extra byte[]` .

olemis:
The struct could look like this too . The point is that old implementation will decode an opaque byte slice , and the new one the typed fields added .

```
}

const (
	INTR_V2_MAGIC = sizeof(PubKey) + sizeof(int32)
)

type IntroductionMessage struct {
	// Mirror is a random value generated on client startup that is used
	// to identify self-connections
	Mirror uint32
	// Port is the port that this client is listening on
	Port uint16
	// Our client version
	Version int32
	c       *gnet.MessageContext `enc:"-"`
	// We validate the message in Handle() and cache the result for Process()
	valid bool `enc:"-"` // skip it during encoding
	// Struct size
	Reserved int32 // must be INTR_V2_MAGIC or the size of what's coming next should further info be added
	BlockchainPubKey cipher.PubKey
	// Reserved for future use
	Reserved byte[]
}
```

By doing this future use of `omitempty` annotation would be limited to quite a few exceptional cases .

p.s. notice that I removed `omitempty` tag in my previous message and reserved for future use comment above blockchain info struct.

src/daemon/messages.go
gz-c: Its transitory, next version will have a defined struct

gz-c:
@iketheadore test failure

src/skycoin/skycoin.go
```diff
@@ -315,6 +315,7 @@ func (c *Coin) configureDaemon() daemon.Config {
 	dc.Daemon.OutgoingMax = c.config.Node.MaxOutgoingConnections
 	dc.Daemon.DataDirectory = c.config.Node.DataDirectory
 	dc.Daemon.LogPings = !c.config.Node.DisablePingPong
+	dc.Daemon.BlockchainPubkey = cipher.MustPubKeyFromHex(c.config.Node.BlockchainPubkeyStr)
```
gz-c: Config.Node doesn't parse BlockchainPubkeyStr to BlockchainPubkey during its initialization?

src/daemon/daemon.go
```diff
@@ -53,6 +53,11 @@ var (
 	ErrDisconnectMaxDefaultConnectionReached                       = errors.New("Maximum number of default connections was reached")
 	// ErrDisconnectMaxOutgoingConnectionsReached is returned when connection pool size is greater than the maximum allowed
 	ErrDisconnectMaxOutgoingConnectionsReached gnet.DisconnectReason = errors.New("Maximum outgoing connections was reached")
+	// ErrDisconnectBlockchainPubkeyNotMatched is returned when the blockchain pubkey in introduction does not match
+	ErrDisconnectBlockchainPubkeyNotMatched gnet.DisconnectReason = errors.New("Blockchain pubkey in Introduction message is not matched ")
+	// ErrDisconnectInvalidExtraData is returned when extra filed can't be parsed as specific data type.
+	// e.g. ExtraData length in IntroductionMessage is not the same as cipher.PubKey
+	ErrDisconnectInvalidExtraData gnet.DisconnectReason = errors.New("Invalid extra data")
```
olemis: What about using the same error emitted on encoding / decoding error ? In the end incorrect `Extra` bytes would represent an encoding error .

iketheadore: What's the encoding / decoding error do you mean? It's not a decode error in either cipher decode nor gnet/message level.

olemis: IMO this should be detected by as a regular cipher decoding error . Pubkey is encoded as `Extra` bytes though . My comment is because of the fact that not finding a pubkey in `Extra` is caused by wrong encoding by the sending peer , which makes no difference with respect to a decoding error . Indeed if we had `omitempty` for struct types then the binary decoder would have caught this one .

gz-c: This is fine, it lets us disambiguate the problem if it occurs

gz-c:
I think the defined structs will look like 

```go
type IntroductionMessage struct {
	Mirror uint32
	Port uint16
	Version int32
	ExtraData []byte
}

type IntroductionMessageExtension struct {
	BlockchainPubkey cipher.Pubkey
}
```

And `ExtraData` would be parsed into the `Extension` struct.  This is necessary since the []byte field adds a length prefix, which would not be present if the struct was like:

```go
type IntroductionMessage struct {
	Mirror uint32
	Port uint16
	Version int32
	BlockchainPubkey cipher.Pubkey
}
```

src/daemon/messages.go
```diff
@@ -230,54 +230,73 @@ type IntroductionMessage struct {
...
+			d.Disconnect(mc.Addr, ErrDisconnectInvalidVersion)
 			return ErrDisconnectInvalidVersion
 		}
 
 		logger.Infof("%s verified for version %d", mc.Addr, intro.Version)
 
+		// Checks the genesis hash if not empty
+		if len(intro.Extra) > 0 {
+			var bcPubKey cipher.PubKey
+			if len(intro.Extra) != len(bcPubKey) {
```
gz-c: I think we want to ignore the length check for the ExtraData.  Otherwise, if we want to add another field later, we have to redo the 3-version upgrade again.  The length should meet the minimum, but if it is too long, we can ignore the extra data.

gz-c:
test failed

iketheadore:
I think the error would be fixed after the #1603 is done.

gz-c:
@iketheadore #1603 merged, can merge develop here now

PR #1607, opened 2018-05-31
Address copy button

### LINKED ISSUE #1575: The copy address button should be added back
The copy address button was moved some time ago to the options menu of the addresses. However, after some time the change has proven to negatively affects the UX, as several people have expressed, so the button should be added again.

mahansky:
Fixes #1575

Changes:
- brings back the old "copy" button next to wallet address


src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.html
```diff
@@ -3,7 +3,10 @@
     <div class="id-column">{{ i + 1 }}</div>
     <div class="address-column">
       <img src="../../../../../assets/img/qr-code-black.png" (click)="showQrCode($event, address.address)">
-      <p>{{ address.address }}</p>
+      <p class="click-to-copy" [ngClass]="{ copying: address.copying }"
+         (click)="copyAddress($event, address)" (mouseleave)="address.copying = false">
+        {{ address.address }} <span data-label="Copied" class="copy-label">Copy</span>
```
Senyoret1: Please, use ngx-translate for the button text and the animated confirmation text.

PR #1611, opened 2018-06-01
[coin] Fixed #1555 Remove logrus

### LINKED ISSUE #1555: Consider removing src/util/logging from package coin dependencies
The src/util/logging dependency in package coin causes logrus and its dependencies to be pulled in when using this package.

This interferes with gopherjs (https://github.com/skycoin/skycoin-lite/issues/9) and possibly other transpilation such as libskycoin.

Ideally, package coin has its dependencies minimized, this was always a goal of this package.

There are two kinds of logging statements in package coin:

* Panics
* Coin and coinhour math overflow warnings

The panics can be replace with `log.Panic` with no issue.

The overflow warnings are good to have, but possibly they could be moved to the caller.  If the returned errors are verbose enough, the caller can log the error when it occurs and it will have the same effect of notifying the operator of the warning.

stdevHan:
Fixes #1555 

Changes:
-Changed `logrus` in` coin` and replaced by `log.Panic` in case of` logger.Panic` and `log.Printf` in case of` logger.Warning`

No


olemis:
@stdevHan 👏

PR #1624, opened 2018-06-04
Advanced spend ui fixes

### LINKED ISSUE #1515: Send page button changes
Currently the send page button is "Verify"

Change this to "Preview" and "Send"

"Preview" will open the transaction preview dialog, and have a "Back" and "Send" button

"Send" will skip the preview dialog and send the transaction.

### LINKED ISSUE #1613: Problems with the UI after an error when trying to send a transaction
If there is an error while creating the transaction (in `SendFormComponent`) or while injecting it (in `SendVerifyComponent`), the error message is displayed in the snackbar. The problem is that after navigating to any other page the error message is still displayed. It should be hidden in `ngOnDestroy`.

To cause an error while creating a transaction, it is only necessary to write an invalid address in the form. To cause a test error while injecting a transaction you can modify the response of `WalletService.injectTransaction` to `return Observable.throw({_body: "Error"});`

This error could also be affecting the advanced form added in https://github.com/skycoin/skycoin/pull/1590

Additionally, when there is an error while injecting the transaction, the "Back" button remains deactivated. It should be reactivated after capturing the error.

### LINKED ISSUE #1618: Advanced spend, rename "Value" to "Coin Hours Share Factor"

### LINKED ISSUE #1619: Chage minimum share factor allowed by the slider to 0.1

### LINKED ISSUE #1620: Hide addresses with no balance from the advanced spend address dropdown

### LINKED ISSUE #1621: Advanced spend: improve display of multiple sending addresses
More than 2 selected addresses overflows the text box, it difficult to see which addresses are selected at a glance if you have a lot of addresses

### LINKED ISSUE #1622: Advanced spend: change address option should be de-emphasized
Change address is chosen automatically by default, and is not required.  This should be indicated in UI.  

Additional possibilities include:

* Bury the change address box under the dropdown
*  Place the change address box at the end of the list, below the destinations

mahansky:
Fixes #1622 
Fixes #1621 
Fixes #1620
Fixes #1619
Fixes #1618
Fixes #1613
Fixes #1515

Changes:
- multiple addresses in select are separated by bullet "•", removed coins/hours from there to make it more readable
- addresses with 0 balance are not shown in the dropdown
- share factor slider starts at 0.1
- fixed problems with snackbars not dismissing
- both simple and advanced forms now have "Preview" and "Send" button


Senyoret1:
I tested the changes and all the related issues were solved. There are only a few things that it might be good to consider:

- To solve https://github.com/skycoin/skycoin/issues/1621, now the addresses occupy less space, so now it is necessary to add more addresses to overflow the select box. However, it is still possible to overflow it, so it might be better to remove the `nowrap` from the styles and display the content in several lines when necessary. This would allow to continue displaying the amount of coins and hours, which may be useful.

- In my opinion, it should not be possible to send coins simply by pressing a button on the form, the user should be asked for confirmation. In fact, during the tests I pressed by mistake the send button while I wanted to see the preview. This could be solved with a modal confirmation window, but in that case there would not be much difference in speed of use compared to pressing the preview button. My point of view is about preferring the security of the funds rather than the speed of use, but I do not know what you think.

- In the simple form the user can send the transaction simply by pressing enter while the "Amount" field has the focus. To maintain consistency, it should be possible to do the same in the advanced form while the "Change address" field has the focus. However, this depends on the previous point, maybe it would be better if pressing enter shows the preview instead of sending the transaction.

- The fact that the "Change address" is optional could be clearer. One option would be to change its name to something like "Custom change address" and indicate in the help text that if an address is not entered the system will automatically choose one. Also, it probably would be good to remove the part that says "(in form of coin hours)" from the help text, since, technically, the change also includes coins.

- The "Destinations" help text says "coin hours" without taking into account if the "Automatic coin hours allocation" checkbox is selected or not.

mahansky:
I would leave the addresses the way they are. It's really simple to select/deselect others, when you do, you see your selection in a clear way. After that, it acts just like a confirmation that your selection is really there.

Confirmation modal could be a good security feature.
It could be done a bit simpler with a checkbox to preview tx that would be checked by default and a single Send button (or something similar but nicer)

Senyoret1:
That may be an option, but maybe simply leaving the preview screen as a confirmation method would suffice. I do not know what @gz-c thinks about it.

gz-c:
When there are a lot of addresses it is important to show which ones are selected to spend from clearly, overflowing will hide the choices. Chosen addresses need to all be visible at the same time

mahansky:
Addresses are displayed in rows, so each one of them is always visible.
Pressing enter shows preview.
Changed the help descriptions.
No changes with the send button for now.

PR #1627, opened 2018-06-05
Toggle local and global pending txs

### LINKED ISSUE #1375: Ability to toggle between showing all and user's pending txs
Add toggle button to pending transactions so user can choose to see all pending txs or just his own.

mahansky:
Fixes #1375

Changes:
- added toggle to switch between showing "my" and "all" pending transactions

### DIFF
--- a/src/gui/static/src/app/components/layout/header/nav-bar/nav-bar.component.html
+++ b/src/gui/static/src/app/components/layout/header/nav-bar/nav-bar.component.html
@@ -20,8 +20,8 @@
     <app-double-button
       *ngIf="navbarService.switchVisible"
       class="-switch"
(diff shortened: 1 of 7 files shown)


Senyoret1:
Works correctly. However, don't you think that the "My" and "All" buttons could be confusing for the users? The page will almost always be empty and when pressing the 2 buttons to see what happens the users will not notice any changes. Maybe it would be more obvious if the "Me" text is changed to "Mine" or if a tooltip is added indicating what each button does.

mahansky:
I think that tooltip saying something like "Only my pending transactions" is just unnecessary.
I changed "my" to "mine" as you said, that's the correct word in this case.

PR #1639, opened 2018-06-07
[lib] Fixes #968 fixes #1191 fixes #1628 fixes #1661 - Added libc and tests for cipher and coin

stdevAlDen:
Fixes #1191 , fixes #1628 , fixes #968 , fixes #1661 .

Changes:
- Adds wrapper functions to create C library.
- Adds tests for cipher and coin
- Implements new error codes

Does this change need to mentioned in CHANGELOG.md?
- yes

PR #1649, opened 2018-06-09
[lib] fixes #1629 - Implement libskycoin error codes

### LINKED ISSUE #1629: [lib] Implement error codes in libskycoin 0.25-alpha
- [x] Lookup error code for a given error object .
- [x] Error code for invalid handle values
- [x] Translate common error objects onto error codes
- ~Correct behavior of `catchApiPanic`~ Cipher panics reorganization

olemis:
Fixes #1629 

Changes:
- Implement `SKY_BAD_HANDLE` for handle lookup errors.

- No

### DIFF
--- a/lib/cgo/libsky_error.go
+++ b/lib/cgo/libsky_error.go
@@ -1,15 +1,35 @@
 package main
 
+import (
+	"errors"
+)
+
 const (
+	// SKY_ERROR generic error condition
 	SKY_ERROR = 0xFFFFFFFF
-	SKY_OK    = 0
+	// SKY_BAD_HANDLE invalid handle argument
+	SKY_BAD_HANDLE = 0xFF000001
+	// SKY_OK error code is used to report success
+	SKY_OK = 0
+)
+
+var (
+	ErrorBadHandle = errors.New("Invalid or unknown handle value")
+	ErrorUnknown   = errors.New("Unexpected error")
+
+	errorToCodeMap = map[error]uint32{
+		ErrorBadHandle: SKY_BAD_HANDLE,
+		ErrorUnknown:   SKY_ERROR,
+	}
 )
 
 func libErrorCode(err error) uint32 {
 	if err == nil {
 		return SKY_OK
 	}
-	// TODO: Implement error codes
+	if errcode, isKnownError := errorToCodeMap[err]; isKnownError {
+		return errcode
+	}
 	return SKY_ERROR
 }
 
--- a/lib/cgo/api.webrpc.client.go
+++ b/lib/cgo/api.webrpc.client.go
@@ -35,7 +35,7 @@ func SKY_webrpc_Client_CSRF(_c C.WebRpcClient__Handle, _arg0 *C.GoString_) (____
 	}()
 	c, okc := lookupWebRpcClientHandle(_c)
 	if !okc {
-		____error_code = SKY_ERROR
+		____error_code = SKY_BAD_HANDLE
 		return
 	}
 	__arg0, ____return_err := c.CSRF()
--- a/lib/cgo/cli.create_rawtx.go
+++ b/lib/cgo/cli.create_rawtx.go
@@ -26,7 +26,7 @@ func SKY_cli_CreateRawTxFromWallet(_c C.WebRpcClient__Handle, _walletFile, _chgA
 	}()
 	c, okc := lookupWebRpcClientHandle(_c)
 	if !okc {
-		____error_code = SKY_ERROR
+		____error_code = SKY_BAD_HANDLE
 		return
 	}
 	walletFile := _walletFile
@@ -49,7 +49,7 @@ func SKY_cli_CreateRawTxFromAddress(_c C.WebRpcClient__Handle, _addr, _walletFil
 	}()
 	c, okc := lookupWebRpcClientHandle(_c)
 	if !okc {
-		____error_code = SKY_ERROR
+		____error_code = SKY_BAD_HANDLE
 		return
 	}
 	addr := _addr
@@ -73,12 +73,12 @@ func SKY_cli_CreateRawTx(_c C.WebRpcClient__Handle, _wlt C.Wallet__Handle, _inAd
 	}()
 	c, okc := lookupWebRpcClientHandle(_c)
 	if !okc {
-		____error_code = SKY_ERROR
+		____error_code = SKY_BAD_HANDLE
 		return
 	}
 	wlt, okwlt := lookupWalletHandle(_wlt)
 	if !okwlt {
-		____error_code = SKY_ERROR
+		____error_code = SKY_BAD_HANDLE
 		return
 	}
 	inAddrs := *(*[]string)(unsafe.Pointer(&_inAddrs))
--- a/lib/cgo/wallet.wallet.go
+++ b/lib/cgo/wallet.wallet.go
@@ -26,7 +26,7 @@ func SKY_wallet_NewWallet(_wltName string, _opts C.Options__Handle, _arg2 *C.Wal
 	wltName := _wltName
 	__opts, okopts := lookupOptionsHandle(_opts)
 	if !okopts {
-		____error_code = SKY_ERROR
+		____error_code = SKY_BAD_HANDLE
 		return
 	}
 	opts := *__opts
@@ -46,7 +46,7 @@ func SKY_wallet_Wallet_Lock(_w C.Wallet__Handle, _password []byte, _cryptoType s
 	}()
 	w, okw := lookupWalletHandle(_w)
 	if !okw {
-		____error_code = SKY_ERROR
+		____error_code = SKY_BAD_HANDLE
 		return
 	}
 	password := *(*[]byte)(unsafe.Pointer(&_password))
@@ -66,7 +66,7 @@ func SKY_wallet_Wallet_Unlock(_w C.Wallet__Handle, _password []byte, _arg1 *C.Wa
 	}()
 	w, okw := lookupWalletHandle(_w)
 	if !okw {
-		____error_code = SKY_ERROR
+		____error_code = SKY_BAD_HANDLE
 		return
 	}
 	password := *(*[]byte)(unsafe.Pointer(&_password))
@@ -101,7 +101,7 @@ func SKY_wallet_Wallet_Save(_w C.Wallet__Handle, _dir string) (____error_code ui
 	}()
 	w, okw := lookupWalletHandle(_w)
 	if !okw {
-		____error_code = SKY_ERROR
+		____error_code = SKY_BAD_HANDLE
 		return
 	}
 	dir := _dir
@@ -120,7 +120,7 @@ func SKY_wallet_Wallet_Validate(_w C.Wallet__Handle) (____error_code uint32) {
 	}()
 	w, okw := lookupWalletHandle(_w)
 	if !okw {
-		____error_code = SKY_ERROR
+		____error_code = SKY_BAD_HANDLE
 		return
 	}
 	____return_err := w.Validate()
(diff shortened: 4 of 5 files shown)


olemis:
@gz-c in here I'd like to implement the right behavior of `catchApiPanic` function . What shall be done?

olemis:
btw I'll start using `libc` label for everything related to `libskycoin`.

gz-c:
As I understand it, the reason for catching panics is due to the cipher library panicking on bad input in a few cases.  Ideally the caller would enforce correct input before allowing it to panic, rather than catching every panic.

olemis:
@gz-c if I understood correctly your comment means that current approach is not correctly .

What shall wee do then ? Halt the OS process on panic ?

olemis:
@gz-c the fundamental manners I see to react to this are :

- exit OS process , if panic is considered severe enough for caller not to be able to recover
- return error code so that the caller can handle the error condition smoothly
- let the panic error bubble up to the caller as an exception in the specific programming language (though C has no support for exceptions)

Built-in support for the third option is too complicated and there is a chance for implementing this from error codes in e.g. SWIG .

Exiting the OS process is too drastic , considering that . Error codes are not suitable for preserving integrity in light of severe errors . We could also discriminate what kinds of panic conditions may be returned as an error code vs halting . Does this make sense ? If this was the case next question that comes to my mind is whether we want to include information in the error codes to know whether an error was returned by a function , raised by a logger or detected at API boundaries in `catchApiPanic`.

gz-c:
* severe panics translate to exit()/abort()
* non-severe panics can be avoided by sanitizing the input prior to calling the method which raises the non-severe panic

This means:

* exit()/abort() on any panic
* wrap methods that may raise a non-severe panic, and check for invalid input prior to use and return an error

olemis:
@gz-c 👍

olemis:
@gz-c for the record this implies making a number of changes in the test suite to assert for crash instead of error code .

gz-c:
Using the approach I described, you cannot cause a panic in tests

olemis:
There are test cases in the go lang test suite that check for panic . How shall we translate them ?

gz-c:
Give me an example and I will demonstrate

olemis:
ok ... let's consider [TestTransactionSignInputs](https://github.com/skycoin/skycoin/blob/89ed9fe/src/coin/transactions_test.go#L240-L268) . There is an assertion for a panic condition [`require.Panics(t, func() { tx.SignInputs([]cipher.SecKey{s}) })`](https://github.com/skycoin/skycoin/blob/develop/src/coin/transactions_test.go#L253) . 

Third party apps written in golang and making use of Skycoin code base have a chance for recovering of the corresponding panic should it ever be raised . Doing so through API boundaries is either impossible or not trivial .

How should this example be translated to make it part of `libskycoin` API test suite ?

gz-c:
@olemis In that case, its not intended for the application to recover.  It is intended that the develop fix the code to not provide a number of signing keys that does not match the number of inputs.

olemis:
considering [this comment](https://github.com/skycoin/skycoin/pull/1649#issuecomment-396301405) what would be the differentiator between severe and non-severe panic ?

src/cipher/crypto.go
```diff
@@ -71,20 +119,27 @@ func MustPubKeyFromHex(s string) PubKey {
...
+		return PubKey{}, ErrInvalidPubKey
 	}
 	return NewPubKey(b), nil
 }
 
 // PubKeyFromSecKey recovers the public key for a secret key
 func PubKeyFromSecKey(seckey SecKey) PubKey {
 	if seckey == (SecKey{}) {
-		log.Panic("PubKeyFromSecKey, attempt to load null seckey, unsafe")
+		err := skyerrors.NewValueError(
```
gz-c: Don't linebreak args, there is no column limit

src/util/errors/errors.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (err ValueError) Error() string {
+	return fmt.Sprintf("Invalid value for '%s' : %s",
+		err.ParamName, err.ErrorData.Error())
+}
+
+// NewValueError instantiate value error from concrete error
+func NewValueError(err error, paramName string, paramValue interface{}) ValueError {
+	return ValueError{
+		err, paramName, paramValue,
```
gz-c: Named initializers are preferred e.g. `ValueError{ErrorData: err, ...}`

src/util/errors/errors.go
```diff
@@ -0,0 +1,42 @@
+package errors
```
gz-c: If this is only used in package `cipher`, move this code to that package. I don't expect this to be used in other packages

olemis: this is meant to be used by other packages e.g. `coin` , but those changes are being implemented in #1639 ... which will be ready to merge after adding this , btw.

src/cipher/crypto.go
olemis: I use vim inside a container . Long lines are clipped , so I try to stick to 80 chars column limit

src/util/errors/errors.go
olemis: [Test results](https://travis-ci.org/skycoin/skycoin/builds/412857210) after merging with [`stdevEclipse_t1191_test_libc_cipher_and_coin`](https://github.com/skycoin/skycoin/pull/1639) . Some of those involve `coin` transactions .

`[====] Synthesis: Tested: 143 | Passing: 132 | Failing: 11 | Crashing: 6`

src/util/errors/errors.go
olemis: changing ...

src/cipher/crypto.go
olemis: for instance  screen shot 2018-08-07 at 7 25 49 pm

stdevStark:
@gz-c done

src/util/errors/errors.go
gz-c: Remove it then, this is unnecessary

olemis: I forgot to mention something ... if this is moved on to cipher then there will be a circular import issue making impossible to compile `libskycoin` code .

olemis: I think I can work around this , maybe ... question is can `coin` import `cipher` ?

olemis:
```
$ grep -nr 'cipher' src/coin/ | wc -l
     151
```

olemis: this is feasible ... will do it this way in `cipher`.

olemis:
@gz-c requested changes implemented . These are the instances in which value errors are used. Please review

```sh
# git merge simelo/olemis_t16
Already up-to-date.
root@14cdcee6b98a:/go/src/github.com/skycoin/skycoin# grep -nr ValueError src/
src/cipher/crypto.go:100:		err := NewValueError(ErrInvalidLengthPubKey, "b", b)
src/cipher/crypto.go:132:		err := NewValueError(ErrPubKeyFromNullSecKey, "seckey", seckey)
src/cipher/crypto.go:138:		err := NewValueError(ErrPubKeyFromBadSecKey, "seckey", seckey)
src/cipher/crypto.go:181:		err := NewValueError(ErrInvalidLengthSecKey, "b", b)
src/cipher/crypto.go:242:		err := NewValueError(ErrECHDInvalidPubKey, "pub", pub)
src/cipher/crypto.go:248:		err := NewValueError(ErrECHDInvalidSecKey, "sec", sec)
src/cipher/crypto.go:266:		err := NewValueError(ErrInvalidLengthSig, "b", b)
src/cipher/crypto.go:281:		err := NewValueError(ErrInvalidLengthSig, "s", s)
src/cipher/hash.go:38:		err := NewValueError(
src/cipher/hash.go:66:		err := NewValueError(
src/cipher/errors.go:8:// ValueError an operation or function receives an argument
src/cipher/errors.go:10:type ValueError struct {
src/cipher/errors.go:16:func (err ValueError) Error() string {
src/cipher/errors.go:21:// NewValueError instantiate value error from concrete error
src/cipher/errors.go:22:func NewValueError(err error, paramName string, paramValue interface{}) ValueError {
src/cipher/errors.go:23:	return ValueError{
src/cipher/errors.go:30:// NewValueErrorFromString shortcut for new value errors given an error message
src/cipher/errors.go:31:func NewValueErrorFromString(errorMessage string, paramName string, paramValue interface{}) ValueError {
src/cipher/errors.go:32:	return NewValueError(errors.New(errorMessage), paramName, paramValue)
src/cipher/errors.go:37:	if _err, isOfType := err1.(ValueError); isOfType {
src/cipher/errors.go:40:	if _err, isOfType := err2.(ValueError); isOfType {
```

src/util/errors/errors.go
gz-c: Just remove it, its not necessary

lib/cgo/libsky_error.go
```diff
@@ -26,9 +613,34 @@ func catchApiPanic(errcode uint32, err interface{}) uint32 {
...
 	if err != nil {
-		// TODO: Fix to be like retVal = libErrorCode(err)
-		return SKY_ERROR
+		if valueErr, isValueError := err.(cipher.ValueError); isValueError {
+			return libErrorCode(valueErr.ErrorData)
+		} else {
+			// Setting flag every time (i.e. even when haltOnPanic is active
+			// protects against hypothetical situations in which panic()
+			// does not abort the current process.
+			isAPILocked = true
```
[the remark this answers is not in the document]
olemis: I'll add a lock .

src/wallet/wallet.go
```diff
@@ -198,69 +236,69 @@ func (c CreateTransactionParams) Validate() error {
...
 		one := decimal.New(1, 0)
 		if c.HoursSelection.ShareFactor.LessThan(zero) || c.HoursSelection.ShareFactor.GreaterThan(one) {
-			return NewError(errors.New("HoursSelection.ShareFactor must be >= 0 and <= 1"))
+			return ErrShareFactorOutOfRange
 		}
 	}
 
 	if len(c.Wallet.UxOuts) != 0 && len(c.Wallet.Addresses) != 0 {
-		return NewError(errors.New("Wallet.UxOuts and Wallet.Addresses cannot be combined"))
+		return ErrWalletConstraint
```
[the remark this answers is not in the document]
olemis: API users will know how to react to an error condition only if different codes are returned back to them . All instances could be just mapped onto `SKY_ERROR` but then app / library code would have a hard time figuring out what happened and how to react accordingly to fix it .

src/cipher/crypto.go
```diff
@@ -447,7 +507,7 @@ func TestSecKeyHash(seckey SecKey, hash SHA256) error {
 		return fmt.Errorf("impossible error, TestSecKey, pubkey recovery from signature failed: %v", err)
 	}
 	if pubkey != recoveredPubkey {
-		return errors.New("impossible error TestSecKey, pubkey does not match recovered pubkey")
+		return ErrPubKeyFromSecKeyMissmatch
```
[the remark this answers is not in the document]
olemis: oops ...

lib/cgo/libsky_error.go
```diff
@@ -26,9 +613,34 @@ func catchApiPanic(errcode uint32, err interface{}) uint32 {
...
 		return errcode
 	}
+	if isAPILocked || err == ErrorLockApi {
+		isAPILocked = true
+		return SKY_API_LOCKED
+	}
 	if err != nil {
-		// TODO: Fix to be like retVal = libErrorCode(err)
-		return SKY_ERROR
+		if valueErr, isValueError := err.(cipher.ValueError); isValueError {
```
olemis: @gz-c this is what `ValueError` is for .

olemis:
@gz-c if `cipher-panics` branch is ready for merge , this one should be too .

PR #1669, opened 2018-06-22
[cmd] Fixed #1329 create a command for the IP monitors

### LINKED ISSUE #1329: Tool to monitor peers.txt IP addresses
Create a tool (either a new cli command or a new `cmd/` binary) that takes in a list of peers (ip:ports, newline separated, skipping comments and empty lines (a function exists somewhere in the `daemon` or `pex` package to parse this format)).  The tool connects to each of the peers, waits for the introduction packet (or times out) and produces a report with the status of the peer (unreachable, reachable, sent_introduction, introduction_parameters).  Introduction_parameters were added in v0.25.0 so will be absent for earlier peer versions.

The initial purpose is to check the status of peers in https://downloads.skycoin.net/blockchain/peers.txt

stdevHan:
Fixes #1329

Changes:
-Create a new `monitor` command with two flags,
 `-f` to give the address the file `peers.txt` that should be checked, in case of not calling it the default value is `skycoin/skycoin/peers.txt`
 `-t` to define the maximum waiting time in seconds for checking the IP, the default value is 5.

Yes


cmd/monitor/monitor.go
```diff
@@ -0,0 +1,201 @@
...
+		peers[pex] = bufferpex
+
+	}
+
+	return output, peers
+
+}
+func main() {
+
+	urlPeer := flag.String("f", urlPeers, "Url the file peer.txt")
```
olemis: You should have a configuration struct with fields initially set to default values and populated by `flag` afterwards. Parsing should happen inside a method (receiver) named `ParseConfig` .

gz-c: Thats not necessary for a simple cli

olemis: 👍

olemis: @gz-c I'll approve changes then . Please review .

olemis: Oh ! Noticed something ...

cmd/monitor/monitor.go
```diff
@@ -0,0 +1,201 @@
...
+	"strings"
+	"time"
+
+	"github.com/skycoin/skycoin/src/util/logging"
+)
+
+var (
+	logger = logging.MustGetLogger("monitor")
+	// ErrPeerlistFull is returned when the Pex is at a maximum
+	ErrPeerlistFull = errors.New("Peer list full")
```
olemis: Is this used somewhere in command source code ?

cmd/monitor/monitor.go
```diff
@@ -0,0 +1,201 @@
...
+	// ErrInvalidAddress is returned when an address appears malformed
+	ErrInvalidAddress = errors.New("Invalid address")
```
olemis: Isn't it possible to reuse the error objects defined in `src/` folder ? e.g.

```sh
$ grep -nr 'ErrInvalidAddress = ' src
src/daemon/pex/pex.go:44:	ErrInvalidAddress = errors.New("Invalid address")
```

cmd/monitor/monitor.go
```diff
@@ -0,0 +1,201 @@
...
+	ErrBlacklistedAddress = errors.New("Blacklisted address")
+
+	whitespaceFilter = regexp.MustCompile(`\s`)
+
+	dialTimeout = 5 * time.Second
+	urlPeers    = os.Getenv("GOPATH") + "/src/github.com/skycoin/skycoin/peers.txt"
+)
+
+// validateAddress returns a sanitized address if valid, otherwise an error
+func validateAddress(ipPort string) (string, error) {
```
olemis: if this function is already defined elsewhere in `src/` isn't it better to export it for reuse ?

```
$ grep -nr 'func validateAddress' src/
src/daemon/pex/pex.go:64:func validateAddress(ipPort string, allowLocalhost bool) (string, error) {
```

cmd/monitor/monitor.go
```diff
@@ -0,0 +1,201 @@
...
+
+	output, outputPeers := getPeers()
+
+	outputJSON, err := json.MarshalIndent(outputPeers, "", "     ")
+	if err != nil {
+		fmt.Println("Error formating wallet to JSON. Error:", err)
+		os.Exit(1)
+	}
+
+	fmt.Println(string(outputJSON))
```
therealssj: use log

cmd/monitor/monitor.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+var (
+	logger      = logging.MustGetLogger("monitor")
+	dialTimeout = 5 * time.Second
+	urlPeers    = getPathGo() + "/src/github.com/skycoin/skycoin/peers.txt"
+)
+
+func getPathGo() string {
+	gopath := os.Getenv("GOPATH")
```
olemis: Use `os.LookupEnv` instead of `os.GetEnv`

gz-c:
#2268

PR #1672, opened 2018-06-25
Fix history db resetting bug

### LINKED ISSUE #1664: verifyCreatedTransactionInvariants should check for genesis block (i.BkSeq > 0)
If blockchain contains only genesis block with single transaction, Visor.CreateTransaction will return error.

It seems, that the reason is the following check inside wallet.verifyCreatedTransactionInvariants https://github.com/skycoin/skycoin/blob/08bed793896013152869cb86786d9d2c35777b8a/src/wallet/wallet.go#L1375-L1377

coin.CreateUnspetns and coin.CreateUnspent will assign SrcTransaction non-null hash, only if provided block has BkSeq > 0.

https://github.com/skycoin/skycoin/blob/08bed793896013152869cb86786d9d2c35777b8a/src/coin/block.go#L203-L208

https://github.com/skycoin/skycoin/blob/08bed793896013152869cb86786d9d2c35777b8a/src/coin/block.go#L228-L236

iketheadore:
Historydb would be reset if node restarts with only genesis block

Fixes #1664 

Changes:
- Do not reset history db if `parsed height` is `0`, as the genesis block seq is `0`.

No.

### DIFF
--- a/src/visor/historydb/historydb.go
+++ b/src/visor/historydb/historydb.go
@@ -49,9 +49,10 @@ func New() *HistoryDB {
 // If we have a new added bucket, we need to reset to parse
 // blockchain again to get the new bucket filled.
 func (hd *HistoryDB) NeedsReset(tx *dbutil.Tx) (bool, error) {
-	if height, ok, err := hd.historyMeta.ParsedHeight(tx); err != nil {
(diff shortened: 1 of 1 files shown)


gz-c:
@iketheadore Can you rename ParsedHeight / SetParsedHeight to ParsedBlockSeq

PR #1676, opened 2018-06-26
Fix a bug in cipher-testdata generator and update the testdata

### LINKED ISSUE #1640: Repeated seeds in the cipher test suite
All the seed-{number}.golden files in https://github.com/skycoin/skycoin/tree/develop/src/cipher/testsuite/testdata have the same seed. This seems to be an error, since the cipher test data generator have the following comment:

https://github.com/skycoin/skycoin/blob/798b881651ad86ed29809d082fddd2688c3fe9a9/cmd/cipher-testdata/cipher-testdata.go#L31-L37

iketheadore:
Fixes #1640 

https://github.com/skycoin/skycoin/blob/ebfab5447157123159aec83e39f279e05f0fe34f/cmd/cipher-testdata/cipher-testdata.go#L152-L160
The loop variable `j` is captured by func literal, we should pass it as argument to the func.

Changes:
- Fix a bug in cmd/cipher-testdata and update testdata in src/cipher/testsuite/testdata. 

No


gz-c:
@Senyoret1

PR #1686, opened 2018-07-01
Integrating usage of the hardware wallet as a feature of the skycoin web app

mpsido:


mpsido:
The remaining important errors are these 2:

1) /usr/include/features.h:374:25: fatal error: sys/cdefs.h: No such file or directory
 #  include <sys/cdefs.h>
require that travis install the proper libc package : libc6-dev-i386 according to this post https://askubuntu.com/questions/470796/fatal-error-sys-cdefs-h-no-such-file-or-directory

2) LIBUSB_ERROR_OTHER 
that is an error with the package trying to connect the hardware device. I don't know what causes it (do not happen on my machine) it but I believe the travis machine need to install something like libusb.

mpsido:
tryed to add libc6-dev-i386 in travis.yml but it did not work
* tryed as a package
* tryed in apt-get sh command as well

gz-c:
`/usr/include/features.h` what is using this?  The code in src/ should not use cgo

mpsido:
I added a new file under usbhid that declares empty function if we are trying to build on the unsupported platform ios. 
As for the automatic tests functions I added they had been programmed to Skip the test if the device is not detected.

make test-amd64
I figured out the failure was caused by a "Panic" inside device-wallet module. It leads to an unexpected function terminaison instead of peacefully returning an error.
The last commit should fix it

mpsido:
If the issue got closed with no modification that means that instead of the last commit (that should be discarded) I should push my generated files in the branch.
DO YOU NEED THESE CHANGES ?

gz-c:
I don't understand the question

gz-c:
Can you describe the approach

mpsido:
@gz-c  are you still wondering my approach ?
I added a step in the .travis checks that is actually necessary everytime we switch branch without updating the generated files

mpsido:
@gz-c I have a bunch of commits for PIN code integration. but before I push them I wish you validate this pull request

mpsido:
@gz-c I fixed the travis build and I created an issue #1765 about PIN code integration. that can come in a new pull request if you approve this one

gz-c:
Weren't you going to remove the changes to the go code in the AP, wallet and coin packages?

mpsido:
I removed the dependencies but I still need some code down there.

Senyoret1:
I was taking a look at the wallet emulator and this PR (I really did not check the Go code in depth) to be aware of future changes that might be necessary in the front-end. I found something in the cli that may require changes, but I'm not sure, because I do not know how definitive are the methods that are being implemented.

console_

In the image you can see that the response of `emulatorSignMessage` says `address that issued the signature is:`, but it actually shows the signature of the message. Also, the `--address` parameter of `emulatorCheckMessageSignature` does not seem necessary, because its value does not seem to affect the response.

mpsido:
@Senyoret1 created an issue about this question: https://github.com/skycoin/services/issues/138

gz-c:
#1945 has been merged with frontend support

This needs to be cleaned up for the golang cli portions, or closed if the cli portions are in github.com/skycoin/hardware-wallet-go

PR #1690, opened 2018-07-02
[Makefile] Fixes #1614 replaced gometalinter with golangci-lint

### LINKED ISSUE #1614: Investigate replacing gometalinter with https://github.com/golangci/golangci-lint
https://github.com/golangci/golangci-lint

stdevHan:
Fixes #1614 

Changes:
-Changing from `gometalinter` to `golangci-lint` in `Makefile`


gz-c:
Can you give a summary of the differences

stdevHan:
@gz-c In summary
-The report of errors gives you in colors and is friendlier at the time of reading.
-It is very strict in the mode of writing the code, keeping it clean and
readable, as in the case of `return nil` if variation in the code.
`example https://travis-ci.org/skycoin/skycoin/builds/400703096#L1840`
-It is more versatile when it comes to your configuration.
-Only you need to install it and automatically puts the dependencies without having to declare another line of code more to the `Makefile`.

gz-c:
@stdevHan does it find the same lint errors as gometalinter?

src/skycoin/config.go
```diff
@@ -391,6 +391,7 @@ func (n *NodeConfig) applyConfigMode(configMode string) {
 	}
 }
 
+// nolint
```
gz-c: Does it allow us to specify which linter to disable for this line?

gz-c: With gometalinter, we would do something like `nolint: unparam`.

Also, gometalinter allows us to emit a warning if a `nolint` directive is unnecessary. If this option is available in the new method, use it.

stdevHan: @gz-c yes, you can use the same also as gometalinter

gz-c: Does it have an option equivalent to `--warn-unmatched-nolint`?

src/skycoin/config.go
```diff
@@ -391,6 +391,7 @@ func (n *NodeConfig) applyConfigMode(configMode string) {
 	}
 }
 
+// nolint: unparam
```
gz-c: When the `nolint` directive is above the function, does it apply to the whole function?  Normally we put this directive at the end of the function signature line

src/skycoin/config.go
stdevHan: That does not need it, because if the directive you declared is not valid, simply omit the declaration `// nolint directive`.

gz-c: What is the directive for this `// nolint` clause?

gz-c:
Merge conflicts

PR #1693, opened 2018-07-04
Show all transactions in transaction history

### LINKED ISSUE #1673: [Wallet] history page is not updated
The history page in wallet is not updated after spending, no new items would be added even if the wallet is restarted.

mahansky:
Fixes #1673

Changes:
- transaction history page shows transactions between own addresses

Does this change need to be mentioned in CHANGELOG.md?
Yes

### DIFF
--- a/src/gui/static/src/app/services/wallet.service.ts
+++ b/src/gui/static/src/app/services/wallet.service.ts
@@ -177,37 +177,50 @@ export class WalletService {
   }
 
   transactions(): Observable<NormalTransaction[]> {
-    return this.allAddresses().filter(addresses => !!addresses.length).first().flatMap(addresses => {
+    return this.allAddresses().first().flatMap(addresses => {
       this.addresses = addresses;
 
       return Observable.forkJoin(addresses.map(address => this.apiService.getExplorerAddress(address)));
-    }).map(transactions => [].concat.apply([], transactions).sort((a, b) =>  b.timestamp - a.timestamp))
-      .map(transactions => transactions.reduce((array, item) => {
-        if (!array.find(trans => trans.txid === item.txid)) {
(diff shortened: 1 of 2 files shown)


src/gui/static/src/app/services/wallet.service.ts
```diff
@@ -177,37 +177,50 @@ export class WalletService {
...
 
-          if (!outgoing && this.addresses.find(address => output.dst === address.address)) {
-            transaction.addresses.push(output.dst);
-            transaction.balance = transaction.balance + parseFloat(output.coins);
-          }
+          return array;
+        }, [])
+        .sort((a, b) =>  b.timestamp - a.timestamp)
+        .map(transaction => {
+          const outgoing = this.addresses.some(address => transaction.inputs[0].owner === address.address);
```
Senyoret1: This is not a new problem added in this PR, because this line is practically the same as the one previously used, but maybe it would be good to fix it now.

This line considers that the transaction is outgoing if the first input came from one of the user's addresses, which is not  technically correct, even if it's almost always true. Although I think that at this moment it is not easy to do so, a transaction may have inputs from multiple wallets, so the user's address could be in the second position or even beyond. If this happens, `relevantOutputs` will be empty and the precautions taken in the line 208 will not fix it, causing undefined behavior.

To solve this, all inputs should be considered.

mahansky: I was thinking about it but didn't know how to recreate it, so just left it like that. But I fixed it to be sure.

Senyoret1: To test it, I did put this above:
`transaction.inputs = [{owner: "123", coins: 1}].concat(transaction.inputs);`

PR #1696, opened 2018-07-05
Add Content-Security-Policy support

### LINKED ISSUE #1426: Add Content-Security-Policy header to HTTP responses

iketheadore:
Fixes #1426 

Changes:
- Add Content-Security-Policy: script-src 'self' 127.0.0.1" header to http responses 
- Add `disable-csp` cli option and set as `false` by default

Yes.

### DIFF
--- a/src/daemon/gateway.go
+++ b/src/daemon/gateway.go
@@ -22,6 +22,7 @@ type GatewayConfig struct {
 	BufferSize      int
 	EnableWalletAPI bool
 	EnableGUI       bool
+	DisabledCSP     bool
 }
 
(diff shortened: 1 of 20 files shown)


src/daemon/gateway.go
```diff
@@ -22,6 +22,7 @@ type GatewayConfig struct {
 	BufferSize      int
 	EnableWalletAPI bool
 	EnableGUI       bool
+	DisabledCSP     bool
```
gz-c: `DisableCSP`

PR #1706, opened 2018-07-11
[pex] Fixed #710 Removing the oldest peers when the list is complete

### LINKED ISSUE #710: Remove old/outdated peers from peerlist when full
When the peerlist fills and we want to add more new peers, remove the oldest LastSeen peers to make room.

Only remove peers if they are older than some amount (different from the current old peer expiration, maybe 8 hours, instead of 1 week).

stdevHan:
Fixes #710 

Changes:
-Changes in the `AddPeer()` the `src/daemon/pex/pex.go `, to erase the oldest peers greater than `8 hours`, when the list is full


src/daemon/pex/pex.go
```diff
@@ -360,7 +360,23 @@ func (px *Pex) AddPeer(addr string) error {
...
 
 	if px.Config.Max > 0 && px.peerlist.len() >= px.Config.Max {
-		return ErrPeerlistFull
+
+		i := float64(0)
+		var pivot string
+		for index, p := range px.peerlist.peers {
+			t := time.Unix(p.LastSeen, 0)
+			hours := time.Since(t).Hours()
+			if hours >= float64(time.Hour*8) {
```
olemis: shouldn't this be just `8` ?

src/daemon/pex/pex.go
```diff
@@ -360,7 +360,23 @@ func (px *Pex) AddPeer(addr string) error {
 	}
 
 	if px.Config.Max > 0 && px.peerlist.len() >= px.Config.Max {
-		return ErrPeerlistFull
+
+		i := float64(0)
```
gz-c: This block of code should be rewritten to find the first peer that exceeds the time threshold and remove them.

The threshold should be either a configured property of `Pex` or a global const, and measured in seconds

gz-c:
Fixed in #1999

PR #1723, opened 2018-07-20
Fixes  #1671. Options for loading peers from file or url and for disabling peers

### LINKED ISSUE #1671: [Request] Feed peers
Currently there are two ways of telling a new node which initial trusted peers connect to:
1) Have a peers.txt file that can be pulled with a GET request. This file has the format address:port. Each line defines a peer.
2) Write a peers.txt that's actually a JSON encoded file into the data directory used by the node. 

If I'm not wrong, right now doesn't matter which way to use, the code will still add the initial hardcoded connections to the new ones, making it impossible to create a new non-localhost network thats isolated from the hardcoded connections.

I think that would be a nice adding to the CLI a way to load a initial set of peers from a given peers.txt that can be stored locally anywhere and which the node won't add the hardcoded ones to.

stdevEclipse:
Fixes #1671 

Changes:
- Added option to load peers from a file
- Added option to load peers from url.
- Added option to disable default peers.


stdevEclipse:
What do you mean how? Travis made the tests.

src/daemon/pex/pex.go
```diff
@@ -570,11 +624,13 @@ func backoffDownloadText(url string) (string, error) {
...
 }
 
-// parseRemotePeerList parses a remote peers.txt file
+// ParseRemotePeerList parses a remote peers.txt file
 // The peers list format is newline separated ip:port
 // Any lines that don't parse to an ip:port are skipped
 // Localhost ip:port addresses are ignored
-func parseRemotePeerList(body string) []string {
+//
+// FIXME : move to iputil
```
gz-c: It doesn't need to go into iputil since this is specific to pex

stdevEclipse:
@gz-c, added new commits taking your comments into account.
There are also new tests with -disable-trusted-peers and setting a custom peer list option.
Full tests for api endpoints are ran using these  options.

src/daemon/pex/pex.go
```diff
@@ -243,6 +252,10 @@ func New(cfg Config, defaultConns []string) (*Pex, error) {
 		go func() {
 			if err := pex.downloadPeers(); err != nil {
 				logger.Errorf("Failed to download peers list: %v", err)
+			} else {
```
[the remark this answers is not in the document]
stdevEclipse: Fixed

src/daemon/pex/pex.go
```diff
@@ -296,10 +309,10 @@ func (px *Pex) downloadPeers() error {
 		return err
 	}
 
-	peers := parseRemotePeerList(body)
+	peers := ParseRemotePeerList(body)
 	logger.Infof("Downloaded peers list from %s, got %d peers", px.Config.PeerListURL, len(peers))
 
-	n := px.AddPeers(peers)
+	n := px.AddTrustedPeers(peers)
```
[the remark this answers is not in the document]
stdevEclipse: Fixed

src/daemon/pex/pex.go
stdevEclipse: Fixed

src/daemon/pex/pex.go
```diff
@@ -218,9 +218,17 @@ func New(cfg Config, defaultConns []string) (*Pex, error) {
 
 	// Load peers from disk
 	if err := pex.load(); err != nil {
+		logger.Errorf("Error loading peers from file: %v", err)
 		return nil, err
 	}
 
+	// If no default connections then all connections are untrusted.
```
gz-c: It should read a config value, not rely upon a side effect of having the config value set in another file

stdevEclipse: Let's say a node is started with disable-trusted-peers. Then peers are loaded from cache. Right now these peers are set as not trusted. 
Should this be specified in a command line argument? 
I mean, is it necessary an argument to say if these loaded peers are trusted or not?

gz-c:
Replaced by #1796

PR #1755, opened 2018-08-02
Add error pages

### LINKED ISSUE #1572: Show error page in electron if app init failed
Building off of #1497,

if the node init fails after the browser has opened in electron, a blank page is shown, because the API shuts down.

The electron app should detect if the node had shut down and show an error page.

This will require electron to detect node failure, and the node may need to use a non-zero exit code if it doesn't already, and also add a special string to the error log line so that the electron app can parse it out. 

This also need a special static page bundled with electron that is not part of the app's normal GUI.

### LINKED ISSUE #1695: The screen stays blank after starting the wallet
Several users have had the problem. Apparently several things may affect the backend and then the problem occurs (see the sticky message in the Telegram support group). To have a good UX, it is necessary to make the UI react properly, instead of showing a white screen.

Apparently, the main problem is that the backend stops responding correctly to requests made to the api. Because of that, when trying to get the list of wallets, an uncatched error occurs and then the whole web app fails, The request is done in:

https://github.com/skycoin/skycoin/blob/0a6a127224d16a8c438d6777fdb2d307b391d26a/src/gui/static/src/app/services/wallet.service.ts#L227-L229

In fact, if the wallet is started with the backend turned off, the white screen is displayed, but if `ApiService.get` is modified so that calls to `/wallets` always returns a list of fake wallets, the web app loads. The response can be simulated with this code:

```
get(url, params = null, options = {}) {
  if (url != 'wallets') {
    return this.http.get(this.getUrl(url, params), this.returnRequestOptions(options))
      .map((res: any) => res.json())
      .catch((error: any) => Observable.throw(error || 'Server error'));
  } else {
    return Observable.of([
      {
          "meta": {
              "coin": "skycoin",
              "filename": "2017_11_25_e5fb.wlt",
              "label": "test",
              "type": "deterministic",
              "version": "0.2",
              "crypto_type": "",
              "timestamp": 1511640884,
              "encrypted": false
          },
          "entries": [
              {
                  "address": "8C5icxR9zdkYTZZTVV3cCX7QoK4EkLuK4p",
                  "public_key": "0316ff74a8004adf9c71fa99808ee34c3505ee73c5cf82aa301d17817da3ca33b1"
              },
              {
                  "address": "23A1EWMZopUFLCwtXMe2CU9xTCbi5Gth643",
                  "public_key": "02539528248a1a2c4f0b73233491103ca83b40249dac3ae9eee9a10b9f9debd9a3"
              }
          ]
      }
    ]);
  }
}
```

The result of using that code without the backend is this screen:
result

Showing a screen like that would be better than just showing a white screen, but it is not entirely adequate. Returning an empty array of wallets causes the web app to navigate to the wizard, which would not be entirely appropriate either. Probably the most appropriate solution, without requiring too many changes in the code, would be  to display a full-screen error message. Something like this could work:
msg1
msg2

With a screen like that it would be possible to give more information to the user. It would be also posible to add a link to a page on the website with more information, which could be updated periodically.

Senyoret1:
Fixes #1695 
Fixes #1572

Changes:
- Show an error message if there is an error with the node when starting the electron wallet. The error is identified as "Error # 1".
- Show an error message in the angular app if the node is not available when starting the wallet.  The error is identified as "Error # 2".

The error message looks like this:
error_msg

The error message within the Angular code is activated if the conditions described in https://github.com/skycoin/skycoin/issues/1695 are met.

The error message showed by electron is displayed basically if the node is terminated and the app window is still open (without discriminating if it ended correctly or with an error), if it is not possible to open a URL and if the node writes `daemon.Pool.Run failed error` to the console (it happens if there is already another node instance running). These three conditions worked correctly in the tests, but it would be good to test it in other operating systems and, if possible, reproducing the conditions in which the node is not started correctly (as was mentioned in https://github.com/skycoin/skycoin/issues/1572 )

No


gz-c:
Not sure if `"daemon.Pool.Run failed"` is reliable or sufficient to detect failure. There are many ways it could fail to start.

gz-c:
In the worst case, which has happened before but should not happen anymore, a subcomponent will fail to initialize but the application will not shut down.  We can add a marker in the log line for each possible subcomponent initialization failure, and this marker can be watched for.  

Although at this point if an error such as that occurs, the application should exit. It might not exit with a nonzero exit code, but it should, and you could detect that, although it will look the same as an unavailable application.

Senyoret1:
@mahansky You idea is very reasonable, because it helps to make the code more maintainable. I made the change by placing the files inside the `assets` folder, so it was not necessary to modify `package.json` to copy the files, as it already copies all the contents of the dist folder, but this also means that in order to see the error screen you will have to rebuild `dist` before recompiling the electron wallet. I also added, within the .html file, code to allow future translations to other languages. I think that at this moment we do not have an adequate page in the wiki to redirect the user, but it would be easy to add link after creating it.

About error 1, on my pc if you start a node with `run.sh` or `make run` and then open a production version of the electron wallet, the node of the electron wallet writes the `daemon.Pool.Run failed error` string (the `error` part is probably added in some other part of the code) to the console and the window stays blank (because tha back-end is not stoped but the api does not respond correctly). The error happens because the code tries to open a connection with a port already in use, but I do not know if the same thing happens in other operating systems, so it is not necessarily something that could be reproduced in all cases. I deleted `error` from the string, just in case.

@gz-c As you say, looking for `daemon.Pool.Run failed` does not seems sufficient to capture all the errors, I used it just because it is the only text that I have identified for an important error that does not turn off the backend.

With the changes made in this PR the error page also appears if the backend is terminated (regardless of whether it ended with 0 or another exit code), it is only shown if the UI window is still open and the message indicates that it is not possible to connect to the backend. Taking this into account, if the node is modified so that the execution is finished in case of error, the message would appear correctly. If, instead, the node is modified to write a specific maker in the console, the modification that would have to be made to show the error message would be very simple, so I think any option would be fine.

gz-c:
@Senyoret1 are you able to get a `daemon.Pool.Run failed` message without the backend shutting down?

Senyoret1:
@gz-c Yes, and when that happens the log starts to show lines this:
```
WARN [gnet]: Waited 8.0046969s while trying to write daemon.gnet.ConnectionPool.IsConnExist to the strand request channel
```

gz-c:
@Senyoret1 how do you cause that? that would be a bug

gz-c:
There are multiple reasons the node may not be available; I've never personally seen `daemon.Pool.Run failed`. We need a better method of detecting node unavailability; this probably requires some change to the logging output of the backend

Senyoret1:
I checked the logs again and I did not find a string that could be used to correctly detect if the node is not available. I thought maybe `ERROR [daemon]`, but that string appears in errors that are not critical. Making changes to the loging output would be good in this case.

Another option would be to make a connection to the node and show the error if there is no response in a certain amount of time or if the connection fails, but this could bring problems in slow computers, so I do not think it's a good solution.

gz-c:
@senyoret1 the error from running two skycoin instances should be:

```
[2018-08-21T10:54:20+08:00] INFO [main]: Opening database /Users/user/.skycoin/data.db
[2018-08-21T10:54:21+08:00] ERROR [main]: Database failed to open: Open boltdb failed, timeout. Is another skycoin instance running?
```

Are you sure you received that `daemon.Pool.Run failed` error when running `./run.sh` prior to opening the electron wallet?

gz-c:
@Senyoret1 The failure to shutdown on error is a bug and will be fixed.  Expect that the process will shutdown if there is an error (with an exit status of 1, if that is helpful)

gz-c:
#1793 should fix the shutdown problem you're having

Senyoret1:
@gz-c That error appears when running the node twice with `./run.sh`. However, if you open a production version of the Electron wallet first (like the one on the website) and then execute `./run.sh`, that error does not happen, because the production version and the dev version store the database in different folders, and then, at least on Windows, is when you get the `daemon.Pool.Run` failed error. The same thing happens if you first run the node with `./run.sh` and then opena prod version of the Electron wallet.

As the problem will no longer be present, on the last commit I deleted the code that looks for that error message and the error screen is displayed only if the node is closed while the UI is still visible.

gz-c:
Oh, which folders are used for the database file for each version on windows?  On OS X, the same folder `~/.skycoin/` is used.  That explains why it got past the database lock

Senyoret1:
dev is in `C:\.skycoin` and prod in `%HOMEPATH%\.skycoin` (shortcut for `C:\Users\<username>\.skycoin`)

PR #1773, opened 2018-08-11
[daemon] fixes #735 - Reply to INTR message with GIVP + RJCT on peer overflow

### LINKED ISSUE #735: Send peers to client before disconnecting due to full peers

stdevStark:
Fixes #735 

Changes:
- New `RJCT` message
- Reply to `INTR` message with `GIVP` + `RJCT` on peer overflow

Yes


src/daemon/messages.go
```diff
@@ -16,6 +17,13 @@ import (
 	"github.com/skycoin/skycoin/src/util/utc"
 )
 
+var (
+	// Every rejection message prefix must start with "RJC" prefix
+	rejectPrefix = [...]byte{82, 74, 67}
```
olemis: `RJCT` is a generic message to reject other messages of any kind . This is not needed anymore .

iketheadore: New message will break the connection with node that can't recognize the message, if the `RJCT` is not used anymore, we'd better remove it.

olemis: @iketheadore this is exactly the intention . Every node receiving a `RJCT` message has to disconnect from peer that sent it , which is consistent with current behavior as you just said .

olemis: `RJCT` can help with other situation I've noticed recently happening quite often . This message notifies of disconnection right away , so peer is removed of PEX connection pool immediately. Now we are not doing this , so many entries are kept to other peers that already disconnected but node is unaware of .

This is a typical log entry reflecting this (obfuscated text is IP:port pairs)

```
[2018-08-18T04:29:50Z] WARN [daemon]: Failed to send *daemon.IntroductionMessage to ###.###.###.###:#####: set tcp ###.###.###.###:####: use of closed network connection
```

This solution will fix that since both ends will be aware of disconnections , thus allowing for graceful disconnection.

stdevStark:
@olemis @gz-c I deleted `RJCT` packet but [Travis has new errors](https://travis-ci.org/simelo/skycoin/jobs/423761632#L1898-L1936) in files not changed by me . What to do ? I merged with `develop` and see errors there again.

gz-c:
@stdevStark new build failures seem to be due to an upgrade in golangci-lint. I will submit a PR to fix them in develop

gz-c:
After #1838 merges, can update and it should be fixed

olemis:
@gz-c cool 👍 , it seems merged now

src/daemon/messages.go
```diff
@@ -352,7 +352,14 @@ func (intro *IntroductionMessage) Handle(mc *gnet.MessageContext, daemon interfa
 // Process an event queued by Handle()
 func (intro *IntroductionMessage) Process(d Daemoner) {
 	d.RemoveFromExpectingIntroductions(intro.c.Addr)
-	if !intro.valid {
+	if intro.validationError != nil {
+		if intro.validationError == pex.ErrPeerlistFull {
+			peers := d.RandomExchangeable(d.PexConfig().ReplyCount)
+			givpMsg := NewGivePeersMessage(peers)
+			if err := d.SendMessage(intro.c.Addr, givpMsg); err != nil {
```
olemis: @stdevStark I think I fixed the lint error , but still I am not quite sure PR is complete. Peer should be disconnected after this and it seems to me that such a thing is not done .

olemis: ... after sending the whole message to the peer , not immediately after invoking the function .

stdevStark:
ouch ! sorry , my mistake

PR #1797, opened 2018-08-22
[api] [util] #1787 - Skycoin REST API sets

### LINKED ISSUE #1787: Option to expose only the /health API endpoint
This is needed for monitoring

stdevStark:
Fixes #1787 

Changes:
- Web interface methods can be tagged using API set labels (many-to-many)
- API sets:
  * `DEFAULT` : Methods available when nodes run without parameters
  * `UX` : Methods for uxouts
  * `TX` : Methods for transactions
  * `STATS` : Methods to be consumed by external monitoring services
  * `BLOCKCHAIN` : Methods for information in blockchain
  * `EXPLORER` : Methods consumed by Skycoin explorer
  * `UX` : Methods for uxouts
  * `WALLET` : Wallet API
  * `SEED` : Wallet seed API
- `-enable-apiset` CLI argument to enable / disable API sets
  * String set CLI flags
  * Consistent behavior when used together with `-enable-wallet-api`
  

yes


gz-c:
@olemis what is this for

olemis:
This is for #1787 . Exposing `/health` API and endpoints needed for monitoring will happen by setting `-enable-apiset STATS` as CLI arg.

stdevStark:
@olemis what happens with [these build errors](https://travis-ci.org/simelo/skycoin/jobs/422822234#L1896-L1902)

gz-c:
@olemis why do we need a general solution when we only care to enable `/health`?

gz-c:
I recommend `ALL`, `WALLET`, `WALLET_SEED`, `READ_ONLY`, `STATS`

`READ` would be everything not covered by `WALLET`, `WALLET_SEED` or `STATS`.

There is no need for `DEFAULT` since we can set the default based upon these other parameters.

`UX`, `TX`, `BLOCKCHAIN`, `EXPLORER` are unnecessary and would be covered by `READ_ONLY`

stdevStark:
@gz-c what methods in `ALL` ?

stdevStark:
Changes done after @gz-c [comment](https://github.com/skycoin/skycoin/pull/1797#issuecomment-417829439).

olemis:
@gz-c I think these changes have been applied

gz-c:
@stdevStark `ALL` is a shortcut for enabling all endpoints.

Don't remove the existing CLI options, let them modify the values in the enable API set.  We can deprecate them in this release and remove them in the next one.

stdevStark:
@gz-c I did it in the way you suggest . Fixing Travis .

src/api/http.go
```diff
@@ -254,6 +267,18 @@ func newServerMux(c muxConfig, gateway Gatewayer, csrfStore *CSRFStore, rpc *web
 		return handler
 	}
 
+	forAPISet := func(hf http.HandlerFunc, mainAPIName string, otherAPINames ...string) http.HandlerFunc {
```
gz-c: use `apiName []string`

src/api/http.go
```diff
@@ -350,121 +375,121 @@ func newServerMux(c muxConfig, gateway Gatewayer, csrfStore *CSRFStore, rpc *web
...
 	//     entropy: entropy bitsize.
-	webHandlerV1("/wallet/newSeed", newWalletSeed(gateway))
+	webHandlerV1("/wallet/newSeed", forAPISet(newWalletSeed(gateway), APIWallet))
 
 	// Gets seed of wallet of given id
 	// GET Arguments:
 	//     id: wallet id
 	//     password: wallet password
-	webHandlerV1("/wallet/seed", walletSeedHandler(gateway))
+	webHandlerV1("/wallet/seed", forAPISet(walletSeedHandler(gateway), APISeed)) // FIXME: APIWallet?
```
gz-c: What was the old behaviour? Did `-enable-seed-api`  require `-enabled-wallet-api`?

src/daemon/gateway.go
```diff
@@ -18,17 +19,22 @@ import (
 	"github.com/skycoin/skycoin/src/visor/historydb"
 )
 
+const (
```
gz-c: This should be defined in api/http.go with the others

src/daemon/gateway.go
```diff
@@ -18,17 +19,22 @@ import (
...
+	// AllAPISets special API set containing all exported methods
+	AllAPISets = "ALL"
+)
+
 // GatewayConfig configuration set of gateway.
 type GatewayConfig struct {
-	BufferSize      int
-	EnableWalletAPI bool
+	BufferSize     int
+	EnabledAPISets collections.StringSet
```
gz-c: Do not change anything in gateway. Handle it all in `api/`.  In skycoin/skycoin.go or skycoin/config.go, during config processing, you can set the gateway.EnableWalletAPI flag from the API set data

src/api/http.go
olemis: Both options are independent , but this handler checks for both flags set .

src/daemon/gateway.go
olemis: done

src/daemon/gateway.go
olemis: This is used by `Gateway` instances to query for wallet API enabled e.g. [for wallet balance](https://github.com/skycoin/skycoin/blob/develop/src/daemon/gateway.go#L952-L954). Not sure it can be completely removed since methods are (re)used at multiple locations.

olemis:
@gz-c Travis succeeded . Could you please review ?

src/util/collections/stringset.go
```diff
@@ -0,0 +1,76 @@
+package collections
```
gz-c: Don't make this a general util nor a public type unless you also make it threadsafe

src/daemon/gateway.go
```diff
@@ -9,23 +9,29 @@ import (
...
 }
 
 // NewGatewayConfig create and init an GatewayConfig
 func NewGatewayConfig() GatewayConfig {
 	return GatewayConfig{
-		BufferSize:      32,
-		EnableWalletAPI: false,
+		BufferSize:     32,
+		EnabledAPISets: collections.NewStringSet(),
```
gz-c: There is no need to change anything in this file

PR #1871, opened 2018-09-10
Gateway cleanup

gz-c:
Changes:

- Add `-enable-spend-api` to enable `/api/v1/wallet/spend`, which is deprecated. It is off by default.
- Add note to changelog of what will be removed next version. `/api/v1/wallet/spend` and the webrpc interface will be removed in v0.26.0.
- If the spend API is disabled, daemon will not use `strand()` and all requests will run concurrently
- Perform DB operations inside of a wallet service read lock if the DB operation requires wallet data.  This is probably not necessary, but added in case it would be problem not to have.
- Add `wallet.Service.{View,ViewSecrets,Update,UpdateSecrets}` to enable safe wallet access from visor using the lock.  Only `View` is used.  `UpdateSecrets` could potentially be used if we generate a new address for change each time.
- Check that password is not provided if the wallet is not encrypted when generating new addresses
- When generating new addresses, don't update the wallet in memory if the wallet save failed

Yes

Notes on the wallet+visor locking:

Wallets are considered a subset of the Visor database.  They are equivalent to a second database.  Some operations require querying both wallets and the database.  To protect these, the wallet service (database) lock is used as the initial lock, before performing a bolt.DB operation (which also locks itself).  The wallet lock is done first because wallet operations can be very slow if the wallet is encrypted (although currently this is not used in the relevant methods), and because it is judged less of a nuisance if the wallet APIs are unduly delayed by slow bolt.DB operations than if bolt.DB operations are delayed by slow wallet operations.

One thing to note is that it is never safe to modify the wallet and bolt.DB data in the same operation, because there is no safe way to perform a rollback across both.

The existing methods which need both wallet and the bolt.DB only do a read of each.  I can't think of any reason to have a wallet read and bolt.DB write together.  If we generate a new change address for spending automatically, we would need to do a wallet write + bolt.DB read together.  Currently, an API user could generate a new address for change first with `POST /api/v1/wallet/newAddress`, then use this address for `POST /api/v1/wallet/transaction` if they wanted this behavior, but that would require two slow decryptions in the current wallet management system.  However, this could also be written with one decryption and avoiding any race conditions between the data internally, so it may be fine to remove the wallet lock wrapper around certain DB operations if necessary to avoid performance issues.

The important part is that all visor and wallet operations must be analyzed for safety in race conditions.  Transactions must be used for all logical operations, unless otherwise guaranteed to be safe against using them, in which case it must be noted in the code.


gz-c:
@senyoret1 can you help me with this? the e2e tests are failing, so i tried to replicate it manually.  When I do "new address" for an unencrypted wallet, i get "Wallet is not encrypted" error, because the request is sending "undefined" as the password.  If you see an API response format was changed, let me know

PR #1881, opened 2018-09-12
Fix advanced form validation and coin values

Senyoret1:
Changes:
- This pr fixes an unusual problem with a validation in `SendFormAdvancedComponent`. If the user selects an address with 2.3 coins and tries to send to two addresses 2.1 and 0.2 coins respectively, the form validation should pass and allow the operation to be performed, but it does not. The problem is that the validation code when adding the two outputs obtains 2.3000000000000003 (due to JavaScript problems with decimals) and when comparing it with the number of coins of the exit address the validation fails.
- Additional changes in my next comment.

No


gz-c:
Is there no fixed precision decimal? Or some kind of int?

gz-c:
https://www.avioconsulting.com/blog/overcoming-javascript-numeric-precision-issues

Senyoret1:
I added a new commit in which bignumber.js (https://github.com/MikeMcl/bignumber.js/) is now used for the coin values. I did not modify the hour values because with the integers there are no problems (since they are not handled internally as fractions). With these new changes it would be good to review this pr with some care.

gz-c:
Hours are still parsed as floats by javascript so can have the similar errors

Senyoret1:
I added a new commit so that also the hours are calculated with bignumber.js

src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.html
```diff
@@ -20,7 +20,7 @@
 
 <div class="-records">
   <ng-container *ngFor="let address of (wallet ? wallet.addresses : []); let i = index">
-    <div class="-record" *ngIf="!wallet.hideEmpty || address.coins">
+    <div class="-record" *ngIf="!wallet.hideEmpty || address.coins > 0">
```
mahansky: this should be `isGreaterThan(0)`

Senyoret1:
You're right. I removed `ROUND_HALF_UP` (just the number 4 on the html files) to simplify the code.

The number formatting pipe was replaced with `decimalPlaces()`, which ensures that a maximum of 6 decimals is always shown for the coins and no decimal for the hours. Is there something else that you think needs formating?

mahansky:
You can chain it like this `decimalPlaces(5).toFormat().toString();` to produce `123,456.789` 
That's pretty much the same result as with `number:'1.0-6'` pipe

PR #1882, opened 2018-09-12
Select change address

### LINKED ISSUE #1803: Advanced spend UI allow change address to be selected from existing wallets
The UI for this is trickier, because we want provide a dropdown to allow the user to choose the change address from their wallets, but also allow them to enter in an address that is not in their wallets

Senyoret1:
Fixes #1803

Changes:
- Now it is possible to select an address from one of the wallets as the change address, in the advanced form. To select one of the addresses, it is necessary to press the "Select" link:
sel-link

After clicking on the link, it is possible to select the address using this dialog:
sel-dialog

- The help text of the "Custom change address" field was modified to indicate that it is possible to select the address by clicking on the "Select" link.


Senyoret1:
I made the change. Now the names can only have one line and the full name can be seen in the tooltip:
long-name

PR #1884, opened 2018-09-13
Hardcode the version in coin.template, not fiber.toml

gz-c:
Changes:
- Hardcode the version in `coin.template`, instead of getting it through `fiber.toml`.  It's not part of configuration, and leaving it in `fiber.toml` adds an extra step to fiber users for upgrading.

No


gz-c:
@nutmix version/build params are removed from the toml config file, fiber users do not need to concern themselves with it

nutmix:
@gz-c great, ill update the handbook, and test it on the next coin creation.

PR #1893, opened 2018-09-18
Improved code for the wallet creation forms

### LINKED ISSUE #1300: The minimum length of the seeds is very short
When creating or loading a wallet, it is possible to manually write the seed. At this time, the form have validators requiring the seed to have at least one character. Maybe, for security, the validators should force the user to use a much larger number of characters, maybe 40.

Another option could be to disable the ability of the user for editing the seed when creating a new wallet.

Since the option for loading a wallet is, basically, the same as creating a new one, the validation should probably be when creating and when loading a wallet, but that could cause problems for anyone with a previously created wallet with a short seed, and it seems unlikely that a users would use the option to load a wallet just to create a new one. Because of this, the requirement for more characters could be active just when creating a new wallet.

If any of those changes is implemented, it must be incorporated into https://github.com/skycoin/skycoin/pull/1238 or a new pr after https://github.com/skycoin/skycoin/pull/1238 is merged.

### LINKED ISSUE #1523: Validate the seeds in the frontend
The situation about how to validate the seeds is complex, so this is for now just an idea, taking into account several important points discussed in the past.

Normal users create their wallets with seeds automatically generated by the system. Taking into account these users, it would be very convenient to validate all the seeds, both when creating and when loading the wallets. Without validation, these users could incorrectly enter the seed when loading a wallet and panic when they do not see their funds.

However, validating the seeds and blocking any seed that does not pass the validation would give problems to advanced users. These users may add additional data or even manually write seeds with a totally unknown format, in order to increase entropy.

A solution that can be implemented for both types of users is to validate the seeds and, if the validation fails, alert the user that the seed may be incorrect and ask him if he wishes to continue. To do so, the contents of the modal window for creating/loading a wallet could be replaced by the alert message, in a similar way to how the send coins form is replaced with the control that allows the user to confirm the details of the transaction. In that alert message, the user will be able to choose whether to continue or return, also as in the transactions validation control.

If this is going to be implemented, it is necessary to create a new api endpoint for validating the seeds, so it will be necessary to create a new issue and wait for it to be solved before working on this.

Senyoret1:
Fixes #1523 
Fixes #1300 

Changes:
- Now the seeds are checked before creating the wallets. The objective is to alert the user if the seed does not correspond to the characteristics of the seeds created automatically by the system, to avoid simple mistakes that could prevent recovering the wallet in the future.
- Now the form used in the modal window and in the wizard is in a reusable component, to avoid repeating code.

This pr is based on https://github.com/skycoin/skycoin-web/pull/447

No


gz-c:
There is a specific method for seed checking (the words would be from the list and there is a checksum built into the seed).  However we don't expose this through the API.

If the seed format is enforced, it breaks for some users, because some modify the generated seed, for example adding extra characters as a password, so I don't think we can enforce it.

gz-c:
Do you want an endpoint to validate the seed? The generated bip39 seeds have a specific method for validating

Senyoret1:
At this moment the code simply checks if the seed is composed of 12 or 24 words separated with one white space and without special characters or numbers. If an endpoint for verifying the seed is created, I think I can use it. However, the current method would continue to be used in the web wallet

gz-c:
@Senyoret1 the web wallet would also be able to use such an endpoint, we'd make it public (or it could be exposed through the skycoin-lite bindings). however this is fine for now, but won't catch typos

PR #1900, opened 2018-09-21
fixes #1897 - API sets with include / exclude filters

### LINKED ISSUE #1897: Change -enable-api-set to have exclude/include filters
If the API sets expand in name, having a single `-enable-api-set` flag will be burdensome.

Change this to have these flags (inspired by :

* `-enable-all-apisets`
* `-disable-all-apisets`
* `-enable-apiset=...`
* `-disable-apiset=...`

Errors: 

* `-enable-all-apisets` and `-disable-all-apisets` cannot be combined
* An API set tag cannot appear in both `-enable-apiset` and `-disable-apiset`

Order of operations:

Starting from the default configuration (which is different for each run mode):

* If `-enable-all-apisets`, turn on all API sets.
* If `-disable-all-apisets`, turn off all API sets.
* For each API set in `-disable-apiset`, disable that set.
* For each API set in `-enable-apiset`, enable that set.

The special `ALL` tag should be removed.

stdevStark:
Fixes #1897 

Changes:

- Command line options for enabling/disabling API sets `-enable-all-apisets`, `-disable-all-apisets`, `-enable-apiset`, `-disable-apiset`.

yes


src/api/explorer.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/skycoin/skycoin/src/cipher"
 	"github.com/skycoin/skycoin/src/coin"
 	"github.com/skycoin/skycoin/src/readable"
+	"github.com/skycoin/skycoin/src/util/collections"
```
gz-c: Remove `collections`

src/api/explorer.go
```diff
@@ -32,12 +33,8 @@ type CoinSupply struct {
...
 
 // newStringSet returns a map-based set for string lookup
-func newStringSet(keys []string) map[string]struct{} {
-	s := make(map[string]struct{}, len(keys))
-	for _, k := range keys {
-		s[k] = struct{}{}
-	}
-	return s
+func newStringSet(keys []string) collections.StringSet {
+	return collections.NewStringSet(keys...)
```
gz-c: Undo

src/api/http.go
```diff
@@ -44,6 +44,8 @@ const (
 	EndpointsWalletSeed = "WALLET_SEED"
 	// EndpointsDeprecatedWalletSpend endpoints implement the deprecated /api/v1/wallet/spend method
 	EndpointsDeprecatedWalletSpend = "DEPRECATED_WALLET_SPEND"
+	// EndpointsAll lists all of the above
+	EndpointsAll = "READ,STATUS,WALLET,WALLET_SEED,DEPRECATED_WALLET_SPEND"
```
gz-c: should be `[]string`, the comma separation is only relevant to the command line parser, this is the `api` package, although it does not belong in `api` since `api` does not use it, `all` is a command line parser idea

gz-c:
#1907

PR #1902, opened 2018-09-21
Add -host-whitelist option

### LINKED ISSUE #1901: Configurable acceptable Host header
Backend part of #1258 

The Host header check enforces that the Host header matches the web interface addr:port and also that it is a localhost address.

Add a whitelist option for possible Host headers

gz-c:
Fixes #1901 

Changes:
- Add `-host-whitelist` option to specify alternate allowed hosts when communicating with the API bound to a localhost interface

Yes


Senyoret1:
It works. If I try to make the connection I no longer receive 403 errors. However, when dealing with a different URL, the browser tries to send a Options request, instead of get, and the following error appears:
options
However, it is a different problem and I do not know if it really is important

gz-c:
I think that is CORS doing a preflight request, which we don't handle. I'll have to add CORS handling.

gz-c:
@Senyoret1 can you check that other endpoints don't work when `-host-whitelist` is not set? especially the wallet endpoints.  It should be getting CORS errors

Senyoret1:
I did some additional tests with Postman and all the endpoints returned this error:

The same error was shown even when trying to access non-existent URLs, such as `http://127.0.0.1:6420/api/v1/wallets/folderName12`

gz-c:
Ah it doesn't get to the CORS error because the explicit Host check overrides that. Looks fine

PR #1905, opened 2018-09-22
Use block head to validate cli createRawTx

gz-c:
Changes:
- Use block head from the API to do complete validation of the transaction in `cli createRawTx`
- Adds `"ux_hash"` to the block header in relevant API responses
- Adds the block header to the response of `GET /api/v1/outputs`

Yes


src/readable/block.go
```diff
@@ -1,29 +1,75 @@
...
-	BkSeq             uint64 `json:"seq"`
-	BlockHash         string `json:"block_hash"`
-	PreviousBlockHash string `json:"previous_block_hash"`
-	Time              uint64 `json:"timestamp"`
-	Fee               uint64 `json:"fee"`
-	Version           uint32 `json:"version"`
-	BodyHash          string `json:"tx_body_hash"`
+	BkSeq        uint64 `json:"seq"`
+	Hash         string `json:"block_hash"`          // TODO -- rename "hash"
+	PreviousHash string `json:"previous_block_hash"` // TODO -- rename "previous_hash"
```
therealssj: why leave this for TODO here?

gz-c: Took them off and added them to the list in #420

gz-c: It can't be changed without breaking the API

PR #1907, opened 2018-09-22
Add -enable-all-api-sets

### LINKED ISSUE #1897: Change -enable-api-set to have exclude/include filters

gz-c:
Fixes #1897 

Changes:
- Add `-enable-all-api-sets` and `-disable-api-sets`
- Rename `-enable-api-set` to `-enable-api-sets`
- `-enable-all-api-sets` will not enable deprecated or insecure endpoints (currently `DEPRECATED_WALLET_SPEND` and `INSECURE_WALLET_SEED`), these must be enable explicitly with an additional parameter of `-enable-api-set=`.
- `-enable-seed-api` and `-enable-wallet-api` were removed because it was not worth the effort in detecting whether or not the user explicitly passed `-enable-seed-api=false` or whether it had defaulted to that value. 

Note: I chose not to include `-disable-all-api-sets` since they are all disabled by default.  When I considered the options from the users standpoint, it didn't make sense to include.

Yes


ci-scripts/integration-test-disable-gui.sh
```diff
@@ -77,7 +77,8 @@ echo "starting $COIN node in background with http listener on $HOST"
             -launch-browser=false \
             -data-dir="$DATA_DIR" \
             -wallet-dir="$WALLET_DIR" \
-            -enable-api-set=ALL \
+            -enable-all-api-sets=true \
```
olemis: My first impression is that the fact that the option name is `--enable-all-api-sets` but not all API sets are enabled , might lead to confusion . `--enable-supported-api-sets` ?

gz-c: The API sets that it does not enable are special endpoints that wouldn't be enabled normally, `INSECURE_WALLET_SEED` is only intended for the desktop client, and for `DEPRECATED_WALLET_SPEND` they need to stop using it.

If it enabled all endpoints it would do the correct behavior for no user.  In no case, except for the integration tests, would anyone want to enable every endpoint including `INSECURE_WALLET_SEED` and `DEPRECATED_WALLET_SPEND`.

`-enable-supported-api-sets` doesn't provide any more clarity, because the meaning of `supported` is ambiguous.  I think any name cannot be expressive enough, so it is described in the doc string.

PR #1916, opened 2018-09-27
Add the reset password page

### LINKED ISSUE #1574: There is no option for resetting the password
If the user loses the password of a wallet, he should be able to reset it with the seed, as indicated here:
info

However, there is no option for reseting the password, nor is it possible to create a wallet with the same seed and another password, or delete an existing wallet. Due to this, if an user loses the password, the only option that remains is to manually delete the wallet file, something that is complicated for the less expert users.

Senyoret1:
Fixes #1574

Changes:
- Changes needed for using version 2 of the api
- Now the modal window in which the user must enter the password shows an "I forgot my password" link:
modal
- A page for reseting the password was added. The form was created in a page instead of a modal window to avoid possible complications in the code when having to update the status of the wallet after resetting the password.

No


gz-c:
Note that the v2 API might still return the old style errors e.g. "403 Unauthorized - something something" in some cases

Senyoret1:
That should not be a problem, because the same function is used to process the errors of v1 and v2

Issue #1957, opened 2018-10-13
Send disconnect packet before disconnecting from peer


gz-c:
When disconnecting from a peer, send a packet which indicates the disconnect reason.

It should be safe to send this new packet since we are disconnecting anyway.  The remote peer will disconnect when it receives an unknown packet.

olemis:
Related to #735 . Indeed , this would consist in sending `RJCT` message back to peers before disconnecting under reasons other than `INTR` mismatch .

PR #1959, opened 2018-10-13
[daemon] fixes #735 - Reply to INTR message with GIVP + RJCT on pex pool overflow

### LINKED ISSUE #735: Send peers to client before disconnecting due to full peers

olemis:
Fixes #735 

Changes: 

- New `RJCT` message
- Reply to `INTR` message with `GIVP` + `RJCT` on peer overflow

yes


src/daemon/daemon.go
```diff
@@ -1081,6 +1083,16 @@ func (dm *Daemon) handleMessageSendResult(r gnet.SendResult) {
 	switch r.Message.(type) {
 	case SendingTxnsMessage:
 		dm.announcedTxns.add(r.Message.(SendingTxnsMessage).GetFiltered())
+	case DisconnectPeerMessage:
```
olemis: @gz-c what do you think about this approach for synchronizing disconnection after `RejectMessage` is successfully sent? It's relatively simple . I plan to submit another patch for #1957 .

olemis: Open question : Should disconnection happen only on success or also on error

gz-c: I think Disconnect should always be called even if there was an error attached to the message.

This approach probably works, but you'd want to call disconnect directly on the pool, and change daemon.Disconnect to send a disconnect packet instead.

gz-c:
I wrote up a guide for running localhost nodes to test this out:

https://github.com/skycoin/skycoin/wiki/Running-multiple-nodes-locally-isolated-for-manual-testing

gz-c:
#1999

PR #1966, opened 2018-10-18
Add user agent to introduction packet

gz-c:
Changes:
- Add a user agent string to the introduction message

The user agent is appended after the blockchain pubkey in the extra data.  If the extra data is not empty, then the user agent is required and must be valid.

A `maxlen` on struct tags for the encoder is added. `maxlen` applies to strings or slices. It restricts the number of elements in the string or slice, not the length of the raw bytes.  `maxlen` is only checked when deserializing, so callers must ensure that they do not encode values that exceed `maxlen`.  This field was added to put a restriction on the size of the user agent sent in the packet.  While messages have a max length defined in `gnet.Config.MaxMessageLength`, this length is 256kb by default and is not specialized enough to cap a single field.

The user agent is documented here:
https://github.com/skycoin/skycoin/wiki/Wire-protocol-user-agent

Yes


src/daemon/messages.go
```diff
@@ -221,30 +223,47 @@ func (gpm *GivePeersMessage) Process(d Daemoner) {
...
 	Mirror uint32
 	// Port is the port that this client is listening on
 	Port uint16
 	// Protocol version
 	Version int32
 	c       *gnet.MessageContext `enc:"-"`
 	// We validate the message in Handle() and cache the result for Process()
-	valid bool `enc:"-"` // skip it during encoding
+	valid         bool           `enc:"-"` // skip it during encoding
+	userAgentData useragent.Data `enc:"-"`
```
olemis: Could it be possible to limit user agent ID to be of a fixed number of chars (e.g. similar to message prefix) ?

gz-c:
Added `user_agent` and `coin` to `/health`

gz-c:
@olemis @stdevStark updates api/README.md docs for those endpoints

Issue #1937, opened 2018-10-10
Run CLI integration tests with coverage


gz-c:
Each invocation of the CLI will need a unique coverage file

Requires #1936 

Once cobra is used, CLI coverage can be set up similar to this:

To build the binary:

`go test -c -tags testrunmain -o main.test -coverpkg=github.com/skycoin/skycoin/...` 

To invoke with test parameters (note, requires a double-dash):

`./main.test --test.v -test.coverprofile=foo.out print to`

Create `cmd/cli/cli_test.go`:

```go
// +build testrunmain

// This file allows us to run the entire program with test coverage enabled, useful for integration tests
package main

import "testing"

func TestRunMain(t *testing.T) {
	main()
}
```

In the cli/cobra init code, add this line, which will allow the added `--test` flags to be passed:
```go
pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
```

gz-c:
cli/integration_test.go needs:

- `TestMain` should compile the CLI binary with the options described above;
- `exec.Command` args need to include the `-test` flags described above;
- Each `exec.Command` invocation needs to use a unique filename for the coverprofile so that they can be aggregated into one master coverage file later (merging performed by `make merge-coverage`)

gz-c:
Unfortunately the coverage will still be incomplete; a go test binary will not produce a coverage file if its exit status is not 0, and many of the cli binary tests will cause a non-zero exit status

PR #1975, opened 2018-10-23
Switch from urfave to cobra

### LINKED ISSUE #1936: Switch from urfave/cli to spf13/cobra for cmd/cli
This will enable config loading from file.
It is also the only way to be able to run the CLI binary with test coverage.

Note that this will require all cli arguments to change from single dash to double dash prefix.

therealssj:
Fixes #1936 , #1937 

Changes:
- Switch from `urfave/cli` to `spf13/cobra`

Yes


src/cli/cli.go
```diff
@@ -100,11 +83,36 @@ COPYRIGHT:
...
+	seed              string
+	strictSeed        bool
+	entropy           int
+	hideSecrets       bool
+	mode              string
+	encrypt           bool
+	addrsFilename     string
+	seedsFilename     string
+	overwrite         bool
+)
```
therealssj: defined all flags globally since they could be easily reused without affecting functionality of other commands

src/cli/cli.go
```diff
@@ -100,11 +83,36 @@ COPYRIGHT:
...
-// App Wraps the app so that main package won't use the raw App directly,
-// which will cause import issue
-type App struct {
-	gcli.App
-}
+var (
+	cliConfig Config
+	apiClient *api.Client
+	quitChan  = make(chan struct{})
+)
```
therealssj: no context in cobra so converted these to unexported global vars

src/api/blockchain_test.go
```diff
@@ -784,7 +784,7 @@ func TestGetBlocks(t *testing.T) {
 			body: &httpBody{
 				Seqs: "1,2,3",
 			},
-			seqs: []uint64{1, 2, 3},
+			seqs:                  []uint64{1, 2, 3},
```
gz-c: Do you have a different `goformat` tool that's causing this?

therealssj: not sure why this happens, I dont think you can configure this?
I am on `go1.11.2 darwin/amd64`

CHANGELOG.md
```diff
@@ -84,6 +84,7 @@ Make sure to upgrade to v0.25.0 so that your node will continue to connect once
 - `cli generateWallet` renamed to `cli walletCreate`
 - `cli generateAddresses` renamed to `cli walletAddAddresses`
 - `/api/v1/explorer/address` is deprecated in favor of `/api/v1/transactions?verbose=1`
+- Switch `skycoin-cli` from `urfave/cli` to `spf13/cobra`
```
gz-c: This requires all options to use `--` instead of `-` right?

src/cli/cli.go
gz-c: Is it possible to scope them to their command?

CHANGELOG.md
therealssj: yes

CHANGELOG.md
```diff
@@ -117,6 +117,7 @@ Make sure to upgrade to v0.25.0 so that your node will continue to connect once
 - Node will send more peers before disconnecting due to a full peer list
 - Refactor CSRF to use HMAC tokens.
 - Add transaction verification parameters to the `GET /health` response
+- Switch `skycoin-cli` from `urfave/cli` to `spf13/cobra`
```
gz-c: Changelog entry needs to be moved to the unreleased section, and add an entry clarifying that all options must use `--` instead of `-`. Make sure all docs that reference single dash usage are updated

therealssj: `-` is allowed in shorthand

PR #1988, opened 2018-10-29
Use POST for long queries

Senyoret1:
Changes:
- Stops using GET and start using POST for the calls made to the api with a lot of addresses.
- A mechanism was created to avoid racing conditions while getting csrf tokens, as the node invalidates them right after the first use and that cause big problems with racing conditions. If this is not done, a racing condition happens if the browser is refreshed on the history or the outputs page, as a csrf token is requested on `AppService` and then another one for getting the data for the page, and then the page remains loading indefinitely.

No


gz-c:
Should we change the CSRF mechanism on the backend?

PR #1999, opened 2018-11-01
Send GIVP before disconnecting due to peer list full

### LINKED ISSUE #735: Send peers to client before disconnecting due to full peers

gz-c:
Fixes #735 

Changes:
- Sends a GIVP message before disconnecting due to full peers
- When adding an incoming peer, if the peerlist is full, kick out the oldest peer if the oldest peer was last seen in over 24 hours

This GIVP is unlikely to be sent since the default peerlist size is 65535 and the oldest peer would be kicked out.  However, there are still certain DoS scenarios that could flood peerlists and cause problems. The peerlist data will need to track the idea of "Seen" as in received via PEX versus "Seen" as in successfully established a connection to resolve part of that problem.

Yes

### DIFF
--- a/src/daemon/messages.go
+++ b/src/daemon/messages.go
@@ -209,8 +209,29 @@ func (gpm *GivePeersMessage) process(d daemoner) {
 	if d.PexConfig().Disabled {
 		return
 	}
+
 	peers := gpm.GetPeers()
-	logger.Debugf("Got %d peers via PEX: %s", len(peers), strings.Join(peers, ", "))
+
+	if len(peers) == 0 {
+		return
+	}
+
+	// Cap the number of peers printed in the log to prevent log spam abuse
+	peersToFmt := peers
+	if len(peersToFmt) > 1 {
+		peersToFmt = peersToFmt[:1]
+	}
+	peersStr := strings.Join(peersToFmt, ", ")
+	if len(peers) != len(peersToFmt) {
+		peersStr += fmt.Sprintf(" and %d more", len(peers)-len(peersToFmt))
+	}
+
+	logger.WithFields(logrus.Fields{
+		"addr":   gpm.c.Addr,
+		"gnetID": gpm.c.ConnID,
+		"peers":  peersStr,
+		"count":  len(peers),
+	}).Debug("Received peers via PEX")
 
 	d.AddPeers(peers)
 }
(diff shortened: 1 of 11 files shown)


src/daemon/messages.go
```diff
@@ -394,14 +391,13 @@ func (intro *IntroductionMessage) verify(d daemoner) (*useragent.Data, error) {
...
-			logger.WithField("addr", addr).Info("Extra data length does not meet the minimum requirement")
+			logger.WithFields(fields).Info("Extra data length does not meet the minimum requirement")
 			return nil, ErrDisconnectInvalidExtraData
 		}
 		copy(bcPubKey[:], intro.Extra[:len(bcPubKey)])
 
 		if d.BlockchainPubkey() != bcPubKey {
-			logger.WithFields(logrus.Fields{
-				"addr":         addr,
+			logger.WithFields(fields).WithFields(logrus.Fields{
```
olemis: Level = warning ?

src/daemon/messages.go
```diff
@@ -411,13 +407,13 @@ func (intro *IntroductionMessage) verify(d daemoner) (*useragent.Data, error) {
 		userAgentSerialized := intro.Extra[len(bcPubKey):]
 		userAgent, _, err := encoder.DeserializeString(userAgentSerialized, useragent.MaxLen)
 		if err != nil {
-			logger.WithError(err).Info("Extra data user agent string could not be deserialized")
+			logger.WithError(err).WithFields(fields).Info("Extra data user agent string could not be deserialized")
```
olemis: Level = warning ?

src/daemon/messages.go
```diff
@@ -411,13 +407,13 @@ func (intro *IntroductionMessage) verify(d daemoner) (*useragent.Data, error) {
...
 			return nil, ErrDisconnectInvalidExtraData
 		}
 
 		userAgentData, err = useragent.Parse(useragent.Sanitize(userAgent))
 		if err != nil {
-			logger.WithError(err).WithField("userAgent", userAgent).Info("User agent is invalid")
+			logger.WithError(err).WithFields(fields).WithField("userAgent", userAgent).Info("User agent is invalid")
```
olemis: Level = warning ?

PR #2003, opened 2018-11-02
[ci] fixes #1996 - Correct version of brew gcc@6 formula to restore macos Travis builds

### LINKED ISSUE #1996: [ci] Travis CI macos builds failing for forks
Builds used to  `brew` install `gcc=6.4.0-2` . Nonetheless `gcc@6` formula was upgraded so as to install `gcc=6.5.0` . Since then `macos` builds fail as explained in Homebrew/homebrew-core#33543 . See also Homebrew/homebrew-core#33544 .

olemis:
Fixes #1996

Changes:
- Separate script to force installation of `gcc=6.4.0-2`

no


ci-scripts/install-travis-gcc.sh
```diff
@@ -0,0 +1,21 @@
+
+# Install gcc6 (6.4.0-2 on Mac OS) for Travis builds
+
+if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+  sudo apt-get install -qq g++-6 && sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-6 90;
+fi
+
+if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
```
olemis: @gz-c this issue can be solved like this , or instead Travis -specific checks can be returned back to `.travis.yml` and only the lines that follow would remain in this script .

PR #1929, opened 2018-10-03
Bip39 mnemonic verification

J0nathan-S:
Fixes #1895

Changes:
- Added Bip39 mnemonic verification

Does this change need to mentioned in CHANGELOG.md?
yes

PR #2021, opened 2018-11-17
Bip39 verify

therealssj:
Fixes #1929 

Changes:
-  Add `/api/v2/wallet/seed/verify` to verify if seed is a valid bip39 mnemonic seed

yes

### DIFF
--- a/src/api/client.go
+++ b/src/api/client.go
@@ -272,11 +272,13 @@ func (c *Client) PostJSONV2(endpoint string, reqObj, respObj interface{}) (bool,
 		return false, rspErr
 	}
 
-	decoder = json.NewDecoder(bytes.NewReader(wrapObj.Data))
-	decoder.DisallowUnknownFields()
+	if respObj != struct{}{} {
+		decoder = json.NewDecoder(bytes.NewReader(wrapObj.Data))
+		decoder.DisallowUnknownFields()
 
-	if err := decoder.Decode(respObj); err != nil {
-		return false, err
+		if err := decoder.Decode(respObj); err != nil {
+			return false, err
+		}
 	}
 
 	return true, rspErr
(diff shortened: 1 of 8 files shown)


src/api/client.go
```diff
@@ -272,11 +272,13 @@ func (c *Client) PostJSONV2(endpoint string, reqObj, respObj interface{}) (bool,
 		return false, rspErr
 	}
 
-	decoder = json.NewDecoder(bytes.NewReader(wrapObj.Data))
-	decoder.DisallowUnknownFields()
+	if respObj != struct{}{} {
```
gz-c: What is this check for

src/api/client.go
```diff
@@ -797,6 +799,15 @@ func (c *Client) NewSeed(entropy int) (string, error) {
 	return r.Seed, nil
 }
 
+// VerifySeed verifies whether the given seed is a valid bip39 mnemonic or not
+func (c *Client) VerifySeed(seed VerifySeedRequest) (bool, error) {
```
gz-c: The arg could be `seed string`

src/api/client.go
therealssj: image

I think its this

src/api/client.go
therealssj: fixed

src/api/client.go
gz-c: The advantage is nullified by avoiding `DisallowUnknownFields()` when the struct is a `struct{}{}`

src/api/client.go
```diff
@@ -797,6 +799,17 @@ func (c *Client) NewSeed(entropy int) (string, error) {
 	return r.Seed, nil
 }
 
+// VerifySeed verifies whether the given seed is a valid bip39 mnemonic or not
+func (c *Client) VerifySeed(seed string) (bool, error) {
+	ok, err := c.PostJSONV2("/api/v2/wallet/seed/verify", VerifySeedRequest{
+		Seed: seed,
+	}, nil)
```
gz-c: use the `struct{}{}` but don't check above

PR #2039, opened 2018-12-13
Swaplab integration

### LINKED ISSUE #2032: Integrate swaplab in a tab in the wallet
Swaplab is a coin purchase service.

Add a new tab to the wallet which integrates with the swaplab service to allow coins to be purchased and sent directly to the user's wallet.

The API sandbox for the service for Skycoin is:
https://swaplab.cc/api-v3-skycoin/index.html

mahansky:
Fixes #2032 

Changes:
- adds "Exchange" tab where you can exchange cryptocurrencies using swaplab.cc 

Screenshots:
screenshot 2018-12-13 at 19 06 24
screenshot 2018-12-13 at 19 06 37

I tested it only using the sandboxed api methods. Dev build uses sandboxed api by default.

Known issues:
- Destination addresses are not validated
- Icons in the navbar are custom images, can someone create new one? (currently using the same as History tab).
- When you are on the status page, you can't click on Exchange tab to go back to form page (it's the same thing as you can not refresh History page without going somewhere else first)

----

Test instructions:

Best way to test the "status" page is to replace `ExchangeService@status` method body with:
```js
  const statuses = [
      'user_waiting',
      'market_waiting_confirmations',
      'market_confirmed',
      'market_exchanged',
      'market_withdraw_waiting',
      'complete',
      'error', // move higher to see error state
    ];

    let index = 0, isDone = false, shouldContinue = true;

    return Observable
      .timer(0, 3000)
      .flatMap(() => this.post('orders/status', { id }, { status: statuses[index++] }))
      .map(data => data.result)
      .takeWhile((order: ExchangeOrder) => {
        if (isDone) {
          shouldContinue = false;
        }

        if (['complete', 'error'].indexOf(order.status) !== -1) {
          isDone = true;
        }

        return shouldContinue;
      });
```
The status endpoint returns 404 for random order IDs, so you should replace `this.order.id` with `"4729821d-390d-4ef8-a31e-2465d82a142f"` in `ExchangeStatusComponent@ngOnInit`


mahansky:
Mainly because there is a lot of cryptocurrencies you can choose from and supporting all of them can be hard as the list can change in the future. There is a few libraries trying to solve this problem, but it looks like there is still a lot of cryptocurriencies which they dont support so it would validate just some of them. (new dependency, half working solution)

mahansky:
If you are asking about the exchange address, I don't think so. At worst, they would return error response which wouldn't show the status page (and exchange address) at all.

src/gui/static/src/app/components/pages/exchange/exchange.component.html
```diff
@@ -0,0 +1,34 @@
...
+
+    <app-exchange-status
+      *ngIf="order"
+      [order]="order"
+    ></app-exchange-status>
+  </div>
+  <div class="-contact">
+    <div>
+      <span *ngIf="!order && hasLast()" (click)="showLast()">
+        <a (click)="showLast($event)" href="#">{{ 'exchange.show-last' | translate }}</a>
```
Senyoret1: The event should not be necessary in both tags.

src/gui/static/src/app/components/pages/exchange/exchange.component.html
```diff
@@ -0,0 +1,34 @@
...
+      <span>
+        {{ 'exchange.need-help' | translate }}
+        <a href="https://swaplabcc.freshdesk.com/support/home" target="_blank" rel="noreferrer nofollow">
+          {{ 'exchange.support-portal' | translate }}
+        </a>
+      </span>
+      <span>|</span>
+      <span>
+        {{ 'exchange.powered-by' | translate }}
+        <a href="https://swaplab.cc">SWAPLAB.CC</a>
```
Senyoret1: This link should be opened with `target="_blank" rel="noreferrer nofollow"` too.

src/gui/static/src/app/services/exchange.service.ts
```diff
@@ -0,0 +1,71 @@
...
+  constructor(
+    private http: HttpClient,
+  ) { }
+
+  tradingPairs(): Observable<TradingPair[]> {
+    return this.post('trading_pairs').map(data => data.result);
+  }
+
+  exchange(pair: string, fromAmount: number, toAddress: string): Observable<ExchangeOrder> {
+    return this.post('orders', { pair, fromAmount, toAddress }).map(data => data.result);
```
Senyoret1: Should not this function also receive and send the `refundAddress` parameter? Probably that will make things easier in case of any error.

src/gui/static/src/app/services/exchange.service.ts
```diff
@@ -0,0 +1,71 @@
...
+
+        if (['complete', 'error'].indexOf(order.status) !== -1) {
+          isDone = true;
+        }
+
+        return shouldContinue;
+      });
+  }
+
+  setLastOrder(order) {
```
Senyoret1: For consistency, maybe it would be better to use a setter.

src/gui/static/src/app/services/exchange.service.ts
```diff
@@ -0,0 +1,71 @@
...
+    this.lastOrder = order;
+  }
+
+  getLastOrder() {
```
Senyoret1: For consistency, maybe it would be better to use a getter.

src/gui/static/src/app/services/exchange.service.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  status(id: string): Observable<ExchangeOrder> {
+    let isDone = false, shouldContinue = true;
+
+    return Observable
+      .timer(0, 30 * 1000)
+      .flatMap(() => this.post('orders/status', { id }))
+      .map(data => data.result)
+      .takeWhile((order: ExchangeOrder) => {
```
Senyoret1: If this `takeWhile` is placed right below `timer` and you put the `if (['complete', 'error'].indexOf(…` part inside `map`, the `shouldContinue` variable is no longer necessary, the code would be simpler and it would avoid making an extra connection to the Server after the next timer execution.

src/gui/static/src/app/services/exchange.service.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  exchange(pair: string, fromAmount: number, toAddress: string): Observable<ExchangeOrder> {
+    return this.post('orders', { pair, fromAmount, toAddress }).map(data => data.result);
+  }
+
+  status(id: string): Observable<ExchangeOrder> {
+    let isDone = false, shouldContinue = true;
+
+    return Observable
```
Senyoret1: This observable does not retry to get the data in case of error. So if, for example, there is a connection error, the observable will not retry to obtain the data, but the UI will continue unchanged, as if the status could still be updated.

gz-c:
Blocked on #2119 #2126

gz-c:
@mahansky https://github.com/skycoin/skycoin/blob/develop/src/api/README.md#key-value-storage-apis The API for storage is available now, use `client` type storage for this

mahansky:
@Senyoret1 PR is ready for review.
Responsivity issues are fixed, there is new "order history" leveraging the custom data endpoint (still "hidden" in the bottom left as it's the best spot I could find it, buttons really get in a way of the exchange flow). 
The refund address is not documented so I didn't implement it (would probably also require another address validation for different coins).
There is no more information like different exchange times, or the time limit on deposit. I based it on their [tutorial videos](https://swaplabcc.freshdesk.com/support/solutions/articles/43000059781-video-tutorial-exchange-from-ethereum-eth-to-ripple-xrp-)

gz-c:
Merge conflicts

@Senyoret1 can you review?

gz-c:
@mahansky the build failure is `ERROR in src/app/components/pages/exchange/exchange-history/exchange-history.component.html(1,59): : Property 'dialogRef' is private and only accessible within class 'ExchangeHistoryComponent'` during `make build-ui-travis`

gz-c:
@Senyoret1 merged, please create new issues for followups

PR #2048, opened 2018-12-20
Add CLI addressTransactions command

BigOokie:
Fixes # NA

Changes:
- add addressTransactions command to CLI. Use /api/v1/transactions

Yes. Added as part of the PR - review needed

### DIFF
--- a/src/api/client.go
+++ b/src/api/client.go
@@ -903,20 +903,6 @@ func (c *Client) PendingTransactionsVerbose() ([]readable.UnconfirmedTransaction
 	return v, nil
 }
 
-// GetTransactions makes a request to POST /api/v1/transactions
-func (c *Client) GetTransactions(addrs []string) ([]readable.TransactionWithStatus, error) {
-	v := url.Values{}
-	v.Add("addrs", strings.Join(addrs, ","))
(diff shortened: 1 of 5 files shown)


BigOokie:
Note: This command only works when running against Skycoin Node v0.25.0.
Running on an earlier version of the Node will return a JSON error.

src/api/client.go
```diff
@@ -903,6 +903,20 @@ func (c *Client) PendingTransactionsVerbose() ([]readable.UnconfirmedTransaction
 	return v, nil
 }
 
+// GetTransactions makes a request to POST /api/v1/transactions
+func (c *Client) GetTransactions(addrs []string) ([]readable.TransactionWithStatus, error) {
```
gz-c: The `Transactions` method already does this

BigOokie: Will take a look at it now.

BigOokie: Changes have been pushed. See below.

BigOokie: Wondering if using `TransactionsVerbose` would be more useful as it provides both `input` and `output` - whereas `Transactions` appears to just provide `output`

gz-c: The Verbose is probably better to use for this

PR #2068, opened 2019-01-03
Revert "Create ru.json : Russian localisation added"

gz-c:
Reverts skycoin/skycoin#2049


gz-c:
The localization still has references to MDL in it

PR #2091, opened 2019-01-12
coinhour bank cli and api

therealssj:
Working
- The client code is in `src/coinhourbank`. This is a trimmed down version of the full implementation with only the deposit api.
- New api set `COINHOUR_BANK` has been added for CHB related apis.
- Two new flags are added: `EnableCHB ` and `CHBNodeURL`
   `EnableCHB`: If this is true then chb api set is enabled and the chb client is initialized aswell.
   `CHBNodeURL`: This is the chb client URL, when `EnableCHB` is true this is used to initialize the chb client.
- The chb client is attched to the gateway https://github.com/skycoin/skycoin/pull/2091/files#diff-e646b38e6b5d33d2539c80b5fc775093R44
- When `EnableCHB` is `true` and `CHBNodeURL` is not empty the `InitCHB()` function is called which initializes the chb client https://github.com/skycoin/skycoin/pull/2091/files#diff-277c4a77cc98d8542ef4e4d39336f3efR328
- The `/api/v2/coinhourbank/deposit` endpoint can be used to deposit coinhours into the coinhour bank 

Changes:
- Add coinhour bank cli commands
- Add chb api

Yes


src/cli/coinhourbank.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	wlt, err := resolveWalletPath(cliConfig, walletFile)
+	if err != nil {
+		return nil, err
+	}
+
+	wltPassword, err := c.Flags().GetString("password")
```
gz-c: Wallet passwords need to use the special password reader so that it does not need to be entered in the terminal

src/coinhourbank/client.go
```diff
@@ -0,0 +1,193 @@
...
+
+// NewHourBankClient creates a new instance of HourBankClient
+func NewHourBankClient(bankURL string) *HourBankClient {
+	return &HourBankClient{
+		bankURL: bankURL,
+	}
+}
+
+// Balance returns balance for current account
+func (c *HourBankClient) Balance(account string) (uint64, error) {
```
gz-c: What is balance for? It can't check balance without auth, and there won't be auth from the desktop wallet

therealssj: oh I assumed its fine to know account balance without auth because it on the blockchain anyways

SNChris: Balance is really public unless you know all addresses the user deposits from.

SNChris: Cant edit for some reason, I meant *not really public.

Issue #2083, opened 2019-01-08
Windows desktop electron client sometimes shows "Error#1" on start


200495:
When wallet update from 0.24.1 to 0.25.0 ,it become that.

200495:
![image](https://user-images.githubusercontent.com/23026128/50816417-7c4ec700-135b-11e9-9494-d608cbc005af.png)

200495:
Windows

gz-c:
After quit wallet, wait a minute, then start new wallet

200495:
I have done it,but it still have problem.

200495:
After restart computer ,it become to normal.

200495:
> After quit wallet, wait a minute, then start new wallet
I have done it,but it still have problem.
I wait for 20 minutes,restart it and it still have problem.
After restart computer ,it become to normal.
It lack of user experience.

200495:
> After quit wallet, wait a minute, then start new wallet

I have done it,but it still have problem.
I wait for 20 minutes,restart it and it still have problem.
After restart computer ,it become to normal.
It lack of user experience.

PR #2101, opened 2019-01-18
Desktop wallet fixes

iketheadore:
Fixes: #2083

yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - `/api/v1/health` will return correct build info when running Docker containers based on `skycoin/skycoin` mainnet image.
+- #2083, Windows desktop wallet sometimes shows "Error#1" on start
 
 ### Changed
 - Extend URI specification to allow plain addresses (i.e. without a `skycoin:` prefix)
--- a/electron/src/electron-main.js
+++ b/electron/src/electron-main.js
@@ -115,7 +115,7 @@ function startSkycoin() {
 		  var id = setInterval(function() {
 			// wait till the splash page loading is finished
 			if (splashLoaded) {
-          	  app.emit('skycoin-ready', { url: currentURL });
+			  app.emit('skycoin-ready', { url: currentURL });
 			  clearInterval(id);
 			}
 		  }, 500);
--- a/electron/src/electron-main.js
+++ b/electron/src/electron-main.js
@@ -21,6 +21,7 @@ global.eval = function() { throw new Error('bad!!'); }
 
 let currentURL;
 let showErrorCalled = false;
+let splashLoaded = false
 
 // Detect if the code is running with the "dev" arg. The "dev" arg is added when running npm
 // start. If this is true, a local node will not be started, but one is expected to be running,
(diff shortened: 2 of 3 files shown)


Senyoret1:
The white screen problem is a bit complicate for me because I can not reproduce it on my pc, so I’m guessing some things.

Placing `event.returnValue = '';` in the line 82 solves the white screen problem, but should generate a new one. The `getSavedWalletsDataSyncEvent = event;` part above should make that line unnecessary, since thanks to it a valid value would be assigned to `event.returnValue` as soon as `setWalletsFolderPath` is called. This, of course, means assuming that `setWalletsFolderPath` will be called at some point, and the error may be there.

Here is the situation: if `setWalletsFolderPath` is never called, the screen will remain white and adding `event.returnValue = '';` in the line 82 solves the problem because assigning a value to `event.returnValue` allows the browser code to continue working instead of stay waiting for a response. However, as `setWalletsFolderPath` was never called, the code does not know were to save the hardware wallets info, so when connecting a hardware wallet `hwSaveWalletsDataSync` will be called and it will throw this error, making the hardware wallet integration unusable:

err

Taking all this into account, a more adequate solution may be to assure that `setWalletsFolderPath` is called with a valid value. My guess is that this part on lines 339-345 is failing:
https://github.com/skycoin/skycoin/blob/c458da61bc9912f85dd447947e01784e69c4deba/electron/src/electron-main.js#L339-L345

As you can see, the `.catch` part is empty. You can change the catch part to something like `.catch(e => { console.log(e); });` to see if it is failing and why. If `.catch` is being called, maybe the problem can be solved by adding `const url = require('url');` at the top of the file and replacing `.get(e.url + '/api/v1/wallets/folderName')` with `.get(new url.URL('/api/v1/wallets/folderName', e.url).href) `. Other solutions could be to create a retry mechanism, like the one you created to wait until the splash screen is loaded, or use a default path in case of error/timeout.

Apart from that, I do not understand what the purpose of `splashLoaded` is. The splash screen is almost guaranteed to be loaded before `if (line.indexOf(marker) !== -1)` is true and, even if it is not, it should not generate any problem. The operation of the splash screen, as far as I know, has not changed since version 0.24.0.

iketheadore:
@Senyoret1 I will add log in the catch to have a test. 

If I don't add the `splashLoaded` to make sure the splash page is loaded before loading the wallet, it will have `did-fail-load` error, and show the error page.

Senyoret1:
@gz-c Just the wallet name, the addresses and if the user completed the configuration the last time it was connected. The data is used to allow to see the addresses and balance even when the device is not connected.

iketheadore:
@Senyoret1, I find it's the problem of my windows laptop, @gz-c do you remember I said that the wallet always try to connect to an unreachable proxy of port 1087, that's the problem, the `axios.get` will fail to get wallet folderName cause it try to connect the `0.0.0.0:1087`.  I don't know why the proxy still exist, I've disabled it,  and tested to run a new proxy with a new port, the log still says it try to connect the proxy of port 1087. Anyway I think the white page should not be a problem on normal windows.

gz-c:
@senyoret1 should we add an api to the back end to store arbitrary data?

gz-c:
I would prefer to remove the need for nodejs's fs interface. Also, where is the data saved?

@mahansky recently mentioned a possible need for a generic way to save data in a file in the data folder through the API, but I forget for which feature

mahansky:
It was for the swaplab integration

PR #2103, opened 2019-01-19
Allow to select specific outputs in the advanced send form

### LINKED ISSUE #2099: Advanced spend: Ability to choose specific outputs to spend in the UI
The `POST /api/v1/wallet/transaction` API allows one to choose which outputs to spend. The Advanced Send UI page should support this feature.

Senyoret1:
Fixes #2099

Changes:
- Added a select field that allows selecting the outputs that will be used to send the coins and hours. If no output is selected, the coins and hours will be sent from any of the outputs of the selected addresses (it is indicated in the UI). The backend does not necessarily use all the outputs selected by the user, but just the ones it considers necessary.

- Now if no address is selected, the coins and hours will be sent from any of the addresses of the selected wallet (it is indicated in the UI). This allows greater flexibility when deciding if the transaction should be limited by wallet, addresses or outputs.

- Some help text were changad.

- The fields for selecting addresses and outputs are only displayed after a wallet is selected.

- Added a text that indicates the maximum amount of coins and hours that can be sent with the wallet, addresses and outputs that the user has selected.

- Fixed the following error: if the user deselects “Automatic coin hours allocation”, enter a number of hours to send and re-selects “Automatic coin hours allocation”, the code will continue validating if the amount of hours entered by the user is lower or equal to the amount of available hours, even being that the hours field is not visible.

- Now the form validation requires the number of hours to be sent to be equal to or less than the number of available hours minus one (the "minus one" part is new), because at least one hour must be reserved for the transaction fee.

form

No


src/gui/static/src/app/components/pages/send-skycoin/send-form-advanced/send-form-advanced.component.ts
```diff
@@ -358,4 +423,43 @@ export class SendFormAdvancedComponent implements OnInit, OnDestroy {
 
     return hoursSelection;
   }
+
+  private updateAvailableBalance() {
```
mahansky: Can you extract those form values into variables and merge those two same conditions into one?

src/gui/static/src/app/components/pages/send-skycoin/send-form-advanced/send-form-advanced.component.scss
```diff
@@ -157,3 +180,21 @@ label mat-icon {
   vertical-align: text-bottom;
   padding-left: 5px;
 }
+
+.-available-msg {
+  background-color: $grey-lightest;
+  border: 1px dotted $grey-light;
```
mahansky: Add a bit of `border-radius` to stay consistent with the form controls

gz-c:
Merge conflict

Senyoret1:
The last commit solves an additional error:

- The user was being told that the number of sendable hours was equal to the total number of hours minus one, which was not correct, since it is the total number of hours minus the minimum fee.

mahansky:
I tried it and it looks good - getting the same values as in the tx preview

PR #2112, opened 2019-01-23
Show a warning while the wallet is synchronizing

### LINKED ISSUE #2109: Do not show the balance while the node is synchronizing
Currently while the node is synchronizing it is possible to see the balance of the wallets and addresses. This has caused confusion among several users, who think that they may have lost coins or that their balance is incorrect. To avoid this, it would be good if, as happens with the general balance, the balances of the wallets and addresses are not displayed while the node is synchronizing.

Additionally, we could allow to see the balance and send transactions in the forms for sending coins, but it would be good to show a warning in the UI.

Senyoret1:
Fixes #2109

Changes:
- The balance of the wallets and addresses is not shown if the database is still being synchronized.
- If the user attempts to send a transaction while the database is being synchronized, a warning is displayed and confirmation is requested.

No

### DIFF
--- a/src/gui/static/src/app/components/layout/header/header.component.html
+++ b/src/gui/static/src/app/components/layout/header/header.component.html
@@ -47,11 +47,14 @@
     </div>
     <div *ngIf="appService.error === 3">{{ 'header.errors.csrf' | translate }}</div>
   </mat-toolbar>
-  <mat-toolbar class="notification-bar" *ngIf="hasPendingTxs">
-    <div>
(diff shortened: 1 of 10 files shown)


gz-c:
i don't know if we should hide the balance while synchronizing, instead we can communicate that the balance may not be accurate until the wallet is fully synchronized

Senyoret1:
I made the change. Now a warning is displayed and it is possible to see the balance:
warning

PR #2116, opened 2019-01-23
Allow to filter the transaction history

### LINKED ISSUE #2104: Allow to filter the transaction history list
It should be possible to show only the transactions of an specific wallet-address.

Senyoret1:
Fixes #2104

Changes:
- Now it is possible to limit the history page to certain addresses or wallets.
filter
filter2

No

### DIFF
--- a/src/gui/static/src/app/components/pages/transaction-list/transaction-list.component.html
+++ b/src/gui/static/src/app/components/pages/transaction-list/transaction-list.component.html
@@ -1,7 +1,13 @@
 <app-header [headline]="'title.transactions' | translate"></app-header>
 <div class="container">
   <div [formGroup]="form" class="form-field" *ngIf="allTransactions">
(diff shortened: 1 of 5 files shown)


mahansky:
Really nice feature, here's a few of my ideas on how to make it better: 
- add a clear button (maybe an X on the right side of the filter)
- add option to see tx history of address to the wallet-detail component's menu
- remove the line under filter, txs are in container that makes it separated enough
- make the text "Showing.." smaller to fit the design (13px)
- selected values could be lighter 
- maybe change the text from "Showing..." to "No filter active"/"Active filter:" or something like that to make it clear that it's something user can change

mahansky:
Could you make the X button smaller so it's the same as the filter icon? 
I tried these values and it looks just fine

```
    width: 24px;
    height: 24px;
    margin-top: 10px;
```

screenshot 2019-02-06 at 13 27 57

Issue #1785, opened 2018-08-16
API endpoint to expose configuration parameters for use by web/mobile wallet multicoin API


gz-c:
@mahansky @Senyoret1 which parameters would be needed to load a new coin in the multicoin wallet?

Senyoret1:
I'm not totally sure about which parameters you want to know about. At this time, to add a coin it is only necessary to add a new a TypeScript class in the multi-coin wallet and some branding images. The parameters that this class must contain are described in https://github.com/skycoin/skycoin-web/blob/develop/MULTICOIN-INFO.md . That file also indicates where the file should be added, information about the images and a simple initialization step that is required.

If those are the parameters that you are refering to, I think that for now there is no need to make changes in the node.

gz-c:
The parameters described in https://github.com/skycoin/skycoin-web/blob/develop/MULTICOIN-INFO.md#adding-a-coin would be exposed by this endpoint and loaded automatically by the web or mobile wallets to apply the settings, after the user provided a URL to the node

Senyoret1:
For that to work, the api should return the following params (not necessarily using the names on this list):

- `coinName`: the full name of the coin (like “Skycoin”).

- `coinSymbol`: the short name of the coin (like “SKY”).

- `hoursName`: the name of the hours produced by the coin (like “Coin Hours”).

- `cmcTickerId` (optional): the ID of the coin in Coin Market Cap (for getting the current USD price. The ticket id of Skycoin is `1619`).

- `coinExplorer` (optional): the URL of the blockchain explorer of the coin.

In addition to returning these data, the images described in https://github.com/skycoin/skycoin-web/blob/develop/MULTICOIN-INFO.md#additional-information-about-the-header-images should be added to the node and be accessible through a public URL, to let remote clients access them.

Senyoret1:
Due to https://github.com/skycoin/skycoin/issues/2136, the CMC API was replaced with the Coin Paprika API, so instead of the ID of the coin in CMC we need the ID in Coin Paprika. The Skycoin ID in Coin Paprika is `sky-skycoin`.

[...]

PR #2117, opened 2019-01-24
Fiberparams to health

### LINKED ISSUE #2082: Add fiber coin parameters to /health endpoint
- Proper name (e.g. "Skycoin")
- Ticker code / short name (e.g. "SKY")
- Coin hours name (e.g. "SKY Hours")

therealssj:
Fixes #2082 

Output:
screenshot 2019-02-10 at 12 23 25 pm

Changes:
- Add `display_name`,`coinhours_name`,`ticker` and `explorer_url` to `/health` endpoint.

yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   Now all options of a cli command must only use `--` prefix instead of a mix of `--` and `-` prefixes.
   `-` prefix is only allowed when using shorthand notation.
 - Use an optimized `base58` library for faster address decoding and encoding.
+- Add `ticker` and `coinhours_name` to `/health` endpoint
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   Now all options of a cli command must only use `--` prefix instead of a mix of `--` and `-` prefixes.
   `-` prefix is only allowed when using shorthand notation.
 - Use an optimized `base58` library for faster address decoding and encoding.
-- Add `display_name, `coinhours_name` and `ticker` to `/health` endpoint
+- Add `display_name`,`coinhours_name`,`ticker` and `explorer_url` to `/health` endpoint.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   Now all options of a cli command must only use `--` prefix instead of a mix of `--` and `-` prefixes.
   `-` prefix is only allowed when using shorthand notation.
 - Use an optimized `base58` library for faster address decoding and encoding.
-- Add `ticker` and `coinhours_name` to `/health` endpoint
+- Add `display_name, `coinhours_name` and `ticker` to `/health` endpoint
 
 ### Removed
 
--- a/cmd/skycoin/skycoin.go
+++ b/cmd/skycoin/skycoin.go
@@ -32,6 +32,8 @@ var (
 
 	logger = logging.MustGetLogger("main")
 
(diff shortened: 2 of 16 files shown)


therealssj:
`coin` works fine for the proper name imo.

gz-c:
Skycoin is different from skycoin and used in different contexts, for example a uri prefix would be skycoin but a display name would be Skycoin

Senyoret1:
Is this related to https://github.com/skycoin/skycoin/issues/1785 ? If it is, then the explorer URL and the Coin Market Cap ID would also be nedded. Additionally, some of the images displayed in the wallet would have to be added to the node and made accessible from a public URL.

gz-c:
These should be part of `fiber.toml`.

@Senyoret1 yes, they are related to #1785. Can you enumerate additional parameters needed in that issue?

therealssj:
@gz-c https://github.com/skycoin/skycoin/blob/develop/src/skycoin/parameters.go#L55
Then we should set `CoinName` through `fiber.toml` 
Though I can see why it was decided to not put it in `fiber.toml` before as these parameters don't really "define" the node per se

gz-c:
CoinName is set on the command line, it can be left at that

Senyoret1:
@gz-c I listed the necessary parameters in the issue (the comment is in https://github.com/skycoin/skycoin/issues/1785#issuecomment-461957270 ). However, there is a new complication due to https://github.com/skycoin/skycoin/issues/2136 : due to the deprecation of the Coin Market Cap public API, it is probably necessary to first define which method will be used to obtain the USD value.

Additionally, one thing I noticed is that in this PR the hours are named “SKY Hours”, but are not they supposed to be called ”Coin Hours”?

gz-c:
I'm not sure what the canonical name for coin hours should be, but I think we use "Coins Hours" everywhere

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   Now all options of a cli command must only use `--` prefix instead of a mix of `--` and `-` prefixes.
   `-` prefix is only allowed when using shorthand notation.
 - Use an optimized `base58` library for faster address decoding and encoding.
+- Add `display_name, `coinhours_name` and `ticker` to `/health` endpoint
```
gz-c: Missing a backtick

cmd/skycoin/skycoin.go
```diff
@@ -32,8 +32,14 @@ var (
 
 	logger = logging.MustGetLogger("main")
 
+	// DisplayName user-facing name
+	DisplayName = "Skycoin"
 	// CoinName name of coin
 	CoinName = "skycoin"
+	// CoinhoursName name of coinhours
+	CoinhoursName = "SKY Hours"
```
gz-c: Use "Coin Hours"

fiber.toml
```diff
@@ -31,6 +31,9 @@ peer_list_url = "https://downloads.skycoin.net/blockchain/peers.txt"
 # create_block_max_transaction_size = 32 * 1024
 # create_block_max_decimals = 3
 # max_block_size = 32 * 1024
+# display_name = "Skycoin"
+# coinhours_name = "SKY Hours"
```
gz-c: Use "Coin Hours"

src/skycoin/config.go
```diff
@@ -668,6 +677,11 @@ func (c *NodeConfig) RegisterFlags() {
 	flag.BoolVar(&c.Arbitrating, "arbitrating", c.Arbitrating, "Run node in arbitrating mode")
 	flag.StringVar(&c.WalletCryptoType, "wallet-crypto-type", c.WalletCryptoType, "wallet crypto type. Can be sha256-xor or scrypt-chacha20poly1305")
 	flag.BoolVar(&c.Version, "version", false, "show node version")
+
+	flag.StringVar(&c.CoinName, "coin", "skycoin", "name of the coin")
```
gz-c: Isn't coin name set a different way?

Senyoret1:
I added some comments that may be relevant on https://github.com/skycoin/skycoin/issues/1785#issuecomment-463778489

PR #2120, opened 2019-01-28
Implement common encoding interfaces for PubKey and SecKey

ap4y:
Fixes -

Changes:
- Implements `fmt.Stringer`, `encoding.Text/BinaryMarshaller` and `encoding.Text/BinaryUnmarshaller` for `cipher.PubKey` and `cipher.SecKey`.

I don't see similar changes in changelog.

### DIFF
--- a/src/cipher/crypto.go
+++ b/src/cipher/crypto.go
@@ -191,6 +191,39 @@ func (pk PubKey) Null() bool {
 	return pk == PubKey{}
 }
 
+// String implements fmt.Stringer for PubKey. Returns Hex representation.
+func (pk PubKey) String() string {
+	return pk.Hex()
+}
+
+// MarshalText implements encoding.TextMarshaler.
+func (pk PubKey) MarshalText() ([]byte, error) {
+	return []byte(pk.Hex()), nil
+}
+
+// UnmarshalText implements encoding.TextUnmarshaler.
+func (pk *PubKey) UnmarshalText(data []byte) error {
+	dPK, err := PubKeyFromHex(string(data))
+	if err == nil {
+		*pk = dPK
+	}
+	return err
+}
+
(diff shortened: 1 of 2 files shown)


gz-c:
What is the purpose

ap4y:
There are 2 problems we are facing at the moment:

- We have to do a lot of manual encoding (`.Hex()` calls) and decoding (`PubKeyFromHex`) for structs containing `Pub/SecKey`. These calls will be redundant if encoding protocols will be implemented. Not doing manual encoding in json for example will cause keys to be store as an array of float values. This is in most cases unexpected behavior as these keys are commonly encoded in their hex form in json payloads.

- Logs are polluted with `.Hex` calls which is again not necessary if `Stringer` would be implemented. These keys are commonly known by their hex form so logging binary values (which is current implicit behavior) doesn't make much sense.

gz-c:
the text marshaller works for json?

ap4y:
Quote from the [docs](https://golang.org/pkg/encoding/json/#Marshal):

> If an encountered value implements the Marshaler interface and is not a nil pointer, Marshal calls its MarshalJSON method to produce JSON. If no MarshalJSON method is present but the value implements encoding.TextMarshaler instead, Marshal calls its MarshalText method and encodes the result as a JSON string.

`BinaryMarshaller` is used by the `gob` encoding.

ap4y:
We use `net/rpc` which is using `gob` under the hood.

gz-c:
You're sending secret keys over the wire or logging them?

ap4y:
Neither of those, but we parse `SecKey` from `json` configs, which will require `TextMarshaller`.

gz-c:
Gob doesn't encode it as a fixed length byte array by default?

gz-c:
Can merge the `String` methods

ap4y:
I will close this one for now.

PR #2122, opened 2019-01-29
[api] fixes #2031 Swagger specs for REST API endpoints

### LINKED ISSUE #2031: Implement Swagger metadata to generate REST clients
Client libraries should export Swagger metadata that can be used in many different ways, e.g. generate client libraries .

stdevMac:
Fixes #2031

Changes:
- Swagger annotiations for v1 and v2 REST API endpoints
- Implement `rawTx` and `injectTransaction` in `v2` thus returning `RawTxn` JSON object

no


gz-c:
What aspect of the `api/v2` endpoints are not Swagger-friendly?

stdevMac:
This is the current status of the endpoints of `api/v?` from `Skycoin API`

gz-c:
regarding `/api/v2/transaction/verify`, it does not return multiple structs

All v2 APIs return a top level `{}` object. This object may or may not contain an `"error": {"message": "foo", "code": 123}` object.  This object may also some other object, such as `"transaction"`.  Both `"error"` and `"transaction"` can be present at the same time, but it may only be one or the other.

gz-c:
If there is any deficiency regarding `/api/v2` and compatibility with swagger, let me know. I don't think we need a v3.  We can migrate the incompatible v1 endpoints to v2.

stdevMac:
I will work con this tomorrow and i will tell you my progress.
I think we may need move some endpoints to `api/v2`.
In fact, `v3` is not needed if we can migrate this endpoints to `v2`.
The point with swagger is that they need a stable response struct, because they build they clients in base of this specifications, so this response struct can't change the state according to params.

olemis:
@gz-c the only ones posing major trouble are 

```
/api/v1/rawtx -> ok_ (Need V3 for response string as json)
/api/v1/injectTransaction -> ok_  (Need V3 for response string as json)
```

All other instances of incompatibilities can be resolved by forcing `verbose=1` and sticking to its output .

olemis:
@stdevMac it's not exactly like `response struct can't change the state according to params.` . The exact situation is that a particular struct field (no matter the level in the hierarchy) must always return same type no matter what parameters are supplied in to the endpoint .

gz-c:
forcing verbose can have performance costs, is there no way to make swagger consider it as two endpoints?

stdevMac:
Swagger does not allow to create multiple return values, it is what I was referring to that the structures must have a fixed form.

olemis:
@stdevMac the question is ... is it possible to include in Swagger spec 2 entries while annotating a single endpoint method ?

src/api/transaction_test.go
```diff
@@ -737,39 +737,61 @@ func TestInjectTransaction(t *testing.T) {
...
-				require.Equal(t, tc.err, strings.TrimSpace(rr.Body.String()), "got `%v`| %d, want `%v`",
-					strings.TrimSpace(rr.Body.String()), status, tc.err)
-			} else {
-				expectedResponse, err := json.MarshalIndent(tc.httpResponse, "", "    ")
-				require.NoError(t, err)
-				require.Equal(t, string(expectedResponse), rr.Body.String(), tc.name)
-			}
-		})
+				if status != http.StatusOK {
+					require.Equal(t, tc.err, strings.TrimSpace(rr.Body.String()), "got `%v`| %d, want `%v`",
```
olemis: Needs assertions for V2 error object

src/api/transaction_test.go
```diff
@@ -954,47 +976,70 @@ func TestGetRawTx(t *testing.T) {
...
```
olemis: Needs assertions for V2 error object

src/api/transaction_test.go
```diff
@@ -1292,7 +1337,8 @@ func TestGetTransactions(t *testing.T) {
 }
 
 type transactionAndInputs struct {
-	txn    coin.Transaction
+	txn coin.Transaction
+	// swagger:ignore
```
olemis: is it possible to put these comments after field definition ?

stdevMac:
> @stdevMac need to add `/transaction/inject` and `/transaction/raw` in API docs README . Just like docs for existing endpoints , Including sample request / response .

Done

src/api/transaction_test.go
stdevMac: no, but especially this one, can be eliminated

stdevMac: is already deleted for next commit

olemis:
@stdevMac merge conflicts need to be resolved before merging this branch .

PR #2126, opened 2019-02-04
filesystem api

### LINKED ISSUE #2119: Create API endpoints to store arbitrary data
The Angular wallet sometimes needs to save data. However, it is not possible to use `LocalStorage`, because the backend port changes in each execution, making the data inaccessible. Also, it would be good not to have to use Electron to access the file system.

To solve this it is necessary to create API endpoints to save data and read it. At least at the moment something very complex is not required, just being able to store arbitrary data (preferably a text string, to be able to write and read it easily from Javascript) and be able to identify the data in some way (probably an string serving as ID).

@mahansky , do you have any special requirements regarding this?

therealssj:
Fixes #2119 

Added `api/v2/data` endpoint which lets you store, get, delete or update data in a file. The url uses HTTP verbs (POST, GET and DELETE) for different actions.

Changes:
- Add `-data-file-path` option to set path of the arbitrary data store file. Defaults to `~/.skycoin/data.json`.
- Add `/api/v2/data` endpoint to store, get, delete or update arbitrary data in a file on disk defined by the `-data-file-path` option.
- Add new api set `FILESYSTEM` to enable `/api/v2/data` endpoint.

Yes


gz-c:
Use HTTP verbs

gz-c:
Use HTTP verbs:

Gopkg.lock
```diff
@@ -145,17 +145,17 @@
   branch = "master"
   name = "github.com/prometheus/client_model"
   packages = ["go"]
-  revision = "5c3871d89910bfb32f5fcab2aa4b9ec68e65a99f"
+  revision = "fd36f4220a901265f90734c3183c5f0c91daa0b8"
```
gz-c: Was a new dep added? I don't see one; if not, don't update the deps yet, unless its necessary, but you can do it in another PR

src/api/README.md
```diff
@@ -1539,6 +1543,126 @@ Result:
...
 ```
 
+## Filesystem APIs
+
+### Save data
+
+API sets: `FILESYSTEM`
+
+```
+URI: /api/v2/data/save
```
gz-c: The endpoints still aren't updated to use HTTP verbs

src/api/filesystem.go
```diff
@@ -0,0 +1,181 @@
...
+	Filename string                 `json:"filename"`
+	Data     map[string]interface{} `json:"data"`
+	Update   bool                   `json:"update"`
+}
+
+// Save arbitrary data to disk
+// URI: /api/v2/data
+// Method: POST
+// Args:
+//     filename: filename [required]
```
gz-c: Filename should not be part of the API, use a single known filename, like data.db uses. Consider this another database file

gz-c:
Can you update the PR description to reflect the URLs

src/api/README.md
```diff
@@ -1648,6 +1652,119 @@ Result:
...
+Args:
+    keys: list of keys to retrieve [required]
+```
+
+Get data from data file on disk
+
+Example:
+
+```sh
+curl -X GET \
```
gz-c: Isn't `-F` for formdata? Should be a query string

src/api/README.md
```diff
@@ -1648,6 +1652,119 @@ Result:
...
+```
+
+Delete data from data file on disk
+
+Example:
+
+```sh
+curl -X DELETE \
+  http://127.0.0.1:6420/api/v2/data \
+  -F 'keys=key1,key2,key3'
```
gz-c: Should be query string

src/api/filesystem.go
```diff
@@ -0,0 +1,143 @@
...
+
+// Save arbitrary data to disk
+// URI: /api/v2/data
+// Method: POST
+// Args:
+//     data: arbitrary data to save [required]
+//     update: update existing values [optional]
+func dataSaveHandler(gateway Gatewayer, w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		wh.Error405(w)
```
gz-c: wrong type of error responses for v2 API

src/api/filesystem.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	err = gateway.SaveData(params.Data, params.Update)
+	if err != nil {
+		resp := NewHTTPErrorResponse(http.StatusInternalServerError, err.Error())
+		writeHTTPResponse(w, resp)
+		return
+	}
+
+	wh.SendJSONOr500(logger, w, "success")
```
gz-c: Wrong type of response for v2 api.
No need to include "success" message; leave it empty. The `200 OK` is sufficient to indicate success

src/api/README.md
therealssj: both work apparently. I been using form data only in postman

src/daemon/gateway.go
```diff
@@ -541,6 +543,69 @@ func (gw *Gateway) GetWalletSeed(id string, password []byte) (string, error) {
 	return gw.v.Wallets.GetWalletSeed(id, password)
 }
 
+// SaveData saves arbitrary data to data file on disk
+func (gw *Gateway) SaveData(data map[string]interface{}, update bool) error {
```
gz-c: Don't put logic in the gateway. Gateway is only a shim to give the api package an interface for mocking

src/skycoin/config.go
```diff
@@ -188,6 +188,9 @@ type NodeConfig struct {
 	LogToFile   bool
 	Version     bool // show node version
 
+	// DataFilePath path of the file to store arbitrary data on disk
+	DataFilePath string
```
gz-c: `Data` is ambiguous, we need better terminology. Maybe `Keystore`

src/skycoin/config.go
```diff
@@ -535,7 +544,7 @@ func buildAPISets(c NodeConfig) (map[string]struct{}, error) {
 		api.EndpointsTransaction,
 		api.EndpointsPrometheus,
 		api.EndpointsNetCtrl,
-		// Do not include insecure or deprecated API sets, they must always
+		// Do not include insecure, deprecated or filesystem API sets, they must always
```
gz-c: I think the filesystem API can be left in the main grouping

src/api/filesystem.go
```diff
@@ -0,0 +1,152 @@
...
+	var params saveDataRequest
+	err := decoder.Decode(&params)
+	if err != nil {
+		resp := NewHTTPErrorResponse(http.StatusBadRequest, err.Error())
+		writeHTTPResponse(w, resp)
+		return
+	}
+	defer r.Body.Close()
+
+	if params.Data == nil {
```
gz-c: len(params.Data) == 0

src/daemon/gateway.go
gz-c: Belongs in its own package. Furthermore, none of this code is tested?

gz-c: The `update` flag makes no sense. If `update` is false it is a no-op?

gz-c: I removed daemon.Gateway; the new way to do this is:

- New package for managing the keystore
- New instance of keystore manager created during coin init
- Instance of keystore manager passed to api.NewGateway
- `Keystorer` interface added to api/gateway.go

gz-c:
Will be handled byb #2223

PR #2135, opened 2019-02-08
Disable the github version check

gz-c:
Since the version check wasn't working anyway, I'm keeping it disabled until we switch to storing the version in a file on our domain version.skycoin.net.


gz-c:
@senyoret1 This only blocks api.github.com in electron; should we also disable the code so that it doesn't run in the regular browser?

Senyoret1:
@gz-c In fact, https://github.com/skycoin/skycoin/pull/2102 solved the problem and the verification was working on both the web version and the Electron version. However, I added a new PR in case you want to disable the verification in both versions anyway: https://github.com/skycoin/skycoin/pull/2138

PR #2140, opened 2019-02-08
Use a double splash screen

Senyoret1:
Changes:
- When the Electron version of the desktop wallet is opened, after hidding the splash screen the window stays blank for a while before displaying the application. This happens because when the splash screen is hidden, the Angular application is opened and it is necessary to wait a long time for it to load. This PR adds a second splash screen inside the Angular application, which prevents the screen from being blank for a long time. The second splash screen has the same design as the first one. The transition is not perfect, since there is a white flash while going from one to the other, but the change is at least a little better than the current behavior.

No


mahansky:
What if we combined this with electron's splash through IPC - basically leave these changes for standalone build but also add a code that would leave the electron's splash running until it receives message from angular (which would emit message only if isElectron is true and in ngOnInit)?

Senyoret1:
I was checking what options we have and the problem is a little complicated. I do not think it's possible to leave the first splash page until the Angular app finishes loading, because the splash page is in a file that is loaded from a URL and to start loading the Angular app it is necessary to go to another URL. I also thought about using only the splash page of the Angular app (waiting for the node to finish initializing), but it is not a good option either because the node may take a long time before making the angular app's URL available.

Perhaps a simple option would be to create 2 different splash pages (with different content), but both with a white background. That way, when switching from one to another the white flash would be much less noticeable and would look more like a change of state. @mahansky, what do you think about this?

Senyoret1:
I tried to design a second loading screen, but nothing convinced me, so I simply modified the main Electron file so that the screen stays black when going from one page to another and I inlined the images to fix a loading problem that appeared occasionally. With those changes the Pr works a little better.

PR #2145, opened 2019-02-09
Stop showing multiple warnings in the same banner

### LINKED ISSUE #2142: Multiple warnings show in the same banner
When the wallet is both syncing and there are pending transactions, the messages are displayed in the same banner horizontally

Senyoret1:
Fixes #2142 

Changes:
- Now when the wallet is both syncing and there are pending transactions, only the syncing warning is shown.

No


gz-c:
Actually I think the pending transactions warning is more important; you can't send while there are pending transactions, but an unsynced blockchain doesn't prevent you from sending.

PR #2147, opened 2019-02-09
Improved transactions processing in the history

### LINKED ISSUE #2128: Possible error in transaction details for desktop app, check out for more details
Possible error in transaction details for desktop app(under develop with HEAD to 58bac2f9a3dd2db999d7b02075a0ef68be6731b0): https://github.com/skycoin/skycoin/tree/develop/electron#for-linux

I was execute the following steps:
- 1 Create an initial wallet (`SkyDev`) and put 20 $KY in it.
- 2 Create a hardware wallet(`HWSkyDev`) importing the `SkyDev` seed, note that the "two wallets" are the same.
- 3 Create new empty wallet (`SkyDevReciver`).
- 4 Send 5 $KY `HWSkyDev -> SkyDevReciver`.

From the mobile app in transaction details I can see the `-5` $KY as expected, but in the desktop app the result is different `-10`. The following pictures show you more details about this.

wallets

transactions

transactio details

android-mobile-app

PD: The balance is ok at all, the `-10` is in desktop app transaction details.

Senyoret1:
Fixes #2128 

Changes:
- When coins were moved between wallet addresses, the history showed that the user sended as many coins as there were in the inputs. Now that case is treated differently, and the UI shows `Internally moved`, instead of `Sent` or `Received`.

No

### DIFF
--- a/src/gui/static/src/app/app.datatypes.ts
+++ b/src/gui/static/src/app/app.datatypes.ts
@@ -31,6 +31,7 @@ export class Transaction {
   txid: string;
   hoursSent?: BigNumber;
   hoursBurned?: BigNumber;
+  coinsMovedInternally?: boolean;
 }
 
 export class PreviewTransaction extends Transaction {
--- a/src/gui/static/src/app/components/pages/send-skycoin/send-preview/transaction-info/transaction-info.component.html
+++ b/src/gui/static/src/app/components/pages/send-skycoin/send-preview/transaction-info/transaction-info.component.html
@@ -23,7 +23,7 @@
         <div class="-data">
           <span>{{ 'tx.hours' | translate }}:</span>
           <span>
-            {{ transaction.hoursSent.decimalPlaces(0).toString() | number:'1.0-0' }} {{ !isPreview && transaction.balance.isGreaterThan(0) ? ('tx.hours-received' | translate) : ('tx.hours-sent' | translate) }}
+            {{ transaction.hoursSent.decimalPlaces(0).toString() | number:'1.0-0' }} {{ hoursText | translate }}
             |
             {{ transaction.hoursBurned.decimalPlaces(0).toString() | number:'1.0-0' }} {{ 'tx.hours-burned' | translate }}
           </span>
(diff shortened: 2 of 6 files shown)


olemis:
does this fix the error reported in #2128 ?

olemis:
@stdevAlDen could you please try this patch and report back whether it fixes the original issue ?

mahansky:
It looks like this never executes (at least in my tx history) which makes every internal tx 0 SKY 0 CH
https://github.com/skycoin/skycoin/blob/52bae358241e9f2410344d3c3b937352691f373e/src/gui/static/src/app/services/wallet.service.ts#L474

That aside, I noticed you changed a few conditions to bignumber methods, there is a few more if you want to do it in this PR: https://github.com/skycoin/skycoin/blob/52bae358241e9f2410344d3c3b937352691f373e/src/gui/static/src/app/components/pages/transaction-list/transaction-list.component.html#L47

The use of `Map`s is not really necessary as you never make use of the boolean part

olemis:
@mahansky I guees the `Map` is used for `O(1)` address lookup (rather than list `O(n)` ).

Senyoret1:
@mahansky That part does run. If the transaction is outgoing, but all the outputs go to the input addresses and/or addresses that are in the wallets of the input addresses, then the code can not find any coin going out and this part is executed:

https://github.com/Senyoret1/skycoin/blob/e6be7aff3ebf2dd8535896914bd40369aa7b8fa6/src/gui/static/src/app/services/wallet.service.ts#L463-L478

It sets `coinsMovedInternally` to `true` to idicate that the transaction is not going out or in, but simply moving coins internally, and calculates the moved coins with the outputs that were not sent to an input address. That way, if you send 2 coins from address A to address B (both in the same wallet), the history will say that you moved 2 coins, but if you send 2 coins from address A to the same address A the history will say that 0 coins were moved, since there really was not any “movement” because the coins remain in the same address, but some hours were burned (which can be seen when opening the details of the transaction).

About the second point, in reality that part had to be eliminated.

About the third point, @olemis is correct, I used maps to perform searches with `O(1)` complexity instead of `O(n)`. In fact I am not sure if that makes any notable difference in this particular case, but it is a habit.

mahansky:
Okay makes sense. There is still this one comparison that doesn't use bignumber
https://github.com/skycoin/skycoin/blob/e6be7aff3ebf2dd8535896914bd40369aa7b8fa6/src/gui/static/src/app/components/pages/transaction-list/transaction-list.component.html#L48

Issue #1965, opened 2018-10-17
Add support for watch wallets


gz-c:
Watch wallets only contain addresses/public keys and do not contain seeds or private keys.

We can create wallet files with the type `"watch"`.  This wallet will not be able to perform any signing operations or generate new addresses automatically.

There are multiple ways we can create a watch wallet:

* Import a list of addresses
* Import a list of public keys
* Generate addresses from a seed one time, and discard the seed
* Import an exported watch wallet file, exported from a deterministic wallet

gz-c:
This requires a written plan since there are multiple approaches and also involves UI changes.

Issue #1513, opened 2018-05-17
API endpoints to create unsigned transaction and sign unsigned transaction from wallet


gz-c:
The endpoint to create an unsigned transaction will be like `POST /wallet/transaction`, except that it will not sign the transaction (so the password will not be required).

The endpoint to sign and unsigned transaction will accept a rawtx and wallet data necessary to sign the transaction with.

This will allow us to implement electrum's send page style, which has "Preview" and "Send" buttons. While "Send" directly sends (after entering password), "Preview" offers the option to "Sign" the previewed transaction, which then displays a signed, encoded transaction that the user can take and broadcast from elsewhere. After "Sign", the user also has the option to "Broadcast".

gz-c:
Useful for web and mobile wallets and exchange integrations

gz-c:
We'll also extend the signing to be able to specify specific inputs to sign, which is necessary to support coinjoin

olemis:
@stdevStark should be submitting something in  the next few days

PR #2159, opened 2019-02-11
Create unsigned transaction from wallet

gz-c:
Fixes part of #1513 (the other part is signing an unsigned transaction from a wallet)

Changes:
- Add `unsigned` as an option to `POST /api/v1/wallet/transaction` to create an unsigned transaction

Later, this can be combined with watch wallets (#1965) to facilitate offline transaction signing

For transaction verification, a `type TxnSignedFlag bool` is added to avoid using a bare `bool` in the verification APIs, which should make reading and assessing the code easier.  The alternative would be making a parallel verification API for unsigned transactions, which is often preferred instead of a `bool` parameter, but the amount of code duplication is onerous, and the code that would be duplicated is of a sensitive, precise nature. The difference in transaction verification between signed and unsigned transactions is minimal.

The differences between an unsigned and signed transaction:

* Unsigned has a signature array of the correct length, but all `cipher.Sig` objects are the null signature (`== cipher.Sig{}`)
* Since it has null signatures, the transaction hash value will change once signed

Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+- Add `unsigned` option to `POST /api/v2/wallet/transaction` to create unsigned transactions from a wallet
+
 ### Fixed
 
 ### Changed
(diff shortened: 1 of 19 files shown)


gz-c:
#2160

PR #2167, opened 2019-02-14
Run electron-rebuild post install

mahansky:
Changes:
- Runs `electron-rebuild` automatically after `npm install` to rebuild native deps

### DIFF
--- a/electron/src/package.json
+++ b/electron/src/package.json
@@ -11,5 +11,8 @@
   },
   "devDependencies": {
     "electron-rebuild": "^1.8.2"
+  },
+  "scripts": {
+    "postinstall": "electron-rebuild"
   }
 }


Senyoret1:
Simply adding `electron-rebuild` does not work on Windows because the libusb dependency has a problem that prevents it from being recompiled in that OS. Actually that dependency is for a part of `hardware-wallet-js` that is not currently being used, so the error is solved if you replace the script with `electron-rebuild --only=\"node-hid\"`. If you do that, then `install-dependencies.sh` is no longer necessary, so it should be deleted and this should be changed to `npm install`:

https://github.com/skycoin/skycoin/blob/2bf1c1167b1b0e0e85db5a9cc61936437eb11ccf/electron/install-node-modules.sh#L13

Also, https://github.com/skycoin/skycoin/blob/develop/electron/src/README.md should be modified.

gz-c:
What is the reason for this?

gz-c:
I'm afraid this will break the build somehow again

PR #2168, opened 2019-02-14
Allow to add amounts in USD when creating transactions

### LINKED ISSUE #2105: Allow to add amounts in USD when creating transactions
In some cases, when sending coins it would be more convenient to enter the amounts in USD and have the system automatically perform the conversion to SKY.

Senyoret1:
Fixes #2105

Changes:
- Now when creating a transaction it is possible to set the value in SKY and USD.  The change was made in the simple form and in the advanced form.

usd

No


gz-c:
This needs a dynamic disabling or hiding; other fiber coins are not listed and have no price.  We could detect it from fiber params in /health

Senyoret1:
The controls are kept hidden while the price is `null`, so it should work well when the price service is down and if the service is never consulted (because the coin has no price yet). After adding the data to `/health` the changes should be simple.

Senyoret1:
I removed the code responsible for changing the text to red in the html file of the advanced form. I removed it instead of making it work because in this form the effect is much less obvious, because the user can enter coins in several fields, and there would not be an equivalent in the hours fields.

About the symbols, the idea was to indicate that when entering the value in SKY the UI shows just an approximation in USD, but when entering the value in USD the UI shows exactly how many SKY would be sent. However, certainly the meaning may not be obvious and end looking like an error for some users, so now it is showing the `~` sign in both cases.

About the changes in the height, I did not see it in my tests. If the problem is not solved with the symbol change, please give me some extra details to try to solve it.

gz-c:
Merge conflict

mahansky:
Adding `line-height: 10px;` to `.coins-value-label` fixes the problem. The `≅` entity is not in Skycoin's font so it renders with different line-height.

Senyoret1:
Oh, I did not notice the font does not have the symbol. I changed the symbol to `&#x007e;` (~
), which is included in the font, to prevent any browser from using an excessively large font and making `line-height: 10px;` to end up clipping part of the symbol or text, because we can't know what will the different OS/browsers do.

PR #2183, opened 2019-02-20
Create unsigned txn from outputs or addresses without a wallet

### LINKED ISSUE #1461: API endpoints to create unsigned transaction from unspent outputs
https://github.com/skycoin/skycoin-web/pull/100

External wallets such as the web wallet have to implement unspent output selection themselves.  The mechanism of unspent output selection can become complicated, it would be better and safer to provide an API to do this.

The API would accept a set of unspent outputs and construct an unsigned transaction based upon other parameters. The other parameters would be similar to /wallet/spend.  

#1460 should probably be done first, to get the code for outputs sorted out.

### LINKED ISSUE #1937: Run CLI integration tests with coverage

gz-c:
Fixes #1461 
Fixes #1937 

Changes:
- Add `POST /api/v2/transaction` to create an unsigned transaction from unspent outputs or addresses without a wallet. It functions the same as `POST /api/v1/wallet/transaction` in terms of choosing which unspents to use.
- `POST /api/v1/wallet/transaction` parameters are slightly modified: the `"wallet"` object in the request is merged to the top level of the request. This requires a UI update
- Add `POST /api/v1/resendUnconfirmedTxns` to the `WALLET` API set
- API sets are now configured per unique (method, urlpath) path, instead of only the urlpath. This is so that e.g. `POST /api/v2/transaction` can be enabled with the `TXN` API set while `GET /api/v2/transaction` can be enabled with the `READ` API set
- Fix a bug with verifying transactions for `POST /api/v2/wallet/transaction/sign`

Internal changes:

- Transaction creation logic moved from package `wallet` to a new package `transaction`
- `wallet.Service` create transaction entrypoint removed since `visor` uses `wallet.Service.ViewSecrets` to access the wallet directly. 
- `CreateTransactionParams` split up into 3 parts:
  - arguments to the visor create transaction functions (`walletID`, `password`)
  - arguments related to querying data from the db (`addresses`, `unspents`, `ignoreUnconfirmed`)
  - arguments related to transaction construction (`hoursSelection`, `to`, `changeAddress`)

Yes


olemis:
@gz-c @stdevMac could you provide some feedback to this change from the perspective of adding Swagger v2  annotations ?

gz-c:
One important thing regarding swagger is that there will eventually be

`POST /api/v2/transaction` and `GET /api/v2/transaction` and they do different things

gz-c:
@Senyoret1  I added minor changes to the frontend for the API change, can you review those changes?

PR #2190, opened 2019-02-22
Change the warning banner message priority

### LINKED ISSUE #2171: Warning banner message priority
Following #2142, #2145,

The pending transactions warning is more important, if only one is shown. Otherwise show both.

Senyoret1:
Fixes #2171 

Changes:
- The pending transactions warning now has priority over the synchronizing warning.
- If the user tries to send a transaction while there are pending transactions, a warning is shown and the user must explicitly confirm the operation:
warning

- A small problem that caused error messages to appear in the console when loading the system in the simple form was fixed.

Notes:
- @gz-c this PR does not prevent the user from sending transactions when there are pending transactions, but it shows an alert. Does it seem good to you or should I change it?

No

### DIFF
--- a/src/gui/static/src/app/components/layout/header/header.component.html
+++ b/src/gui/static/src/app/components/layout/header/header.component.html
@@ -48,12 +48,12 @@
     <div *ngIf="appService.error === 3">{{ 'header.errors.csrf' | translate }}</div>
   </mat-toolbar>
   <mat-toolbar class="notification-bar" *ngIf="hasPendingTxs || !synchronized">
-    <div *ngIf="hasPendingTxs && synchronized">
+    <div *ngIf="hasPendingTxs">
       {{ 'header.pending-txs1' | translate }}
       <a routerLink='/settings/pending-transactions'>{{ 'header.pending-txs2' | translate }}</a>
       {{ 'header.pending-txs3' | translate }}
     </div>
-    <div *ngIf="!synchronized">
+    <div *ngIf="!synchronized && !hasPendingTxs">
       {{ 'header.synchronizing' | translate }}
     </div>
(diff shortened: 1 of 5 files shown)


gz-c:
The message is erroneous, it is safe to send while there are pending transactions.

If `ignore_unconfirmed` is false, the request will fail.
If `ignore_unconfirmed` is true, the request will exclude outputs that are spent by an unconfirmed transaction.

Senyoret1:
Oh, I misunderstood the comment in https://github.com/skycoin/skycoin/pull/2145#issuecomment-464046612. I did some testing and saw what the current behavior is.

Currently when creating transactions `ignore_unconfirmed` is `false`, so I can remove the warning and simply let the error returned by the node to be displayed. However, the web wallet creates the transactions locally, so it may be a good idea to place an error that prevents sending coins if any of the unspent outputs of the selected wallet is in an unconfirmed transaction (like the node API does when `ignore_unconfirmed` is true). Even after mergin https://github.com/skycoin/skycoin/pull/2183 this would be necessary for the web wallet to remain compatible with nodes from version 0.24

@gz-c Is this fine or should I instead change `ignore_unconfirmed` to `true`?

gz-c:
web wallet compatibility with v24 won't matter. ignore unconfirmed true is probably a better option, but we should test the ui for what happens when there are no confirmed outputs left

Senyoret1:
Ok, I removed the warning and created a new issue for future changes: https://github.com/skycoin/skycoin/issues/2192

I already did some preliminary testing and the node returns an error saying that there are not enough coins when there are not enough outputs outside the unconfirmed pool, but I think it is better to solve some things first to avoid problems (the list is in the issue).

PR #2193, opened 2019-02-24
Confirm hardware wallet addresses

Senyoret1:
Changes:
- Now before being able to see the addresses of the hardware wallet for the first time it is necessary to confirm them first in the hardware wallet, just like it happens with Trezor:
c1
c2

- It is possible to re-show the address in the hardware wallet by using the menu on the right:
c3

- This change does not prevent sending coins before confirming the addresses.

No

### DIFF
--- a/electron/src/hardware-wallet.js
+++ b/electron/src/hardware-wallet.js
@@ -160,8 +160,8 @@ ipcMain.on('hwGetFeatures', (event, requestId) => {
   );
 });
 
-ipcMain.on('hwGetAddresses', (event, requestId, addressN, startIndex) => {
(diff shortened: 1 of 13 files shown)


olemis:
@Senyoret1 are features to support hw wallet not added in CHANGELOG ?

Senyoret1:
As far as I know, no. If necessary, I can add them, but I think that nothing has been added about the hardware wallet, so maybe everything can be summarized as "integration with the hardware wallet".

gz-c:
@olemis @Senyoret1 yes, for the changelog, when the hardware wallet integration is done it will be summarized with a single entry

PR #2203, opened 2019-02-25
Improved seed validation

Senyoret1:
Changes:
- Now the code in charge of validating the seeds when creating/loading a wallet verifies if the words entered by the user are in the BIP39 dictionary.

- Small improvements in `Bip39WordListService`.

No

### DIFF
--- a/src/gui/static/src/app/components/pages/wallets/create-wallet/create-wallet-form/create-wallet-form.component.ts
+++ b/src/gui/static/src/app/components/pages/wallets/create-wallet/create-wallet-form/create-wallet-form.component.ts
@@ -2,6 +2,7 @@ import { Component, OnInit, OnDestroy, Input } from '@angular/core';
 import { FormControl, FormGroup, Validators } from '@angular/forms';
 import { ISubscription } from 'rxjs/Subscription';
 import { ApiService } from '../../../../../services/api.service';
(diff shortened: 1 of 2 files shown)


gz-c:
This validation is performed in the frontend? What about the `/api/v2/wallet/seed/verify` endpoint?

Senyoret1:
Sorry. Modified to use `/api/v2/wallet/seed/verify`

gz-c:
@Senyoret1 What is `Bip39WordListService`? We don't need the wordlist in the frontend do we?

Senyoret1:
@gz-c `Bip39WordListService` was originally created for the hardware wallet integration. When restoring a seed using a hw wallet, it is used to block the user from entering words that are not part of the BIP39 dictionary (as the hw wallet does not accept other words). It was part of the procedure for checking the seeds because I originally created it for the web wallet ( https://github.com/skycoin/skycoin-web/pull/536 ) and in that wallet it is not possible to call the `/api/v2/wallet/seed/verify` endpoint. In the second commit of this PR I stoped using it for validating the seeds, but I did forget to remove the references in `create-wallet-form.component.ts` (all usage in that file was removed, only the reference was left). This PR also includes a performance improvement for `Bip39WordListService`, which is still useful for the hardware wallet integration.

PR #2204, opened 2019-02-26
Update bip39 package to newer API

gz-c:
Changes:
- `POST /api/v2/wallet/seed/verify` returns an error if the seed's checksum is invalid
- Increase the detail of error messages for invalid seeds sent to `POST /api/v2/wallet/seed/verify`
- Move package `github.com/skycoin/skycoin/src/cipher/go-bip39` to `github.com/skycoin/skycoin/src/cipher/bip39`

The bip39 package was updated to have the latest changes from https://github.com/tyler-smith/go-bip39 plus some changes to the API for our needs.  The `NewSeed` method is now exposed, which was commented out before, possibly due to the use of `pbkdf2`. However, `pdkdf2` has been manually vendored in the `cipher` package since the introduction of wallet encryption.

The `NewSeed` method is a prerequisite for correct bip39 spec compatibility which specifies the use of `pbkdf2` and the mechanism for adding a password to a seed.  We'll also need the `NewSeed` method as a leading point into bip44 HD wallets.

We do not use the `NewSeed` method yet, but eventually this will be used for bip44 wallets.

The updated bip39 package includes wordlists for other languages. However, other languages are not supported yet, because issues regarding unicode normalization and word boundary separators have to be resolved.  The original `go-bip39` package (https://github.com/tyler-smith/go-bip39) has a branch with a v2 refactor for handling all languages properly but it hasn't been updated or merged in some time.  We may bring that code in later, when we add multi-language seed support.

Yes


src/api/integration/wallet_test.go
```diff
@@ -562,7 +562,7 @@ func TestVerifyWallet(t *testing.T) {
 
 	// check with incorrect seed
 	isBip, err = c.VerifySeed("nut ")
-	testutil.RequireError(t, err, "seed is not a valid bip39 seed")
+	require.Equal(t, bip39.ErrSurroundingWhitespace, err)
```
iketheadore: The `err` returned by `c.VerifySeed` is a different error type.

gz-c: Fixed; test had failed

PR #2221, opened 2019-03-03
Modify the procedure for creating transactions (front-end)

### LINKED ISSUE #2184: Change UI flow for preview transactions
Currently the UI does:

- create txn
- inject txn

for both preview and send.

For preview, this can be changed to:

- create txn (unsigned)
- sign txn
- inject txn

Which allows encrypted wallets to preview their transaction without entering a password until they are ready to sign and inject it

Senyoret1:
Fixes #2184 

Changes:
- Now when sending coins, if the user selects the preview option, it is not necessary to enter the password or confirm the transaction in the hardware wallet before viewing the preview.

- When the preview option is selected, the transaction is created with `/transaction`. `/wallet/transaction` is not used because it is necessary to enter the password even when `unsigned` is `true`.

- Now the advanced form works with the hardware wallets.

- The ID of the transaction was removed from the preview, since it changes after signing and sending the transaction.

Notes:
- This PR includes changes in important parts of the wallet, so it should be reviewed carefully. In any case, I plan to review it again in a day or so.

No


gz-c:
@Senyoret1 merge conflict

Senyoret1:
Conflicts solved

PR #2223, opened 2019-03-03
Implement transaction notes API in backend 

### LINKED ISSUE #1256: Implement transaction notes API in backend
Frontend design has a "note" field when sending, but this does nothing because the backend does not support it.

Backend APIs should expose method(s) for setting and retrieving notes for specific transactions.  Notes should be saved in the data dir.

nkryuchkov:
Fixes #1256

Changes:
- Implemented transaction notes API in backend with the following API:
`GET /api/v2/notes` returns all notes
`GET /api/v2/note?txid=` returns note by txid
`POST /api/v2/note` creates or replaces a note by txid
`DELETE /api/v2/note?txid=` deletes a note by txid
- Implemented tests for the implementation
- Added documentation for the implementation

Yes


src/api/README.md
```diff
@@ -1642,6 +1642,126 @@ Result:
...
+        "a5cf149da9cab9fdff681cec9fe83983aada218a46e26292a2c977ceff5bb1a5": "note1",
+        "db6fec68266296fcf6bf98a26cf25d86c83bfc31b8248575724977d90426addd": "note2",
+        "fef07801a566c3eafd680c9d29ccc18657c600e8b9d8f2c0eb89e3c98f5019c4": "note3"
+    }
+}
+```
+
+### Get transaction note by transaction id
+
+API sets: `READ`
```
olemis: Should be added to `NOTE` as well . Otherwise explicitly enabling/disabling `NOTE` will note affect `GET` endpoints . For instance , the way it works now disabling everything and enabling `NOTE` API set , will only enable `POST` and `DELETE` , but would be impossible to `GET` . If that's expected behavior , ok.

API sets are classifiers . Multiple values possible for a single endpoint .

src/api/README.md
```diff
@@ -1642,6 +1642,126 @@ Result:
 }
 ```
 
+## Transaction Note APIs
+
+### Get all existing notes
+
+API sets: `READ`
```
olemis: Should be added to `NOTE` as well .

src/api/README.md
```diff
@@ -1642,6 +1642,126 @@ Result:
...
+    }
+}
+```
+
+### Get transaction note by transaction id
+
+API sets: `READ`
+
+```
+URI: /api/v2/note
```
olemis: @gz-c is it possible for URL to be `/api/v2/note/{txid}` e.g. `http://127.0.0.1:6420/api/v2/note/a5cf149da9cab9fdff681cec9fe83983aada218a46e26292a2c977ceff5bb1a5` ?

src/api/gateway.go
```diff
@@ -19,14 +20,16 @@ type Gateway struct {
 	*daemon.Daemon
 	*visor.Visor
 	*wallet.Service
+	*note.Manager
 }
 
 // NewGateway creates a Gateway
-func NewGateway(d *daemon.Daemon, v *visor.Visor, w *wallet.Service) *Gateway {
+func NewGateway(d *daemon.Daemon, v *visor.Visor, w *wallet.Service, n *note.Manager) *Gateway {
```
olemis: I'm not sure about adding a new parameter here . Could this be implemented by keeping signature as it used to be and adding a new `GetNotesManager()` pointer receiver (or alike) ?

src/api/README.md
nkryuchkov: You're right, thanks. Fixed.

src/api/README.md
nkryuchkov: Fixed.

nkryuchkov:
> Needs to be mentioned in CHANGELOG

Fixed.

src/api/gateway.go
nkryuchkov: It could be. But it wouldn't look similar to other dependencies then.

src/api/README.md
gz-c: I was thinking it could support other types too (uxid, address).  Maybe this should be combined into the general key-value store and let the application put any string they want.

src/note/note.go
```diff
@@ -0,0 +1,93 @@
...
+	ErrNoteAPIDisabled = NewError(errors.New("wallet api is disabled"))
+	// ErrNoteNotExist is returned if a note does not exist
+	ErrNoteNotExist = NewError(errors.New("note doesn't exist"))
+	// ErrInvalidTxID is returned if a transaction ID is not valud
+	ErrInvalidTxID = NewError(errors.New("invalid transaction ID"))
+)
+
+const (
+	// NoteExt file extension
+	NoteExt = ".txnote"
```
gz-c: `".notes"`

src/note/note.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+// LoadNotes Loads all notes contained in notes dir. If any regular file in notes
+// dir fails to load, loading is aborted and error returned. Only files with
+// extension NoteExt are considered.
+func LoadNotes(dir string) (map[string]string, error) {
```
gz-c: I think there should only be one notes file. Not sure if we should allow multiple and merge them or not.

src/api/README.md
gz-c: @olemis we don't follow the `/path/to/{var}` convention for the other endpoints so we won't do it here

src/api/gateway.go
gz-c: @olemis the extra parameter here is the correct way

src/note/note.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Save saves the note to the given dir
+func Save(dir, txID, note string) error {
+	fileName := txID + NoteExt
+	path := filepath.Join(dir, fileName)
+	return file.RewriteBinary(path, []byte(note), 0600)
+}
+
+// Remove removes the note from the given dir
```
gz-c: Can delete (we don't have any methods that delete files because of bug potential and testing limitations)

src/note/note.go
```diff
@@ -0,0 +1,93 @@
...
+func loadNote(path string) (string, error) {
+	data, err := file.LoadBinary(path)
+	return string(data), err
+}
+
+// Save saves the note to the given dir
+func Save(dir, txID, note string) error {
+	fileName := txID + NoteExt
+	path := filepath.Join(dir, fileName)
+	return file.RewriteBinary(path, []byte(note), 0600)
```
gz-c: Can use `file.SaveJSON` passing the map[string]string directly

nkryuchkov:
@gz-c , so we use a "type" as a file name, right? But should these files be connected to the specific wallet? Since, they will not be if we have the "type" as a file name. And this way the user will see the note to all the transactions from all of his wallets. And what should we use the "category" for? If "type" is for file name and "key" is for the actual key, I do not understand the purpose of the "category"

gz-c:
* Notes file will be for all wallets, not per wallet
* "category" or "type" are options for the same parameter name. Call it either one

PR #2224, opened 2019-03-05
Verify spendable tx cli

### LINKED ISSUE #788: CLI command to verify if transaction is spendable
The `decodeTransaction` or `transaction` commands might already do this. If this is the case, then update the documentation for them.

Otherwise, add a new command `verifyTransaction` that checks the transaction against the blockchain (similar to the checks that `InjectTxn` make)

nkryuchkov:
Fixes #788 

Changes:
- Added the CLI command verifyTransaction [txid]

Yes


nkryuchkov:
@gz-c I got it fixed

PR #2225, opened 2019-03-06
Modify Visor constructor

mihis:
1. Visor's constructor has been modified to accept instances of interfaces.
2. Code from old constructor has been moved to **src/skycoin/skycoin.go**
3. Small refactoring of **src/skycoin/skycoin.go** to make things more readable.

### DIFF
--- a/src/skycoin/skycoin.go
+++ b/src/skycoin/skycoin.go
@@ -30,6 +30,7 @@ import (
 	"github.com/skycoin/skycoin/src/util/logging"
 	"github.com/skycoin/skycoin/src/visor"
 	"github.com/skycoin/skycoin/src/visor/dbutil"
+	"github.com/skycoin/skycoin/src/visor/historydb"
 	"github.com/skycoin/skycoin/src/wallet"
 )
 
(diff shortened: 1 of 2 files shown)


src/skycoin/skycoin.go
```diff
@@ -390,27 +489,82 @@ func (c *Coin) ConfigureVisor() visor.Config {
 	return vc
 }
 
-// ConfigureWallet sets the wallet config values
-func (c *Coin) ConfigureWallet() wallet.Config {
-	wc := wallet.NewConfig()
+// initHistory
+func (c *Coin) initHistory(
```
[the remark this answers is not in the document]
mihis: Will make it public then.

src/skycoin/skycoin.go
```diff
@@ -390,27 +489,82 @@ func (c *Coin) ConfigureVisor() visor.Config {
...
+		c.logger.WithError(err).Error("parseHistoryTo failed")
+		return err
+	}
 
-	return wc
+	return nil
+}
+
+// parseHistoryTo
+func (c *Coin) parseHistoryTo(
```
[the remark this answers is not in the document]
mihis: Will make will public then.

mihis:
New method will will be called `NewWith`

mihis:
@gz-c , can you review again?

src/skycoin/skycoin.go
```diff
@@ -368,8 +297,178 @@ func (c *Coin) initLogFile() (*os.File, error) {
...
+		log.Panic(err)
+	}
+
+	wc.CryptoType = cryptoType
+
+	return wc
+}
+
+// initVisor creates new instance of *visor.Visor
+func (c *Coin) initVisor(
```
gz-c: Can this be a method in `visor` package?

mihis: It can be only if it will return underlying instances of UnconfirmedTransactionPooler, Blockhchainer and Historyer.

Look at [this issue](https://github.com/watercompany/obelisk/issues/1l) and attached commit. I need access to instances of those interfaces to override some methods and define new one.

And in that case I can not create, for example, custom Blockchainer with overridden methods and put it into Visor. However, that is not needed as for now.

gz-c: The function can be broken down:

- creating Blockchain, UnconfirmedTxnPool, HistoryDB
- Initializing database contents (unspent pool, history db indexes)

So i think it can be rearranged so that the db.Update remains in the visor package.

PR #2228, opened 2019-03-10
Improved CSP

Senyoret1:
This PR improves de CSP header of the wallet.

Changes:
- All connections to servers other than the local server and https://api.coinpaprika.com are blocked.

- It is only possible to obtain content through calls to authorized URLs. Inline elements are prohibited, except for images and styles, since are used by Angular Material.

- Calls to `eval()` are blocked.

- The use of forms and plugins is blocked.

- It is not possible to change the base URL.

- The wallet can’t be opened inside a frame. 

No

### DIFF
--- a/src/api/middleware.go
+++ b/src/api/middleware.go
@@ -14,7 +14,15 @@ import (
 
 // ContentSecurityPolicy represents the value of content-security-policy
 // header in http response
-const ContentSecurityPolicy = "script-src 'self' 127.0.0.1"
+const ContentSecurityPolicy = "default-src 'self'" +
+	"; connect-src 'self' https://api.coinpaprika.com" +
+	"; img-src 'self' 'unsafe-inline' data:" +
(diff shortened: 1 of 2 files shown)


gz-c:
Test failure:

```
--- FAIL: TestContentSecurityPolicy (0.00s)
    --- FAIL: TestContentSecurityPolicy/enable_CSP_GET_/ (0.00s)
	Error Trace:	middleware_test.go:263
	Error:      	Not equal: 
	            	expected: "script-src 'self' 127.0.0.1"
	            	actual: "default-src 'self'; connect-src 'self' https://api.coinpaprika.com; img-src 'self' 'unsafe-inline' data:; style-src 'self' 'unsafe-inline'; object-src\t'none'; form-action 'none'; frame-ancestors 'none'; block-all-mixed-content; base-uri 'self'"
```

gz-c:
This is something that should be mentioned in the changelog

PR #2231, opened 2019-03-13
Limit the hw wallet addresses

Senyoret1:
This PR includes most of the changes recently discused about the hw wallet.

Changes:
- The recursive method for obtaining addresses from the hardware wallet was removed, since it is now more efficient to obtain several addresses at a time.

- The integration with the hardware wallet was limited to one address only. Due to this, the buttons for adding addresses and hiding the addresses without funds were hidden.

- Little code related to compatibility with several addresses was deleted, in case it is useful in the future.

- A warning message was added to the modal window that allows to recover an old seed in the hardware wallet, indicating that only seeds created with a hardware wallet should be used.

No

### DIFF
--- a/src/gui/static/src/assets/i18n/en.json
+++ b/src/gui/static/src/assets/i18n/en.json
@@ -370,6 +370,7 @@
     "restore-seed" : {
       "text": "Before proceeding, please select the number of words that the seed you want to recover has.",
       "check-text": "You can use this option to enter a seed and check if it is equal to the one in the hardware wallet. Before start, select the number of words the seed you want to check has.",
+      "warning" : "WARNING: to avoid potential problems, use only seeds created with a hardware wallet form the same brand/model.",
       "error-wrong-word": "Error: The retyped word does not match the one requested by the hardware wallet.",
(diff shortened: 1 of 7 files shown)


src/gui/static/src/assets/i18n/en.json
```diff
@@ -370,6 +370,7 @@
     "restore-seed" : {
       "text": "Before proceeding, please select the number of words that the seed you want to recover has.",
       "check-text": "You can use this option to enter a seed and check if it is equal to the one in the hardware wallet. Before start, select the number of words the seed you want to check has.",
+      "warning" : "WARNING: to avoid potential problems, use only seeds created with a hardware wallet form the same brand/model.",
```
mahansky:
```suggestion
      "warning" : "WARNING: to avoid potential problems, use only seeds created with a hardware wallet from the same brand/model.",
```
but I would probably delete the "from the same brand/model" part altogether

Senyoret1: I fixed it as suggested. I would like to mantain the last part because the integration with Ledger will not be compatible with our hardware wallet (at least initially). @gz-c , what do you think about this text?

gz-c:
yes only compatible with our hw wallet initially

Issue #2206, opened 2019-02-26
When scanning wallet addresses, use txn history, not current balance


gz-c:
When loading a wallet and scanning ahead for addresses, use the transaction history of the addresses instead of the current balance for deciding if an address has been used or not

Senyoret1:
The alert added in https://github.com/skycoin/skycoin/pull/2232 should be modified when solving this.

PR #2232, opened 2019-03-13
Show a warning if the user tries to add too many addresses

### LINKED ISSUE #2229: Show a warning if the user tries to add too many addresses
If the user generates alot of addresses and receives coins in an address at a high index like 20-30 can create issues when recovering wallet as the wallet might not scan that far for balance thereby showing wrong balance

Senyoret1:
Fixes #2229 

Changes:
- If the user tries to create too many addres, so that it would be posible to use the last address and then left 20 or more unused address before it, an alert is shown informing the user that there could be problems when restoring the wallet using the seed.

Notes:
- The code checks if the address has been used simply by checking if it has balance, as the node is doing now when restoring a wallet. This must be changed after solving https://github.com/skycoin/skycoin/issues/2206

No

### DIFF
--- a/src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.ts
+++ b/src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.ts
@@ -77,7 +77,7 @@ export class WalletDetailComponent implements OnDestroy {
             let lastWithBalance = 0;
             this.wallet.addresses.forEach((address, i) => {
               if (address.coins.isGreaterThan(0)) {
-                lastWithBalance = 1;
+                lastWithBalance = i;
               }
             });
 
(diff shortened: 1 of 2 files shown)


src/gui/static/src/app/components/pages/wallets/wallet-detail/wallet-detail.component.ts
```diff
@@ -73,7 +73,30 @@ export class WalletDetailComponent implements OnDestroy {
         .subscribe(response => {
           if (response) {
             this.howManyAddresses = response;
-            this.continueNewAddress();
+
+            let lastWithBalance = 0;
+            this.wallet.addresses.forEach((address, i) => {
+              if (address.coins.isGreaterThan(0)) {
+                lastWithBalance = 1;
```
mahansky:
```suggestion
                lastWithBalance = i;
```

PR #2234, opened 2019-03-15
Arbitrary data kv store

### LINKED ISSUE #1256: Implement transaction notes API in backend

nkryuchkov:
Fixes #1256 

Changes:
- Implemented arbitrary data key-value storage in the backend with the following API:
  GET `/api/v2/data?type=` return all values stored in the storage of type
  GET `/api/v2/data?type=&key=` return the value associated with the key from the storage of type
  POST `api/v2/data` adds or replaces value in the storage
  DELETE `api/v2/data?type=&key=` removes the value associated with the key from the storage of type

  Types allowed are: `txid` for notes, `client` for general data
- Added tests for the API
- Added documentation for the API   

Yes


src/api/storage.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+
+		resp := NewHTTPErrorResponse(http.StatusMethodNotAllowed, "")
+		writeHTTPResponse(w, resp)
+	}
+}
+
+// serves GET requests for /data enpdoint
+func getStorageValuesHandler(w http.ResponseWriter, r *http.Request, gateway Gatewayer) {
+	if r.Header.Get("Content-Type") != ContentTypeForm {
```
gz-c: Don't need to check content-type for `GET` requests

src/skycoin/config.go
```diff
@@ -259,9 +266,13 @@ func NewNodeConfig(mode string, node NodeParameters) NodeConfig {
 		WebInterfaceCert:  "",
 		WebInterfaceKey:   "",
 		WebInterfaceHTTPS: false,
-		EnabledAPISets:    strings.Join([]string{api.EndpointsRead, api.EndpointsTransaction}, ","),
-		DisabledAPISets:   "",
-		EnableAllAPISets:  false,
+		EnabledAPISets: strings.Join([]string{
+			api.EndpointsRead,
+			api.EndpointsTransaction,
+			api.EndpointsStorage,
```
gz-c: `EndpointsStorage` shouldn't be in the default set (applies to all config modes, see https://github.com/skycoin/skycoin/#configuration-modes), since it is intended for client only.

src/util/maputil/maputil.go
```diff
@@ -0,0 +1,13 @@
+// Package maputil provides Go maps related utility methods
+package maputil
+
+// Copy copies the map contents to the new map
+func Copy(data map[string]string) map[string]string {
```
gz-c: I would keep this in package kvstorage since it is specific to map[string]string

src/util/file/file.go
```diff
@@ -288,3 +288,9 @@ func CopyFile(dst string, src io.Reader) (n int64, err error) {
 	n, err = io.Copy(out, src)
 	return
 }
+
+// Exists checks whether the file exists in the file system
+func Exists(fn string) bool {
```
gz-c: Should this return `(bool, error)` in case the error is a filesystem error?

gz-c:
The binary run tests are consistently failing with exit status 2, check that the program can be run

gz-c:
#2266

PR #2235, opened 2019-03-17
Integrate hw wallet labels

Senyoret1:
Changes:
- Now the name of the hardware wallets in the UI is the label saved in the hardware wallets itself.

- The JS library was updated. Due to several changes made in that library, the script that installs the Electron dependencies was also updated (`install-dependencies.sh`)

No


Senyoret1:
The last 2 commits solve some problems caused by recent changes in the hw wallet related code, but there are still several problems that should be solved in other repos, like https://github.com/skycoin/hardware-wallet/issues/188 , https://github.com/skycoin/hardware-wallet-js/issues/105 and https://github.com/skycoin/hardware-wallet/issues/169#issuecomment-484668710 . Also, there is a problem while trying to change the wallet name when connecting the device for the first time, and Snyk is complaining for some of the recent changes, so more changes will be hended for this PR to be ready.

Senyoret1:
The last 3 PRs solve https://github.com/skycoin/hardware-wallet-js/issues/105 , the problem of trying to change the wallet name when connecting the device for the first time and the failing Snyk tests. For making the Snyk tests pass, the non-relevant problems were ignored, one dependency was updated, a patch was applied to another dependency (because it was not possible to update it at this moment) and another one was ignored, due to the lack of solution at this time. For the Snyk patch to be applied when installing the dependencies, Snyk was added to the `package.json` file.

The options to restore and confirm a seed still do not work. Likewise, the option to change the PIN always responds PIN mismatch, regardless of whether the operation was completed successfully or not. However, these errors are not due to changes in this PR and must be solved in other repositories.

With those changes this PR should be ready to be reviewed.

Senyoret1:

PR #2239, opened 2019-03-21
Small changes for the hw wallet integration

### LINKED ISSUE #2237: Change hw wallet pin setup symbol
A solid circle would be better I think
Screenshot 2019-03-20 at 6 16 20 AM

### LINKED ISSUE #2238: Send cancel command on X press
Send `cancel` operation on press of `X` otherwise the user has to take the wallet out and put in again before continuing 

Screenshot 2019-03-20 at 6 46 44 AM

Senyoret1:
Fixes #2237 
Fixes #2238 

Changes:
- The ability to close the modal windows for creating a new seed and restoring an old one was beign disable when asking for the number of words of the seed, but in fact it had to be blocked while waiting for the user to confirm the action on the device (as it happens in other parts of the hw wallet integration). The error was added in the PR for adding compatibility with seeds with 12/24 words. This PR solves the problem.

- The “#” simbol on the modal window for entering the PIN code were replaced with “•” simbols:
pin

No

### DIFF
--- a/src/gui/static/src/app/components/layout/hardware-wallet/hw-pin-dialog/hw-pin-dialog.component.html
+++ b/src/gui/static/src/app/components/layout/hardware-wallet/hw-pin-dialog/hw-pin-dialog.component.html
@@ -19,19 +19,19 @@
   </div>
   <div class="select-address-theme">
     <div class="num-pad-row">
-      <button mat-button color="primary" (click)="addNumber('7')"><span>•</span></button>
-      <button mat-button color="primary" (click)="addNumber('8')"><span>•</span></button>
-      <button mat-button color="primary" (click)="addNumber('9')"><span>•</span></button>
+      <button mat-button color="primary" (click)="addNumber('7')"><span>{{ buttonsContent }}</span></button>
+      <button mat-button color="primary" (click)="addNumber('8')"><span>{{ buttonsContent }}</span></button>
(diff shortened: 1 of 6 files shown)


src/gui/static/src/app/components/layout/hardware-wallet/hw-pin-dialog/hw-pin-dialog.component.html
```diff
@@ -19,19 +19,19 @@
   </div>
   <div class="select-address-theme">
     <div class="num-pad-row">
-      <button mat-button color="primary" (click)="addNumber('7')"><span>#</span></button>
-      <button mat-button color="primary" (click)="addNumber('8')"><span>#</span></button>
-      <button mat-button color="primary" (click)="addNumber('9')"><span>#</span></button>
+      <button mat-button color="primary" (click)="addNumber('7')"><span>•</span></button>
```
therealssj: can u define this as some sort of constant and than use that constant here instead?

Senyoret1:
The last 2 PRs update the hw wallet JS library and solve the remaining problems from https://github.com/skycoin/skycoin/issues/2238 (only present on MacOS). For the changes to take effect, it is necessary to reinstall the dependencies in `electron/src` using `install-dependencies.sh`.

jdknives:
This issue is still present?

Senyoret1:
If you mean if this still should be merged, the answer is yes

Senyoret1:

PR #2258, opened 2019-04-07
[WIP] Add the option for removing the hw wallet pin

Senyoret1:
Not ready for review.

Changes:
- Adds an option for removing the pin protection of the connected hw wallet:
rm
The implementation is a bit hacky, as it calls the function for changing the pin, waits for the user to introduce the current pin and then sets the new pin to `null`. Setting the new pin to `null` makes the hardware wallet to remove the pin.

No


Senyoret1:
The last commits changed the buttons layout:
rm

Senyoret1:
The method for deactivating the pin will be changed, so this is not ready to be merged

Senyoret1:

PR #2261, opened 2019-04-07
update travis to xenial

therealssj:
Fixes: #2260 
Update travis dist to `xenial`

no

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ before_script:
   - if [[ "$TRAVIS_OS_NAME" == "osx" && "$TRAVIS_PULL_REQUEST" == false ]]; then ./ci-scripts/add-key.sh; fi
 
 services:
-  - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then xvfb; fi
+  - xvfb
 
 script:
   - make check-newcoin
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -55,7 +55,9 @@ install:
 
 before_script:
   - if [[ "$TRAVIS_OS_NAME" == "osx" && "$TRAVIS_PULL_REQUEST" == false ]]; then ./ci-scripts/add-key.sh; fi
-  - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then export DISPLAY=:99.0 && sh -e /etc/init.d/xvfb start; fi
+
+services:
```
gz-c: Why is this changed?

gz-c:
Does not need to be in changelog, its irrelevant to users

.travis.yml
therealssj: that's how it works now in xenial. the old `before_script` execution gives error

therealssj: https://changelog.travis-ci.com/ubuntu-xenial-16-04-build-environment-update-86123

gz-c: The old way doesn't run on OS X, the new way does?   You link doesn't mention xvfb

PR #2273, opened 2019-04-13
Improvements for the password modal window

Senyoret1:
Changes:
- Now the password modal uses the snackbar as it should.
- The link for restoring the password is disabled while the operation is being done.
- The modal window stays open while decrypting the wallet in the forms for sending coins, which is more consistent with the rest of the application.
- Fixed a case in which the error response was not being send to the modal window, when it should.
- A new error text was added.

Notes:
- The behavior of the password modal window will change a bit while sending coins with https://github.com/skycoin/skycoin/pull/2221, so some additional changes could be necessary.

No


gz-c:
merge conflicts

Senyoret1:
The conflicts were fixed and some improvements that were necessary after https://github.com/skycoin/skycoin/pull/2221 were added.

PR #2274, opened 2019-04-15
Improvements for some messages of integration with the hw wallet

Senyoret1:
Changes:
- Now the error message that tells the user that it is not possible to send the transaction because it has too many inputs or outputs gives the user info about an option for sending the coins (split the transaction into several parts), instead of suggesting to contact support. With this change there are no longer 2 separate messages, because after merging https://github.com/skycoin/skycoin/pull/2221 an additional change will be made and it will not allow knowing if the problem was due to the quantity of inputs or outputs.

- Now the message that is displayed while confirming the seed does not asks the user to confirm the operation in the hardware wallet, because when this operation is made the user is not asked for confirmation.

No


Senyoret1:

Issue #2276, opened 2019-04-18
The /data endpoint always returns an error


Senyoret1:
When calling the `/data` endpoint, the response always has this error at the end:

```
{
    "error": {
        "message": "Method Not Allowed",
        "code": 405
    }
}
```

The error is there no matter what the result of the operation is. For example, if you add data, the answer is:

```
{}{
    "error": {
        "message": "Method Not Allowed",
        "code": 405
    }
}
```

If you recover the data, the response is like this:

```
{
    "data": {
        "key1": "val1",
    }
}{
    "error": {
        "message": "Method Not Allowed",
        "code": 405
    }
}
```

And if you do not set the data type, the response is this:

```
{
    "error": {
        "message": "type is required",
        "code": 400
    }
}{
    "error": {
        "message": "Method Not Allowed",
        "code": 405
    }
}
```

There are 2 json objects, and that causes problems when procesing the response.

PR #2277, opened 2019-04-18
Allow to change the language of the UI

### LINKED ISSUE #1981: Translate the wallet
It will be necessary to add the option to change the language and test if it works correctly in all screens. Probably the option should not be active by default, until analyzing well its maintainability.

### LINKED ISSUE #2242: Language switcher
Add ability to change the UI language.

We already have the i18n support

Senyoret1:
Fixes #2242
Fixes #1981

Changes:
- Controls were added to allow to change the language of the UI.
l1
l2
- The Spanish translation was added.
- The flags of the 4 languages that are available so far have been added.
- Added documentation on how to make the translation files available in the application.

Pending:

- The code still does not remember the language selected by the user, but selects English every time the application is opened. It will be solved after https://github.com/skycoin/skycoin/issues/2276
- It would be good to make a revision of the Russian and Chinese translations.

No


Senyoret1:
With the last two commits now the user's language selection is correctly saved and recovered. Additionally, if the user has not yet selected a language, the language selection modal window is shown just after the application is opened.

gz-c:
Merge conflict

Senyoret1:
The conflicts were solved. I also made the change to use `StorageService` and added an small code to check if the language selected by the user is valid.

I did not changed the flags as those are the ones in use in the web wallet and the explorer. If the change is going to be made, it should be done in all three repos.

gz-c:
@Senyoret1 this should be in the changelog

PR #2283, opened 2019-04-25
Allow to hide the wallets in the UI

### LINKED ISSUE #2226: Remove wallet in electron app GUI
**Describe the solution you'd like**
Each wallet in the list should have a button to remove it .

**Possible implementation**
Even if [there is intentionally no API method to delete a wallet](t.me/skycoindev/4194) it'd be nice to , at least hide it . If the later is the way to go then it should be possible to select a specific wallet to un-hide it.

Senyoret1:
Fixes #2226

Changes:
- Now the user can hide any wallet.
h1

- When there is one or more hidden wallets, at the bottom of the wallet list a new button is displayed, which allows the user unhide any wallet.
h2

- If the user hides all wallets, the wizard is displayed. If the user does not want to create or load a wallet in the wizard, but simply unhide a wallet, he can select the "Want to recover a hidden wallet?" link.
h3

- The information about which wallets must be hidden is stored using the `/data` endpoint.

No


gz-c:
Not doing this

PR #2295, opened 2019-05-07
Improvements for the Swaplab integration

### LINKED ISSUE #2290: Improvements for the Swaplab integration
This is a list with improvements that would be good to add to the Swaplab integration. I added everything in one issue to avoid flooding the repository, because the list is quite long but none of the points is really critical (some are quite simple). As the points can be fixed in groups, not all at the same time, I created a task list

**UI/UX:**

- [X] It would be good to add a more prominent indication for the user to know that the “you get” value is just an stimate. Currently a symbol is used to indicate it, but some users may not notice it and then complain.

- [X] It should be possible to change the “you get” value. When doing so, the amount of coins to send should be updated automatically. It is important to be careful with decimals when doing this. (Was not done, but some improvements used in the the Swaplab website were added to the UI)

- [X] `ExchangeCreateComponent` checks if the address is valid before leting the user to continue. That is not how it works in the forms for sending coins, so maybe, for consistency, the code should check the value only after the user tries to start the operation. That way the component would not have to use an async validation and the UI would respond faster. If this is not done, at least `ExchangeCreateComponent. validateAddress` should use a `debounceTime`, if possible, to make the UI respond faster. (The first approach was used)

- [X] Maybe the status component should show more data, like the destination address, for example. This is because the user totally lose access to most of the data that is shown in the initial form, data that may be usefull, specially taking into account that the status component is the one used to show the data about the operations when using the operations history, and the user could be trying to get information about a exchange made several months ago.

- [X] The links at the bottom of the exchange section do not have a lower margin, so if the window has to scroll, they touch the lower edge. A lower margin similar to the upper margin that they already have should be added.

- [X] The link for opening the history should be more prominent. Possibly it could be improved making it a little bigger, with an icon and without transparency.

- [X] When selecting an operation in the history, the code should not contact the remote server before opening the status component. In slow internet connections it appears that the app is not working correctly. It would be better if the status component is opened immediately, showing a loading control while the initial connection to the external server is made.

- [X] If the status component is being shown, selecting a different operation in the history does not make the status component to be updated with the info of the newly selected operation. This is because the status component makes all its initialization when it is being created and selecting a new operation only updates the value of `order`, without restarting the component. There are various options for solving this: one is restarting the component after selecting a new operation, and another one is converting `ExchangeStatusComponent.order` to a setter and acting accordingly every time the value is modified. (The second approach was used)

- [X] For consistency, maybe the buttons of the history modal window should not have the “details” blue text, as no other modal window in the wallet has something similar. However, it could be good to add the last known state of the operations (but we must be careful when doing it, since the user might think that the UI is showing an updated value). (The last known state was not added, to avoid confusing the user and overcomplicating the code. Other improvements were added)

- [X] When an invalid value is entered as the number of coins that will be sent, the “you get” field shows “NaN”. If would be better to show 0 or an error message. (0 is shown)

- [X] When opening the Exchange section, if a previous operation has been saved in `ExchangeService.lastOrder` and that operation was still active the last time the code knew about it, then the status of that operation, instead of the form, is shown. That is good, but there are some problems with the implementation. The first problem is that the value of `ExchangeService.lastOrder` is not saved in persistent storage, so the effect does not work if the user refreshes the page or closes the wallet. Adding the value of the last operation in the history to `ExchangeService.lastOrder` would not be an ideal solution, as the user could close the wallet while seing another operation. The other problem is that, as the status component does not have a “back” button, it is impossible to start a new operation before the current one is completed, unless the user refreshes the page or closes the wallet. Taking this second problem into account, it would be good to add a “back” button in the status component and ask the user for confirmation if it is pressed while the oepration is still active.

- [X] The different status of the operations could be somewhat confusing for some users, so it would probably be good to add a description to each one.

- [X] The XMR addresses are extremely long and the QR modal window does not show them properly.

- [X] When starting an operation with certain coins, in addition to sending the indicated amount of coins to the indicated address, the user must also include additional information, which Swaplab identifies as the “tag”, so the UI tells the user which “tag” should be used. The problem is that “tag” is the way in which the api identifies the value, but in reality the name “tag” could be weird and confusing for the user. In fact, when using XMR (one of the currently available pairs) the value should be identified as “Payment ID”, while when using XRP it should be called “Destination Tag”.

- [X] When the window is too narrow, the double button shown in the navigation bar in the “Send” and “Pending transactions” sections does not look too good.

- [X] The order of the operations in the history should be inverted (the most recent first), to be consistent with the transactions history section.

- [X] In the status component, maybe we should ask the user to save the operation ID, just in case, but not in a very intrusive way. (The change seemed excessive, since the user can use the history to retrieve the ID, so the change was reverted)

- [X] It may be good not to request the list of trading pair just when starting the form, but every so often, to update the price.

**Code:**

- [X] `ExchangeCreateComponent` calls the API to know how many decimals Skycoin is currently using. However, the data is already available in `BlockchainService. currentMaxDecimals`, so there is no need to call the API.

- [X] The subscriptions in `ExchangeCreateComponent` should be cleaned.

- [X] The `wallet-container` class should be removed from `exchange-history.component.html`.

- [X] The `lastKnownStatus` variable in `ExchangeService.status` does not seem to be necessary.

- [X] If the value of `ExchangeCreateComponent.defaultFromCoin` is not found in any of the pairs returned by the service, then `ExchangeCreateComponent.updateActiveTradingPair` sets `activeTradingPair` to `null` and then the UI shows a message indicating that the service is not available, even if the service returns other valid pairs. If the service returns valid pairs but the value of `ExchangeCreateComponent.defaultFromCoin` is not in the list, the correct behavior would be to select another coin as the default one (maybe the first in the list).

- [X] The way in which `ExchangeService.storeOrder` works is a bit dangerous. It first tries to get all the previous operarions from the persistent storage, then appends the new one at the end and, at last, saves all the operarions again in the persistent storage. The problem is that if the first step fails, the code asumes that there are no previous operarions and only saves the current one. If there is a temporary error causing the first step to fail, then the entire history of operations would be lost. The current approach was used because the API responds with an error when there is no data, but it would be good to differentiate that error from the others, in order to make the code safer.

- [X] Due to some changes that were made recently, there is no reason for `ExchangeComponent.showLast` being a function, and less for it to be a public one. Its code could be added directly in the only place where that function is called.

- [X] Maybe `ExchangeService.lastOrder` should be called `ExchangeService.lastViewedOrder` instead, as the current name was created when the implementation did not have a history and now the utility of the variable has changed.

- [X] `ExchangeService.status` returns an observable that tries to get the data again in case of a problem. However, in my tests (disabling the internet connection) it only tried again like 4 times. It must keep trying again indefinitely or show an error.

Senyoret1:
Fixes #2290

Changes:
- All the points in the original issue were addressed. Only the ability to modify the "You get" field was left out, but a visual indication was added in the UI, just as in the Swaplab website.

- The testing procedure described in https://github.com/skycoin/skycoin/pull/2039 was added to the code. I can be activated by setting `AppConfig.swaplabTests.activateTestMode` to true. The testing code can only be activated in development builds, it is ignored in production builds.

- Some additional improvements for things like the UI and error handling.

No


Senyoret1:
@mahansky , please take a look at this.

gz-c:
merge conflict

Senyoret1:
The conflicts were fixed.

@mahansky Thanks for the PR with additional changes, I merged it and corrected the texts. However, about the style of the history link, I think it should be more prominent. The style you added is more consistent with the links at the right, which makes the design more harmonious, but also makes the link harder to notice and I'm afraid of having many users simply ignoring it. To reach a mid-point, I made some more changes, in which I made the icon and the text a little larger, but not as much as before, in such a way that the link is not as inconsistent as before but it is a little easier to notice it. I hope the change seems appropriate.

PR #2300, opened 2019-05-09
Remove .zip release of osx electron build

### LINKED ISSUE #2063: Remove compressed .zip release of the OS X electron dmg

iketheadore:
Fixes #2063 

Changes:
- Remove compressed zip release of osx electron build

No

### DIFF
--- a/electron/build-electron-release.sh
+++ b/electron/build-electron-release.sh
@@ -46,8 +46,6 @@ if [ ! -z "$OSX64_ELN" ]; then
     if [[ "$OSTYPE" == "darwin"* ]]; then
         echo "run dist-mac"
         npm run dist-mac
-    elif [[ "$OSTYPE" == "linux"* ]]; then
-        npm run pack-mac
     else
         echo "Can not run build script in $OSTYPE"
     fi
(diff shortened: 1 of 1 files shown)


electron/build-electron-release.sh
```diff
@@ -60,7 +60,7 @@ if [ -e "mac" ]; then
         mv "${PDT_NAME}-${APP_VERSION}.dmg" "../${PKG_NAME}-${APP_VERSION}-gui-electron-osx-x64.dmg"
     elif [ -e "${PDT_NAME}.app" ]; then
         if [[ "$OSTYPE" == "darwin"* ]]; then
-            tar czf "../${PKG_NAME}-${APP_VERSION}-gui-electron-osx-x64.zip" "${PDT_NAME}.app"
+            rm -rf "${PDT_NAME}.app"
```
gz-c: I think this line needs to be in the higher if block, so that it applies to all host OS build machines

iketheadore: Actually we should remove the `npm run pack-mac` on linux, cause we now only need `dmg` for osx version. The `npm run pack-mac` is used on linux to build a osx `.app` file, and compress it as a `.zip` file. After removeing the osx app build on linux, we can remove the `OSTYPE` check here.

gz-c: Does linux/windows make a dmg file?

PR #2315, opened 2019-05-13
Improved code for showing amounts and coin identifiers

Senyoret1:
Changes:
- Now the format of the amounts of coins and hours that are shown in the UI is obtained using pipes. The first advantage of this is that a good amount of repeated code for calculating the correct number of decimals is eliminated, thus improving maintainability. The sencond one is that now the IU shows correctly "Coin Hour" and "Coin Hours" (singular and plural) as appropriate.

- Now the names of the coins and hours are saved in a code configuration file and not in the language files. This means that the way to identify the coins is more similar to the one used in the web wallet, which is more flexible to easily customize the wallet for Fiber coins in the future. In addition, this will make it much easier to integrate into the UI the change requested in https://github.com/skycoin/skycoin/issues/2082

No


Senyoret1:
This pr does not touch the translation files, so it could be merged for 0.26, but is not critical, so it could also be left for later.

gz-c:
#2366 will add the fiber parameter strings like "Skycoin", "Coin Hours", "SKY" etc

Senyoret1:
Good, I will make the changes for getting the names from the `/health` endpoint after that PR is merged

Senyoret1:
The change was made, now the strings are obtained from the `health` API endpoint. I also modified the explorer link to use the URL obtained from the `health` API endpoint.

With this change the pipe is not longer able to correctly show "1 Coin Hour", it shows "1 Coin Hours" (in plural). @gz-c this is not a big problem, but if you want it to be solved the singular version of the coin hours name will have to be added to the `health` endpoint.

gz-c:
#2375 will distinguish the plural and singular forms of Coin Hours

Senyoret1:
The latest commits add the following changes:

- Unnecessary code that was used to limit the number of decimals was removed (the new pipe does that automatically).

- The new pipe is now used in more places.

- The merge conflicts were fixed.

- The changes were added to the new parts added in https://github.com/skycoin/skycoin/pull/2405

PR #2327, opened 2019-05-15
Stop showing simple errors in the browser console

### LINKED ISSUE #2319: Multiple console errors (Safari)
Screen Shot 2019-05-14 at 10 10 56 PM

@Senyoret1 will some of these go away when we remove the hardware wallet code?

Senyoret1:
Fixes #2319

Changes:
- This PR deactivates the default error handler used by Angular, to prevent it to continue adding unnecessary entries in the browser console every time there is an error.

No

### DIFF
--- a/src/gui/static/src/app/app-error-handler.config.ts
+++ b/src/gui/static/src/app/app-error-handler.config.ts
@@ -1,0 +1,6 @@
+import { ErrorHandler, Injectable} from '@angular/core';
+
+@Injectable()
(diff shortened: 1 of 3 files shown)


gz-c:
Oh, its a default setting?  Should we leave it?  I thought it was leftover debugging code

Senyoret1:
It is a default setting, but not really an important one, at least from my point of view.

gz-c:
@mahansky do you want to keep the logging? I don't mind it, I thought it was leftover debugging code

PR #2329, opened 2019-05-16
Send a minimum of 1 coin hour when using auto share hours

### LINKED ISSUE #1604: Send a minimum of 1 coin hour when using auto share hours
In `POST /wallet/transaction`, send a minimum of 1 coin hour to all outputs when using auto share hours, if possible.  If there are not enough hours to send a minimum of 1 coin hour to all outputs, it is not an error.

In general the client should avoid sending 0 hours to destinations.  However, if they specify say `0.01` share factor, its possible they will send 0 hours when they could have easily sent at least 1 coin hour.

iketheadore:
Fixes #1604 

Changes:
- Update the POST `/wallet/transaction` endpoint to send a minimum of 1 coin hour to all outputs when possible. 

Yes

### DIFF
--- a/src/transaction/create.go
+++ b/src/transaction/create.go
@@ -144,6 +144,19 @@ func create(p Params, auxs coin.AddressUxOuts, headTime uint64, callCount int) (
 				toCoins[i] = to.Coins
 			}
 
+			// If sharefactor > 0 and sharehours < len(p.To), the sharehours cannot ensure each destination
+			// address has at least 1 coin hour. In this case, if we have enough hours, i.e the hours >= len(p.To),
+			// set the sharehours as len(p.To). If we don't have enough hours, i.e. hours < len(p.To),
+			// share all hours we have.
+			toNum := uint64(len(p.To))
+			if p.HoursSelection.ShareFactor.GreaterThan(decimal.New(0, 0)) && allocatedHours < toNum {
+				if hours < int64(toNum) {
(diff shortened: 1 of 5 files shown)


src/transaction/create.go
```diff
@@ -144,6 +144,19 @@ func create(p Params, auxs coin.AddressUxOuts, headTime uint64, callCount int) (
 				toCoins[i] = to.Coins
 			}
 
+			// If sharefactor > 0 and sharehours < len(p.To), the sharehours cannot ensure each destination
+			// address has at least 1 coin hour. In this case, if we have enough hours, i.e the hours >= len(p.To),
+			// set the sharehours as len(p.To). If we don't have enough hours, i.e. hours < len(p.To),
+			// share all hours we have.
```
gz-c: This should consider if there is change, then the change output should be guaranteed at least one hour, if sharefactor < 1

iketheadore:
Yes, this change will affect the `/api/v2/transaction` as well, all api that call the `transaction.Create` method will be affected.

gz-c:
Merge conflict

gz-c:
Can you update the changelog to say that it also affects `/api/v2/transaction` endpoint

gz-c:
I will continue this PR

gz-c:
Decided not to implement due to the complexity for marginal benefit

PR #2333, opened 2019-05-16
Delete the skycoin-lite lib and use a TypeScript version of the TX encoder

Senyoret1:
Changes:
This PR adds a TypeScript version of the transaction encoder found in `transaction_skyencoder.go`. The idea is to use it to remove the skycoin-lite lib.

The code still needs some changes and test, but I would like to add the changes after merging some hw wallet related PRs, and the test after checking how the unit tests are working.

The code is currently not being used anywhere, so it should not affect the wallet, it is just for being used in the future. However, it adds the  "base-x" npm package as dependency. The dependency contains only one small code file and was added using an specific version number (not asking npm to update it), so it should not be a problem for the security. The dependency has the “safe-buffer” npm package as internal dependency, but it was already being added to the wallet by some other packages.

No


gz-c:
This is important to have, to eliminate the skycoin-lite dependency

Senyoret1:
The last commit adds the following changes:

- The encoder is now totally functional is and being used for the hw wallet transactions.

- The encoder now has unit test. There are several test cases, encoding transactions with various numbers of inputs, outputs and signatures.

- The skycoin-lite lib was removed.

PR #2363, opened 2019-05-25
Try to send a minimum of 1 coin hour when using auto share hours

### LINKED ISSUE #1604: Send a minimum of 1 coin hour when using auto share hours

gz-c:
Fixes #1604

Changes:
- When using auto share hours, ensure at least one hour is sent to each output if possible
- If there aren't enough hours for each output to get one, the change output has priority, and the remaining outputs are ranked in order of the amount of coins they receive

Yes


gz-c:
This version doesn't properly handle hours

PR #2285, opened 2019-04-26
Limit the hw wallet txs to 7 inputs/outputs

Senyoret1:
Fixes #2284

Changes:
- Now the hw wallet txs can have a maximum of 7 inputs and 7 outputs.

Does this change need to mentioned in CHANGELOG.md?
No

PR #2264, opened 2019-04-09
Use the hardware wallet daemon

Senyoret1:
This PR is not ready for review, as there are still many changes being made to the daemon and the code still needs to react correctly in case of some problems (mainly when there are connection schemes and when the daemon is not running).

Changes:
- Now the connection to the hardware wallet is done through the daemon and not the JS library. However, the code for connecting using the JS library is still present and can be activated with a variable (this could be changed when finishing the PR).

Does this change need to mentioned in CHANGELOG.md?
No

PR #2367, opened 2019-05-27
Multiple improvements for the hardware wallet integration

### LINKED ISSUE #2237: Change hw wallet pin setup symbol

### LINKED ISSUE #2238: Send cancel command on X press

### LINKED ISSUE #2284: Limit the hw wallet transactions to 7 inputs and 7 outputs
Must be done to avoid this problem while a better solution is implemented: https://github.com/skycoin/hardware-wallet-js/issues/102

https://github.com/skycoin/skycoin/pull/2221 should be merged first

Senyoret1:
Fixes #2237
Fixes #2238
Fixes #2284

Changes:
- This PR combines the changes made in https://github.com/skycoin/skycoin/pull/2235 https://github.com/skycoin/skycoin/pull/2239 https://github.com/skycoin/skycoin/pull/2258 https://github.com/skycoin/skycoin/pull/2264 https://github.com/skycoin/skycoin/pull/2274 https://github.com/skycoin/skycoin/pull/2285

- The conflicts were solved.

- Now all the changes work with the hw wallet daemon.

- The compatibility with the JS library was deactivated because some changes made to the library and the firmware made the existing code to stop working. However, the code for using the JS library was left in place just in case.

- Now the data about the known hardware wallets is saved using the `/data` API endpoint, instead of custom Electron code.

- Various improvements for the hardware wallet integration.

No


electron/src/install-dependencies.sh
```diff
@@ -3,4 +3,7 @@
 set -e -o pipefail
 
 npm install
-#./node_modules/.bin/electron-rebuild --only="node-hid"
+# ./node_modules/.bin/electron-rebuild --only="node-hid"
```
gz-c, line 8: What are these extra lines for?

Senyoret1: These lines are for compiling the native dependencies used by the js library, but the js library is deactivated. I did not remove the code just as a precaution in case an important unforeseen problem forces us to return to the js library, but the idea is to remove it later.

gz-c: Can you add a comment explaining why the code is commented out and not just deleted

Senyoret1: Done, in c05a202. I added comments indicating that the code is still there just for precaution and with some instructions about how to reactivate it if nedded.

PR #2372, opened 2019-05-29
Hardware wallet firmware updater

Senyoret1:
Changes:
- Now it is possible to update the hardware wallet firmware using the desktop wallet.

- When the user connects a hardware wallet and opens the hardware wallet options, the desktop wallet checks if there is an update available and, if that is the case, shows a modal window telling the user that the device firmware is outdated and a buttom for opening the updater. If the user cancels the modal window, the hardware wallet options window shows a blinking buttom for opening the updater.

- When the user connects a hardware wallet in bootloader mode and opens the hardware wallet options, the firmware updater is opened.

Notes:

- This PR was made on a branch based on the one used for https://github.com/skycoin/skycoin/pull/2367 , so it currently shows a lot of changes that are not directly related to the updater and should not be merged before https://github.com/skycoin/skycoin/pull/2367 . After merging that PR this one should show less changes.

- As the files with the version number and the firmware have not yet been uploaded to a remote server, this PR gets the version number and the firmware from the `\src\gui\static\src\assets\temp` folder. When making the change for getting the files from a remote server, maybe it will be necessary to add CSP exceptions.

- The code was made for using the hardware wallet daemon, it does not work with the js library.

No


Senyoret1:
The changes made in this PR start in the last commit before this comment (be0fa69)

PR #2374, opened 2019-05-31
Fix #2373 - Correct misc panics with `skycoin-cli` command argument handling

### LINKED ISSUE #2373: Fix misc skycoin-cli panics from command arguments handling
Similar to #2348 - further testing has highlighted other panics resulting from incorrect command arguments passed to `skycoin-cli`. In most cases this appears to relate to issues with the setting the the `Args:` property.

For example, the following:
https://github.com/skycoin/skycoin/blob/cbea117910c86b0f9776852375d1511f7eb09b92/src/cli/last_blocks.go#L15

should be set to `gcli.ExactArgs(1)` as the command expects exactly one argument in all cases. As it is, a panic will occur if no argument is passed - and this is not caught because only a maximum number of arguments is currently specified.

Rather than raising seperate issues (as I expect there are other instances), I will use this one issue to try to clean these up.

BigOokie:
Fixes #2373
Changes:
- Correct panic in lastBlocks and checkdb command argument handling

Yes. Included in PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Don't send wire protocol messages that exceed the configured 256kB limit, which caused peers to disconnect from the sender
 - #2348 Fix panic in `skycoin-cli` `transaction` command if no (zero) arguments are passed. Exactly one argument is expected.
 - #2373 Fix and clean-up further panics with various `skycoin-cli` commands which were not correctly handling arguments.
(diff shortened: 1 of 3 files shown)


therealssj (approved):

gz-c (changes requested):
Changelog entry is in the wrong block of text

BigOokie:
Thanks for calling that out. Will take a look at some point over the weekend and correct. Thanks

BigOokie:
@gz-c @therealssj I have corrected the issue with the changelog. Thanks again for pointing that out.

PR #2378, opened 2019-06-02
Allow to add notes to the transactions

### LINKED ISSUE #2296: Allow to add notes to the transactions (GUI)
If I understood correctly, the notes will be stored locally using the `/data` API endpoint, only for the user's personal reference. @gz-c I am right?

Senyoret1:
Fixes #2296

Changes:
- Now the user can add a note when creating a transaction, with the simple and advanced forms.

- The notes are saved locally with the `/data` api endpoint.

- The user can see the notes in the history.

- The user can modify the notes at any time.

- As the procedures for injecting the transaction and saving the note are different, the code first injects the tx and then saves the note. If the transaction is sent but the note is not saved, a message indicating that is shown.

- It is possible to add notes to old transactions to which no notes were added before.

No

### DIFF
--- a/src/gui/static/src/app/app.datatypes.ts
+++ b/src/gui/static/src/app/app.datatypes.ts
@@ -33,6 +33,7 @@ export class Transaction {
   hoursSent?: BigNumber;
   hoursBurned?: BigNumber;
   coinsMovedInternally?: boolean;
+  note?: string;
 }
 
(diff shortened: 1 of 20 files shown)


gz-c:
@Senyoret1 Anything which adds new functionality to the frontend must be in the changelog.  The only thing we do not put in the changelog are things like internal refactorings.

PR #2380, opened 2019-06-02
Small fixes for the hw wallet integration

### LINKED ISSUE #2379: Delay the hw wallet connection event check
Currently the code start checking if the hw wallet is connected as soon as the app is started. The problem is that if the daemon is not running each failed attempt adds an error to the console, so the console can accumulate a lot of errors after a while.

To solve this, the code that verifies if the hw wallet is connected should start working only after the first operation that involves the hw wallet is started.

Senyoret1:
Fixes #2379

Changes:
- Now the procedure that periodically checks if the hw wallet is connected does not start running just after opening the app, but only after getting a response from the daemon.

- Now the snackbar is closed before starting some operations in `WalletDetailComponent`

No

### DIFF
--- a/src/gui/static/src/app/components/pages/wallets/change-name/change-name.component.html
+++ b/src/gui/static/src/app/components/pages/wallets/change-name/change-name.component.html
@@ -7,7 +7,7 @@
       </div>
     </div>
     <div class="-buttons">
-      <app-button (action)="closePopup()">
+      <app-button (action)="closePopup()" [disabled]="button && button.isLoading()">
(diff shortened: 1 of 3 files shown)


Senyoret1:
The last commit has the following changes:

- Now the cancel button of `ChangeNameComponent` is deactivated while an operation is in progress, to be consistent with the fact that it is not possible to close the modal window.

- Now the cancel operations of the hw wallet do not timeout so quikly, to prevent some errors.

PR #2386, opened 2019-06-03
Get the firmware data from an external URL

### LINKED ISSUE #2377: Remove the test data of the hw wallet firmware updater
Currently the hw wallet firmware updater uses test files for getting the number of the lastest firmware version and the firmware file itself. As soon as possible, these data should be obtained from an external server and the test data should be deleted.

Senyoret1:
Fixes #2377

Changes:
- The test data related to the firmware was deleted. The data is now retrieved from a server.

- The remote server URLs for getting the data realted to the firmware were added to a configuration file.

- Small fix in the version checking code.

No


Senyoret1:
The last commit adds the new URLs to the CSP, so it now works well in proa build.

I also added the daemon URL to the CSP. I did it in this PR to avoid merge conflicts.

PR #2394, opened 2019-06-04
Change the warning of the hw wallet firmware updater

Senyoret1:
Changes:
- Now the warning in the hw wallet firmware updater tells the user that the extra steps before being able to update the firmware are not needed when configuring a new device.

- The text incorrectly said that in order to install the new firmware it was necessary to connect the device in "firmware mode", instead of "bootloader mode". The problem was fixed.

No


Senyoret1:
Replaced by https://github.com/skycoin/skycoin/pull/2397

PR #2397, opened 2019-06-05
Improve the hw wallet firmware updater UX

Senyoret1:
Changes:
- Now the hw wallet firmware updater checks the state of the connected device and shows different messages to the user depending on what the user should do. This is more important after knowing that the devices may not have a firmware, but just a bootloader, so the messages must be more specific to be able to have a good UX.

No


Senyoret1:
An error that appeared when connecting the hardware wallet in firmware mode was fixed in the last commit.

Senyoret1:
In the lastest commit some changes were removed from this PR, because were added to https://github.com/skycoin/skycoin/pull/2414

PR #2406, opened 2019-06-08
Improved code for starting the hw wallet firmware updater

Senyoret1:
Changes:
- An error that prevented the outdated firmware warning to appear in some cases when opening the hw wallet options window was solved.

- The code for detecting if the device is in bootloader mode now react much faster and does not depend in a timeout that could cause other problems.

No


gz-c:
conflict

Senyoret1:
Conflicts solved

PR #2408, opened 2019-06-09
Create a custom snackbar

### LINKED ISSUE #2401: Create a custom snackbar
The current snackbar is very impractical, as it just allow to show a single text block, does not have a lot of customization options, there is no way to manually close it (something specially needed then it is covering some controls), etc. It would be good to create a customized one. The new snackbar should have a button for closing it, as well as allowing to easily customize the background color, add icons and any other needed feature.

When doing this, some improvements could be added to the code used to open the snackbar. Specifically, it would be good to have a centralized code for showing the snackbar.

Also, when using the new snackbar we should stop showing the error messages both in the buttons and the snackbar, as it is redundant.

Senyoret1:
Fixes #2401

Changes:
- `SnackBar` has been replaced with a new custom component called `MsgBar`. The most relevant change at this moment is that it is possible to close the new component by clicking a X button, so now the user has control about when to remove the bar, as sometimes it is displayed above other important UI elements. However, the new bar also displays a title and a icon, and also allows to easilly configure the background color, so it is more flexible and can be use in the future for confirmation messages and other things. The new bar is also compatible with the translation files.

To avoid overcomplicating things, this PR just adds the new bar and uses it in all places where `SnackBar` was previously used. The last part of #2401, about removing the error messages from the buttons to avoid redundancy, was not implemented. Also, after merging this PR and when implementing the last part of #2401, it would be good to take advantage of the fact that the new bar is compatible with the translation files, to be able to remove a lot of intermediate translation code.

MsgBar

No


Senyoret1:
The last commit adds improvements for the code that shows the snackbar. Now the code is simpler. Also, there is a delay for showing the snackbar if it is already in the screen, since showing it again immediately could cause the feeling that the UI is not doing anything.

gz-c:
conflict

Senyoret1:
Conflicts solved

gz-c:
Again (you'll have to keep checking back here as I merge these PRs)

Senyoret1:
All current conflicts in this and the other PRs are now solved, at least for now.

gz-c:
One more conflict

PR #2416, opened 2019-06-14
Improvements for the GUI

Senyoret1:
Changes:
- Now the warning indicating that there are pending transactions and the data may not be updated responds faster.

- The forms for sending coins were responding to the "enter key pressed" event in the wrong field, that have been solved.

- Small code fix for the `ReplaySubject` instances.

No


Senyoret1:
The 2 lastest commits add chages for updating the data of various pages periodically. In some pages the code was not totally optimal and in others the data was only being updated when opening the page.

Also, a small changes was made to the arrows that shows more info in the advanced form and the Swaplab status page. Now the arrows are rotated when showing more info.

PR #2422, opened 2019-06-22
codesign daemon and standalone release

### LINKED ISSUE #2421: codesign the portable binary release for OS X

therealssj:
Fixes #2421 

Changes:
- Codesign daemon and standalone binaries

yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add `/api/v2/data` APIs for transaction notes and generic key-value storage.
 - Update `/metrics` endpoint to add metrics from `/health`: `unspent_outputs`, `unconfirmed_txns`, `time_since_last_block_seconds`, `open_connections`, `outgoing_connections`, `incoming_connections`, `start_at`, `uptime_seconds`, `last_block_seq`.
 - Add to the GUI the ability to choose specific unspent outputs to spend
+- Codesign daemon and standalone binaries
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Document the daemon's CLI options
 - Add the ability to save transaction notes
 - Add `package bip44`, implementing the bip44 spec https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki
+- Codesign daemon and standalone binaries
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add `/api/v2/data` APIs for transaction notes and generic key-value storage.
 - Update `/metrics` endpoint to add metrics from `/health`: `unspent_outputs`, `unconfirmed_txns`, `time_since_last_block_seconds`, `open_connections`, `outgoing_connections`, `incoming_connections`, `start_at`, `uptime_seconds`, `last_block_seq`.
 - Add to the GUI the ability to choose specific unspent outputs to spend
+- Codesign daemon and standalone binaries
```
gz-c: Wrong section of the changelog

electron/package-daemon-release.sh
```diff
@@ -46,12 +46,29 @@ function copy_if_exists {
     fi
 }
 
+function codesign_if_exists {
+     if [ -z "$1" ]; then
+        echo "copy_if_exists requires binary path"
```
gz-c: should be `codesign_if_exists`

electron/package-standalone-release.sh
```diff
@@ -51,12 +51,29 @@ function copy_if_exists {
     fi
 }
 
+function codesign_if_exists {
+     if [ -z "$1" ]; then
+        echo "copy_if_exists requires binary path"
```
gz-c: should be `codesign_if_exists`

electron/package-standalone-release.sh
```diff
@@ -51,12 +51,29 @@ function copy_if_exists {
...
+        exit 1
+    fi
+
+    BIN="${GOX_GUI_OUTPUT_DIR}/${1}"
+
+    if [ -f "$BIN" ]; then
+        echo "signing osx binary"
+        codesign --force --sign "Developer ID Application: yunfei mao" "${BIN}"
```
gz-c, line 64: What does `--force` mean?

gz-c (changes requested):

electron/package-standalone-release.sh
therealssj: `-force When signing, causes codesign to replace any existing signature on the path(s) given. Without this option, existing signatures will not be replaced`
from the manpages

therealssj: the standalone binary has a signature already ( I dont know here it comes from )
```
Executable=/Users/therealssj/Downloads/skycoin-0.26.0-gui-standalone-osx-darwin-x64/skycoin
Identifier=skycoin-c3ab3afec9fc367ce1c47c00967020e26d883dd4
Format=Mach-O thin (x86_64)
CodeDirectory v=20100 size=146369 flags=0x2(adhoc) hashes=4569+2 location=system
Hash type=sha256 size=32
CandidateCDHash sha1=bff9109e71fb6c4b1660c7df6e3bf08d54cfb903
CandidateCDHash sha256=dca09a2947ff9893b5f23c0c21ed9184662a7130
Hash choices=sha1,sha256
Page size=4096
CDHash=dca09a2947ff9893b5f23c0c21ed9184662a7130
Signature=adhoc
Info.plist=not bound
TeamIdentifier=not set
Sealed Resources=none
Internal requirements count=0 size=12
```
if you try to sign it without `--force`
it returns this error: `skycoin: is already signed`

gz-c: Strange because what motivated me to request this is that when I ran the standalone binary from the website, it warned me that it was from an unknown developer

therealssj: https://developer.apple.com/documentation/security/seccodesignatureflags/kseccodesignatureadhoc
because it has no identity attached to it
`codesign --force --sign - myapp.app"`
this is how adhoc signing is done, by using `-` as a "pseudo" identity

PR #2426, opened 2019-06-23
Improvements and fixes for the hw wallet integration

Senyoret1:
Changes:
- Add a new state to the hw wallet updater to tell the user when the firmware is being downloaded (important when using a really slow Internet connection).

- Fix a problem detecting when the user cancels an operation (due to a change in the text that the desktop wallet is receiving from the daemon).

No


gz-c:
conflict

PR #2428, opened 2019-06-26
Use the correct name for the hw wallet firmware file

Senyoret1:
Changes:
- Now the hw wallet firmware updater expect the firmware file to be called `skywallet-firmware-v$(VERSION_FIRMWARE).bin` instead of `skyfirmware_$(VERSION_FIRMWARE).bin`. This makes the updater follow the naming convention described in https://github.com/skycoin/hardware-wallet#versioning-policies

IMPORTANT: at this moment the version returned by https://version.skycoin.net/skywallet/version.txt is `v1.0.0-dev` but the server does not have a file called `skywallet-firmware-v1.0.0-dev.bin`, so for this PR to work the file in the server will have to be renamed or, even better, the new firmware version will have to be compiled and uploaded to the server using the correct naming convention.

No

### DIFF
--- a/src/gui/static/src/app/app.config.ts
+++ b/src/gui/static/src/app/app.config.ts
@@ -3,7 +3,7 @@ export const AppConfig = {
   maxHardwareWalletAddresses: 1,
   useHwWalletDaemon: true,
   urlForHwWalletVersionChecking: 'https://version.skycoin.net/skywallet/version.txt',
-  hwWalletDownloadUrlAndPrefix: 'https://downloads.skycoin.net/skywallet/skyfirmware_',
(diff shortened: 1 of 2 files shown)


gz-c:
The `v` prefix will always be part of the version.  

I have uploaded https://downloads.skycoin.net/skywallet/skywallet-firmware-v1.0.0-dev.bin so you can test this

Senyoret1:
I checked it and the file was downloaded correctly. The only problem is that the firmware file is a bit outdated and it is not compatible with the lastest version of the daemon, but that will really be a problem when starting the tests, so I think there is no hurry for updating it right now

PR #2431, opened 2019-06-30
Multiple improvements for the hw wallet integration

Senyoret1:
Changes:
- The code for processing errors and showing the state of the hw wallet operations now is much better, as it got messy after all the changes made latelly.

- Improvements for the management of the input focus.

- Fix for a problem in the layout when using Firefox.

- Fix the code that detect when a operation was cancelled because the device was disconnected in Linux/MacOS.

- Improvements for the UX when confirming an address.

- Improvements for the event that detects if the device is connected.

No

As this PR is big, I will check the changes again later.


Senyoret1:
I checked the PR again and added some changes in the lastest commit:

- Some fixes for the code modified in the previous commit.

- Improvements for the UI/UX in various parts.

PR #2432, opened 2019-07-02
Implement "collection" type wallets

### LINKED ISSUE #2252: Wallet file modified with cli addPrivateKey does not work properly in advanced send form

gz-c:
Fixes #2252 

Changes:
- Restructure the Wallet struct into an interface, in order to support multiple wallet types (defined by behavior). This will make implementing a bip44 wallet easier. I chose to implement the simpler "collection" type wallet first, in order to focus on a good solution to handling multiple wallet types. 
- The interface allows us to define different internal structures for "Entries" (needed for bip44, which has multiple chains of keys), and to define different keypair generation mechanisms. 
- Support multiple serialized wallet structures. The only common requirement is the wallet metadata struct. The "entries" can be serialized whichever is most appropriate for the wallet.  "collection" and "deterministics" wallets both serialize their "entries" as a single array, which is unchanged from previous behavior.  Bip44 HD wallets may need to use multiple chains.
- Add a new "collection" type wallet. The "collection" type wallet has no keypair generation method, and thus no seeds. It is an arbitrary collection of private keys. It can be managed with `cli addPrivateKey`.  In the future, we can add support to create and manage it through the UI, but this is not necessary yet.
- "collection" wallets can be empty, whereas "deterministic" wallets can't be.
- To solve wallet deduplication, wallets define a Fingerprint() method.  "collection" wallets have no fingerprint. "deterministic" wallets have a fingerprint of "deterministic-$first_address".  In the future, bip44 wallets would have a fingerprint of "bip44-$first_address".
- There is a single constructor for all wallet types (`newWallet()`)

I've added documentation of the wallet types in https://github.com/skycoin/skycoin/wiki/Wallet-File-Formats-and-Types

Needed changes:
- Empty "collection" wallets break the UI, and give it an error#2.  To replicate: from this branch, do `go run cmd/cli/cli.go walletCreate -t collection`, which creates an empty collection wallet, and then `./run-client.sh` to run the client.  You should experience a redirect to http://127.0.0.1:6420/assets/error-alert/?2

Yes


Senyoret1:
The GUI error should be solved with https://github.com/gz-c/skycoin/pull/8 . Also, I don't know if it was done or if it would be appropiate for this PR, but may be https://github.com/skycoin/skycoin/issues/1965 should be taken into consideration, not for solving it, but just to make it easier to solve it in the future

gz-c:
The watch wallets in #1965 should be easier after this restructuring

gz-c:
This also may address #2252 (although I never replicated that bug report)

src/wallet/deterministic_wallet.go
```diff
@@ -0,0 +1,337 @@
...
+
+	for {
+		// Generate the addresses to scan
+		addrs, err := w2.GenerateSkycoinAddresses(n)
+		if err != nil {
+			return err
+		}
+
+		// Find if these addresses had any activity
+		active, err := tf.AddressesActivity(addrs)
```
iketheadore, line 235: Maybe we can have a method in the `tf` interface to check addresses activity from the end to first. And returns the index of the first encountered address with balance. We actually don't care if the addresses have balance before the last active address. In this way, the method don't have to go through all addresses unless all are empty. And because we already know the address index, we don't have to go through the addresses again in the following code to get the `keepNum`.

olemis (changes requested):
I might have overlooked something , but does this need a note in `cmd/cli/README.md` about wallet types and deterministic as default ?

src/wallet/deterministic_wallet.go
gz-c: Can make that a separate issue (optimize scanner). It doesn't matter for majority of users who use few addresses, but after we switch to automatic change addresses it will.

gz-c: https://github.com/skycoin/skycoin/issues/2437

PR #2434, opened 2019-07-03
Add new methods to Visor's public method set

not4x217:
Three public methods are added to visor:

1. CreateBlock
2. VerifyBlock
3. ExecuteSignedBlockUnsafe

Those methods are necessary for applications, which:

1. Want decoupled block creation, block verification and block execution.
2. Are going to perform block signing and block signature verification on their own.

### DIFF
--- a/src/visor/blockchain.go
+++ b/src/visor/blockchain.go
@@ -255,7 +255,7 @@ func (bc *Blockchain) processBlock(tx *dbutil.Tx, b coin.SignedBlock) (coin.Sign
 			}
 			b.Body.Transactions = txns
 
-			if err = bc.verifyUxHash(tx, b.Block); err != nil {
+			if err := bc.verifyUxHash(tx, b.Block); err != nil {
 				return coin.SignedBlock{}, err
 			}
 
--- a/src/visor/blockchain.go
+++ b/src/visor/blockchain.go
@@ -231,40 +231,38 @@ func (bc Blockchain) NewBlock(tx *dbutil.Tx, txns coin.Transactions, currentTime
 	return b, nil
 }
 
-func (bc *Blockchain) processBlock(tx *dbutil.Tx, b coin.SignedBlock) (coin.Transactions, error) {
+func (bc *Blockchain) processBlock(tx *dbutil.Tx, b coin.SignedBlock) (coin.SignedBlock, error) {
 	length, err := bc.Len(tx)
 	if err != nil {
-		return nil, err
+		return coin.SignedBlock{}, err
 	}
-
-	var blockTxns coin.Transactions
 
 	if length > 0 {
 		if isGenesis, err := bc.isGenesisBlock(tx, b.Block); err != nil {
-			return nil, err
+			return coin.SignedBlock{}, err
 		} else if isGenesis {
 			err := errors.New("Attempted to process genesis block after blockchain has genesis block")
 			logger.Warning(err.Error())
-			return nil, err
+			return coin.SignedBlock{}, err
 		} else {
 			if err := bc.verifyBlockHeader(tx, b.Block); err != nil {
-				return nil, err
+				return coin.SignedBlock{}, err
 			}
 
-			var err error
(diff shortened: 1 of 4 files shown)


src/visor/blockchain.go
```diff
@@ -253,10 +253,9 @@ func (bc *Blockchain) processBlock(tx *dbutil.Tx, b coin.SignedBlock) (coin.Sign
 			if err != nil {
 				return coin.SignedBlock{}, err
 			}
-
 			b.Body.Transactions = txns
 
-			if err := bc.verifyUxHash(tx, b.Block); err != nil {
+			if err = bc.verifyUxHash(tx, b.Block); err != nil {
```
gz-c: undo

src/visor/visor.go
```diff
@@ -380,21 +390,53 @@ func (vs *Visor) CreateAndExecuteBlock() (coin.SignedBlock, error) {
...
+	if err := b.VerifySignature(vs.Config.BlockchainPubkey); err != nil {
+		return err
+	}
+
+	return vs.db.Update("ExecuteSignedBlock", func(tx *dbutil.Tx) error {
+		return vs.executeSignedBlock(tx, b)
+	})
+}
+
+// ExecuteSignedBlockUnsafe adds block to the blockchain, or returns error.
```
gz-c, line 425: Explain what this is for?

gz-c (changes requested):

src/visor/visor.go
not4x217: That is needed for applications, which Are going to perform block signing and block signature verification on their own.

src/visor/blockchain.go
not4x217: Done.

src/visor/visor.go
```diff
@@ -380,21 +390,53 @@ func (vs *Visor) CreateAndExecuteBlock() (coin.SignedBlock, error) {
...
+// Blocks must be executed in sequence. Block signature is not verified.
+func (vs *Visor) ExecuteSignedBlockUnsafe(b coin.SignedBlock) error {
```
gz-c, line 427: Need to update the string label in the DB.Update call below

src/visor/visor.go
```diff
@@ -380,21 +390,53 @@ func (vs *Visor) CreateAndExecuteBlock() (coin.SignedBlock, error) {
...
+		return nil
+	})
+
+	return sb, err
+}
+
+// VerifyBlock verifies specified block against local copy of blockchain.
+// Signature is not verified.
+func (vs *Visor) VerifyBlock(b coin.SignedBlock) error {
+	return vs.db.View("VeirfyBlock", func(tx *dbutil.Tx) error {
```
gz-c: typo `VeirfyBlock` -> `VerifyBlock`

gz-c (changes requested):
There are multiple calls to `executeSignedBlock` that you have not updated. 

The original behavior should be preserved in all ways.  

Make two internal methods:

- `executeSignedBlock`: does signature verification, then calls `executeSignedBlockUnsafe`
- `executeSignedBlockUnsafe`

src/visor/visor.go
not4x217: Done.

src/visor/visor.go
not4x217: Done.

src/visor/visor.go
```diff
@@ -380,21 +390,59 @@ func (vs *Visor) CreateAndExecuteBlock() (coin.SignedBlock, error) {
 	return sb, err
 }
 
+// CreateBlock creates a Block from specified set of transactions according to set of determinstic rules.
+func (vs *Visor) CreateBlock(txns coin.Transactions, when uint64) (coin.Block, error) {
```
gz-c: The name should match the function it calls, in this case `CreateBlockFromTxns`, and make sure the db.Update label matches too

gz-c (changes requested):

PR #2439, opened 2019-07-07
Add a guided method for entering the seeds in the GUI

### LINKED ISSUE #2407: Create a guided method for entering the seeds
If the user wants to recover a wallet, the UI shows a textarea for manually entering the seed. It was made this way to let advanced users to make changes to the seed and offer more flexibility. However, this has also resulted in many less experienced users having problems. There have been some concerned users who hought they entered the seed correctly but got a new address without funds, normally due to problems with blank spaces, an uppercase letter or a misspelled word.

To avoid this, and also to improve the UX, we should add an additional mode to enter the seed, but allowing the user to use the current method if desired. The new method could be based in the one used for the hw wallet: ask for the words one by one, in a field that only allows to use words from the BIP32 dictionary.

Senyoret1:
Fixes #2407

Changes:
- Now, by default, when creating or restoring a wallet the UI does not show a textarea for manually entering the seed, but a guided method. When creating a new wallet, the UI shows the seed in a non editable field and for confirming it the user has to enter each word, one by one, in a modal window that allows BIP39 words ony. When restoring a wallet, the user must enter the seed using the same method used for confirming the seed when creating a new wallet. This way, it is much more difficult to make mistakes with the seed and the method is consistent with the one used with the hw wallet.

- If an user wants to use the previous method, there is a button for switching between methods.

- The e2e tests were modified to work well with the changes.

Now the form for creating wallets looks like this:
create

Yes


Senyoret1:
@gz-c please take a look at the texts in the `en.json` file

Senyoret1:
I checked the changes again and added some improvements in the last commit. The code should be ok now.

gz-c:
Is there a fallback method to let the user enter any arbitrary text?

Senyoret1:
Yes, you just have to press the button with the lock icon at the lower right corner of the field that shows the address and accept a warning. After that, the form returns to be just as how it is now. It looks like this:

seed

The only difference is the button for returning to the guided method.

Issue #2395, opened 2019-06-05
Encode transaction JSON object


olemis:
**Feature description**

Get raw transaction from JSON

**Is your feature request related to a problem? Please describe.**
While running pre-release testing procedure of hardware wallet libraries it is necessary to test signing with the device. After doing so and adding signatures in JSON object representing the target transaction , there is no way to encode it to raw so as to broadcast it.

**Describe the solution you'd like**
Facilitate transaction serialization to users 

**Describe alternatives you've considered**
[`/api/v2/transaction/sign`](https://github.com/skycoin/skycoin/tree/master/src/api/README.md#sign-transaction) is not appropriate since signing is performed by the node.

**Additional context**
Related to skycoin/hardware-wallet-js#112 .

**Possible implementation**

- [x] CLI command to encode transaction JSON object
- [x] (Optionally) update transaction inner and outer hash (see skycoin/hardware-wallet-go#144)
- ~REST API endpoint to encode transaction JSON object~

olemis:
While working on skycoin/hardware-wallet-go#144 had to apply this (incorrect) patch against e2db0000541af05632174970ed7b7bf1afcbf4b8 so as to be able to finish the process .

```diff

diff --git a/src/api/spend.go b/src/api/spend.go
index 9c91ae556..e0e2e1950 100644
--- a/src/api/spend.go
+++ b/src/api/spend.go
@@ -190,8 +190,10 @@ func (r *CreatedTransaction) ToTransaction() (*coin.Transaction, error) {
                return nil, err
        }

-       if t.Hash() != hash {
-               return nil, errors.New("readable.Transaction.Hash does not match parsed transaction hash")
+       if false {
+               if t.Hash() != hash {
+                       return nil, errors.New("readable.Transaction.Hash does not match parsed transaction hash")
+               }
        }

        return &t, nil
```

Need a way to relax such constraints by updating txn hashes .

jdknives:
@therealssj has this been fixed by #2468 ?

olemis:
> @therealssj has this been fixed by #2468 ?

yes

PR #2453, opened 2019-07-12
[cli] refs #2395 - Add --fix CLI arg in encodeJSONTxnCmd

olemis:
refs #2395

Changes:
- New `--fix` arg in `encodeJSONTxnCmd` for re-computing transaction hashes prior to encoding

no

### DIFF
--- a/src/api/spend.go
+++ b/src/api/spend.go
@@ -126,18 +126,14 @@ func NewCreatedTransaction(txn *coin.Transaction, inputs []visor.TransactionInpu
 	}, nil
 }
 
-// ToTransaction converts a CreatedTransaction back to a coin.Transaction
-func (r *CreatedTransaction) ToTransaction() (*coin.Transaction, error) {
+// ConvertToTransaction converts a CreatedTransaction back to a coin.Transaction, by optionally recomputing inner and outer hashes
+func (r *CreatedTransaction) ConvertToTransaction(fix bool) (*coin.Transaction, error) {
 	t := coin.Transaction{}
 
 	t.Length = r.Length
 	t.Type = r.Type
 
 	var err error
-	t.InnerHash, err = cipher.SHA256FromHex(r.InnerHash)
-	if err != nil {
-		return nil, err
-	}
 
 	sigs := make([]cipher.Sig, len(r.Sigs))
 	for i, s := range r.Sigs {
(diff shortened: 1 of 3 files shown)


src/api/spend.go
```diff
@@ -185,18 +181,43 @@ func (r *CreatedTransaction) ToTransaction() (*coin.Transaction, error) {
 
 	t.Out = out
 
+	// Attempt fixing inner hash first
+	// use case https://github.com/skycoin/hardware-wallet-go/issues/144
```
gz-c, line 185: Please explain why the InnerHash is wrong and needs to be recomputed

src/api/spend.go
```diff
@@ -185,18 +181,43 @@ func (r *CreatedTransaction) ToTransaction() (*coin.Transaction, error) {
...
+		r.TxID = t.Hash().Hex()
+	}
+
 	hash, err := cipher.SHA256FromHex(r.TxID)
 	if err != nil {
 		return nil, err
 	}
 
 	if t.Hash() != hash {
+		fmt.Printf("Hash() %s\nhash %s", t.Hash().Hex(), hash.Hex())
```
gz-c, line 209: Don't print here, put the message inside of the error

src/api/spend.go
```diff
@@ -126,18 +126,14 @@ func NewCreatedTransaction(txn *coin.Transaction, inputs []visor.TransactionInpu
 	}, nil
 }
 
-// ToTransaction converts a CreatedTransaction back to a coin.Transaction
-func (r *CreatedTransaction) ToTransaction() (*coin.Transaction, error) {
+// ConvertToTransaction converts a CreatedTransaction back to a coin.Transaction, by optionally recomputing inner and outer hashes
+func (r *CreatedTransaction) ConvertToTransaction(fix bool) (*coin.Transaction, error) {
```
gz-c, line 130: Don't use a bool arg on a public method

src/cli/transaction.go
```diff
@@ -90,6 +90,11 @@ func encodeJSONTxnCmd() *cobra.Command {
 				return err
 			}
 
+			fixHashes, _err := c.Flags().GetBool("fix")
+			if _err != nil {
+				return _err
```
gz-c, line 95: Don't use underscored prefixed variables, ever

gz-c (changes requested):

PR #2465, opened 2019-07-19
Add distributeGenesis cli command

gz-c:
Changes:
- Add `distributeGenesis` command (formerly, `fiber-init`) to the CLI tool
- Add `no_broadcast` bool flag to `/api/v1/injectTransaction` to disable the txn broadcast. This is to allow transaction injection to a network consisting of a single signing node and no other node type.  Otherwise, broadcast would fail due to a lack of peers.
- Expose `block_publisher` status in `/api/v1/health`

Yes


therealssj (approved):
looks good

PR #2474, opened 2019-07-29
Make the GUI work with the lastest version of the hw wallet daemon

Senyoret1:
Changes:
- The GUI now works with the lastest version of the hw wallet daemon, including the improvements for the timeout and the changes made to how the API responds.

No


Senyoret1:
The lastest commit updates the dist folder, to make it easier to test the changes.

Senyoret1:
The lastest commit shows a warning when changing the name of a hw wallet, to make the user know that any non-standard English character may not be correctly displayed in the hw wallet.
 
The change was added here because this PR is needed for using the hw wallet integration with the lastest version of the daemon.

PR #2478, opened 2019-08-15
Improve the forms for sending coins

Senyoret1:
Changes:
- Now when the user switches from USD to SKY and viceversa, the value or values already entered by the user are converted to the new currency, but only if the user accepts a confirmation. If the code detects precision problems due to the limitation of decimal places, an alert if shown.

- Now there is an option for sending all the coins. In the simple form it simply adds all the coins in the wallet to the “amount” field (the value is converted to USD if needed). In the advanced form, if the user already added coins to other destinations, it only adds the remaining coins to the selected destination.

- The advanced form is supossed to be blocked if an address field is empty. That was not working as it should and this PR solves the problem.

- Now the white spaces at the beginning and end of the addresses and amounts are ignored, to make it easier to paste values.

- The simple form now shows the name of the selected currency at the right of the amount field, just like in the advanced form.

No


Senyoret1:
The changes in the lastest commit are:

- If the user has only one wallet, that wallet is selected by default.

- The method for validating amounts in the advanced form is now centralizad.

- A problem with the state of a variable was solved in `SendSkycoinComponent`.

- The advanced form now shows the total amount of coins and hours that the user is going to send if there is more than one destination.

PR #2483, opened 2019-09-13
Change .net to .com in the GUI

Senyoret1:
Changes:
- The .com domain is now used intead of the .net domain in the GUI.

- The dist folder was updated.

Note: there is still Go code pointing to the .net domain and that could affect the GUI. An example is that the explorer URL is obtained from the Go code.

Yes, but maybe it would be better after making the changes in the Go code, so I did not add it.


Senyoret1:
The lastest commit updates the webdriver-manager dependency, to make the e2e tests run with the lastest version of Google Chrome. It also fixes a problem that was causing the e2e tests to timeout when the server does not return the lastest version of the destop wallet.

iketheadore:
Already changed to .com.

Issue #2312, opened 2019-05-13
Consider proxying the swaplab API with our own server


gz-c:
Swaplab prefers us to use the API key server-side and not in the client, so we'll want to switch to our own server proxy in the future, to avoid service interruptions due to key abuse.

PR #2548, opened 2020-11-21
Make the Swaplab integration work in electron and dev builds

Senyoret1:
Changes:
- The problem with the cors policy while connecting to the Swaplab API was solved in the dev enviroment and the Electron version. In the dev enviroment the dev proxy was modified to use the appropiate headers. For Electron code was created for using something like a "cors response emulator", which includes and removes some headers and redirects some connections to a local URL, to avoid sending unnecessary requests to the backend and make it possible to complete the operations. Also, an error msg indicating that the Swaplab integration is not available in the portable version was added.

Note: adapted from https://github.com/SkycoinProject/skycoin/pull/242#issuecomment-599097810

No

### DIFF
--- a/electron/app/electron-api.js
+++ b/electron/app/electron-api.js
@@ -1,4 +1,4 @@
-// Warning: for this to work the contextIsolation property of the browser window must be set to true.
+// Warning: for this to work the contextIsolation property of the browser window must be set to false.
 
 window.isElectron = true;
 window.ipcRenderer = require('electron').ipcRenderer;
--- a/electron/app/electron-main.js
+++ b/electron/app/electron-main.js
@@ -207,6 +207,52 @@ function createWindow(url) {
     console.log('Cleared the caching of the skycoin wallet.');
   });
 
+  // When an options request to Swaplab using an https endpoint is detected, asume that it is a
+  // cors request and redirect it to an invalid endpoint on the node API.
+  ses.protocol.registerHttpProtocol('https', (request, callback) => {
+    if (request.method.toLowerCase().includes('options') && request.url.toLowerCase().includes('swaplab.cc')) {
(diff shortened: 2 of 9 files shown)


iketheadore:
Hi, do we just call swaplab's APIs or we will also load data and js code from them. If we have to load their js code, we will stop this service, as it may has potential security issues.

iketheadore:
I find callings of https://swaplab.cc/api/v3/trading_pairs in the wallet when click the `Buy Skycoin`, it is not an API endpoint. Maybe we should hide the `Buy Skycoin` service before we figure out how to do it through APIs.

Senyoret1:
The integration is made using an API and no js code is download from the Swaplab server. The process is just calling API endpoints and receiving JSON responses. When you called https://swaplab.cc/api/v3/trading_pairs the server returned the whole web page because the URL can not be called just like that, you must send a POST request with some credentials. The documentation about the API was posted in this issue: https://github.com/skycoin/skycoin/issues/2032

Also, note that there is a pending issue to make the API request using an intermediate web server managed by us and not by directly calling the Swaplab API. Making that change would also solve the problem this PR is trying to solve, but would also require much more work: https://github.com/skycoin/skycoin/issues/2312

The code currently used for sending the API requests is in https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/services/exchange.service.ts , in the https://github.com/skycoin/skycoin/blob/77258703160b376d44ca52e7a242e2fceb50084c/src/gui/static/src/app/services/exchange.service.ts#L96-L105 function you can see how the POST requests are created for making the server respond as it should.

However, if you have any additional questions, just ask me. If you still have some concerns, I can just remove the tab from the UI.

iketheadore:
Thanks for your explanation, I'll take a look.

iketheadore:
This looks good.

PR #2576, opened 2021-01-13
Update the UI dependencies

Senyoret1:
Changes:
- Angular was updated to v11.
- Electron was updated to v11, which includes an updated Chromium version and compatibility with Apple Silicon (but compatibility would have to be added to the node).
- The Electron code was updated to work with the new version.
- Axios was updated.

No


iketheadore (approved):
👍

PR #2585, opened 2021-03-11
Update Electron to v12

Senyoret1:
Changes:
- Electron was updated from v11 to v12. This includes an update to Chromium 89.

No

### DIFF
--- a/electron/package-lock.json
+++ b/electron/package-lock.json
@@ -21,9 +21,9 @@
             }
         },
         "@electron/get": {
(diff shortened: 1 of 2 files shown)


Senyoret1:
The Github tests appear to be failing due to a download problem, but I'm not sure about the exact cause. @iketheadore could you please take a look?

iketheadore:
I'll try to fix the test errors in a new PR.

PR #2619, opened 2021-06-05
Format the decimal values

### LINKED ISSUE #2602: Make the number easier to read
Requested by Synth:

When typing numbers in wallet, like 10 000.

Can we put a slight space, every 3 digits to make it easier to read number, and make it easier to see difference between 10,000 and 100,000

or put an implicit comma, that the user doesn't type and which is not in the text field

Senyoret1:
Fixes https://github.com/skycoin/skycoin/issues/2602#issuecomment-850651360

Changes:
- Now the decimal part of the numbers shown in the UI is a bit transparent and smaller, to make it easier to diferentiate it from the integer part.

No


iketheadore:
https://github.com/skycoin/skycoin/issues/2602#issuecomment-871078034

Issue #2630, opened 2021-08-07
Problems related to the temporal wallets


Senyoret1:
When using the `/api/v1/wallet/createTemp` API endpoint, the `password` and `encrypt` params are ignored and the wallets are always created unencrypted. However, it is possible to use the `/api/v1/wallet/encrypt` API endpoint for encrypting the wallet.

If the desired behavior is to prevent temporal encrypted wallets to be created, then the encryption options must be removed from the UI for those wallets and maybe the `/api/v1/wallet/encrypt` API endpoint should return an error. Otherwise, the `/api/v1/wallet/createTemp` API endpoint should encrypt the wallets just like the `/api/v1/wallet/create` API endpoint does.

Appart from this, the `/api/v1/wallet/createTemp` API endpoint should be added to the API documentation.

iketheadore:
Yes, it is designed to do not encrypt temporary wallet. I'll update the API `/api/v1/wallet/encrypt` to ignore the temporary wallets. And update the API document.

PR #2631, opened 2021-08-07
Allow to load wallets temporarily with the UI

### LINKED ISSUE #1109: Temporary wallet load feature
User should be able to load a wallet temporarily:

* Checkbox during "Load Wallet" to load the wallet temporarily
* After loading, wallet is stored unencrypted in memory 
* Wallet is not saved to disk

This allows users to restore a wallet from seed, perform some transaction with it and exit the client without retaining any wallet file on disk.

Senyoret1:
Fixes #1109

Changes:
- Now the UI allows to load wallets temporarily, using the wizard and the modal window. The option is not shown when using the option for creating a new wallet.

- Temporal wallets are indentified in the wallet list.

@iketheadore please check #2630, as it affects this.

No


Senyoret1:
Remove the password options for temporal wallets

The last commit includes the following changes:
- The password options are removed when loading temporal wallets, in the wizard and the modal window. Also, no options for adding/removing passwords for temporal wallets are shown in other parts of the UI.
- Some additional related changes needed in the UI.

Solves part of #2630

PR #2639, opened 2021-08-29
Add `/api/v1/wallet/xpub` API endpoint

### LINKED ISSUE #2522: Create an API endpoint for getting the xPub key of a wallet
Senyoret1 commented on Apr 6

There should be an API endpoint for getting the xPub key of a wallet. The endpoint should work with BIP44 wallets, so the user can later use the xPub key for creating a read-only wallet. It should also work on xPub wallets, as a way to let the user create a backup, just like the user can request the seed for being able to create a backup.

Please take #2523 into account.

iketheadore:
Fixes #2522 

Changes:
- Add API endpoint `/api/v1/wallet/xpub` to get xpub key of bip44 wallets.

Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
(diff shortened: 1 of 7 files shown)


0pcom:
These changes have been included / merged on #2744 - and will be on develop after #2740 is merged

PR #2667, opened 2022-02-28
Update Angular and Electron

Senyoret1:
Changes:
- Angular was updated to V13.
- TSLint was replaced with ESLint, as TSLint was removed in Angular 13. Several changes were made to solve problems returned by the new linter.
- Electron was updated to v17.

No


Senyoret1:
Test are failing because the test system wants Chrome 99 and the test enviroment has Chrome 98

PR #2720, opened 2025-07-18
Add peer

0pcom:
Add peer ; update code deps


0pcom:
CI tests are broken. I've only added a peer and updated code deps. Merging at Synth's request.

PR #2929, opened 2026-07-11
skycoin-web: embeddable improvements (custom-node GUI + --socks5-proxy + context-cancellable serve)

0pcom:
## Summary

Two small, related changes so the skycoin-web thin client can be pointed at arbitrary nodes — including nodes reachable over a skywire mesh — both at runtime and via a proxy.

### 1. Enable the custom-node GUI in production builds
`environment.prod.ts`: `enableCustomNodes: false -> true`.

The Settings → Nodes page (runtime node-URL management, persisted to `localStorage['nodeUrls']`) already exists but was gated off in prod. Turning it on gives production builds parity with dev and lets a user repoint the thin client at any node without a rebuild. The committed `src/gui/dist` is rebuilt to match (main bundle hash + index.html only).

### 2. `--socks5-proxy` flag on the skycoin-web server
Sets `HTTP_PROXY`/`HTTPS_PROXY` for the node HTTP client. With a resolving SOCKS5 proxy this lets `--node-url` be a skywire-mesh hostname (`<name>.<pk>.dmsg` / `.skynet`) resolved remotely by the proxy. Equivalent to exporting `HTTP(S)_PROXY` yourself; the flag just makes it a first-class option. The longer explanation lives in the command's Long description; the flag help is kept short.

## Test plan
- `go build ./cmd/skycoin-web/...` passes.
- Production dist rebuilt with `ng build --configuration production`; `wasm_exec.js` unchanged (real 16992-byte standard-Go loader), only the main bundle + index.html differ.
- Settings → Nodes GUI reachable in the prod build; custom node URLs persist and drive the thin client's API base.


0pcom:
Added a third commit: **context-cancellable graceful serve.** `serve()` now takes a `context.Context`, runs an `http.Server` with graceful `Shutdown` on cancel (instead of the blocking gin `router.Run`), and returns errors instead of `log.Fatalf`. The standalone CLI passes `cmd.Context()` (background → Ctrl+C, no behavior change); an embedder using `ExecuteContext(ctx)` can now stop it cleanly and won't be killed by a server error.

This is the prerequisite for running skycoin-web **in-process as a skywire internal-launcher app** (the convergence work — so skycoin-web is 'just another internal app' on both the host-native and wasm visor, launched via the same path as skychat, rather than a raw external subprocess).

PR #3017, opened 2026-08-11
chore: scope the gosec suppressions instead of silencing them repo-wide

0pcom:
`.golangci.yml` carried seven repo-wide `text: "G###:"` exclusions, switching off
**G101, G115, G117, G120, G602, G703 and G706 everywhere**. gosec was in the
enabled linter list while the categories that matter most were muted — so a real
hardcoded credential or path traversal introduced later would have been
suppressed silently.

They were also redundant with a convention the codebase already follows: there
are **428 existing `//nolint:gosec` comments**.

## What removing them surfaced

64 findings, in two groups.

**Three systematic clusters** — every hit is the same intentional pattern in the
same kind of file, so a `//nolint` on each of ~55 lines would be noise nobody
maintains. These keep an exclusion, but **scoped by path and rule** rather than
applied globally:

| rule | path | why |
|---|---|---|
| G101 | `_test\.go` | test vectors, not credentials |
| G117 | `_test\.go` | fixtures that intentionally carry a password field |
| G115 | `src/(cipher\|btc)/` | deliberate truncation while packing bytes |

Everything outside those paths still reports.

**Twelve one-offs** carry `//nolint:gosec` at the site with a reason, matching
the existing convention:

- `src/coin/transactions.go` — G602, guarded by `verifyInputSignaturesPrelude`
- `firmware_update.go` — G120, body already bounded by `http.MaxBytesReader`
- `src/cli/address_gen.go` — G703 on an operator-supplied output path; G115 on a validated count
- `src/daemon/gnet/message.go` — G115, packing a rune into a protocol byte
- `src/skycoin-lite/mobile/api.go` — G117, an API that deliberately returns key material
- `cmd/explorer`, `cmd/skycoin-web` — G706 ×4, logging request paths on a local server

## The two that looked serious were checked, not assumed

Both are false positives:

- **`transactions.go:239`** — `verifyInputSignaturesPrelude` checks
  `len(txn.In)` against `len(uxIn)` and `len(txn.Sigs)` before the loop runs.
- **`firmware_update.go:28`** — `http.MaxBytesReader` is applied on the line
  immediately above the multipart parse.

## Verification

gosec reports **0 findings with no global suppressions**; `go build .` and
`go vet ./src/...` pass; tree is gofmt clean.

> Stacks with #3014, which excludes `node_modules` (the `flatted` npm package
> ships a Go file that `./...` walks into). Until that merges, a lint run here
> still shows 3 revive findings from those vendored files.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -68,6 +68,24 @@ linters:
       - src/cipher/debug_
       - src/cipher/secp256k1-go/debug_
     rules:
+      # gosec, scoped rather than silenced globally.
+      #
+      # These three are systematic: every hit is the same intentional pattern in
+      # the same kind of file, so a per-site //nolint on each of ~55 lines would
+      # be noise nobody maintains. Everything outside these paths still reports,
+      # and genuine one-offs carry //nolint with a reason at the site.
+      - linters:
+          - gosec
+        text: "G101:"          # test vectors, not credentials
+        path: _test\.go
+      - linters:
+          - gosec
+        text: "G117:"          # fixtures that intentionally carry a password field
+        path: _test\.go
+      - linters:
+          - gosec
+        text: "G115:"          # deliberate truncation while packing bytes
+        path: "src/(cipher|btc)/"
       - linters:
           - revive
         text: "var-naming: avoid meaningless package names"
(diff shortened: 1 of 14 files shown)


0pcom:
Rebased onto develop (picks up the `node_modules` exclusion from #3014) and fixed two problems the CI runs exposed:

**Malformed nolint directives.** Five sites were written as `//nolint:gosec:G703`, which is not valid — the format is `//nolint:<linter>`, so `gosec:G703` parsed as an unknown linter name and the directive was silently ignored. Corrected to `//nolint:gosec // reason`.

**G120 is taint-based and nondeterministic.** It flags `r.FormValue` in the node's HTTP handlers and explores as far as its analysis budget allows, so a *different subset* of handlers is reported on each run — the darwin job named `wallet.go` on one run and `transaction.go:428/458` on the next, neither of which my local run reported. Per-site `//nolint` cannot keep up with that, so G120 is now scoped to `src/api/` with the reasoning recorded in the config.

`make lint` is clean: **0 issues**, `go vet` passes, `go build .` succeeds.

